PR #1, opened 2016-08-29
[JENKINS-37788] First work on removing isConstant.

abayer:
[JENKINS-37788](https://issues.jenkins-ci.org/browse/JENKINS-37788)

Needs more testing to make sure we're not breaking valid use cases of
expressions - right now, it's just special casing $class cases, which
isn't ideal. Still a work in progress.

cc @reviewbybees @oleg-nenashev @michaelneale @kohsuke

### DIFF
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTValue.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTValue.groovy
@@ -44,12 +44,6 @@ public abstract class ModelASTValue extends ModelASTElement {
     Object value
 
     /**
-     * If the value can be determined without side-effect at AST parsing time,
-     * this method returns true, and {@Link #getValue()} returns its value.
-     */
-    public abstract boolean isConstant();
-
-    /**
      * Returns a value or an expression that represents this value.
      *
      * This model is used at the compile time, so it's not always possible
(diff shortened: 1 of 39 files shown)


src/main/groovy/org/jenkinsci/plugins/pipeline/config/parser/JSONParser.groovy
```diff
@@ -223,11 +223,15 @@ class JSONParser {
...
 
     public @CheckForNull ConfigASTValue parseValue(JSONObject o) {
-        if (o.getBoolean("isConstant")) {
-            return ConfigASTValue.fromConstant(o.get("value"), o)
-        } else {
-            return ConfigASTValue.fromGString(o.getString("value"), o)
+        // TODO: Probably need better type-checking? Not sure yet.
+        if (o.getString("value") != null) {
+            String valString = o.getString("value")
+            if (valString.contains('$class')) {
```
abayer: One of the places I'm concerned with special casing - I need some way to tell when a value is a string representation of an expression, and so far, that's the best I've got...but it won't work with some other things (like passing a list or map to a step on purpose) so...hrm.

src/main/groovy/org/jenkinsci/plugins/pipeline/config/validator/ConfigValidator.groovy
```diff
@@ -255,17 +258,24 @@ class ConfigValidator {
     }
 
     private boolean validateParameterType(ConfigASTValue v, Class erasedType, ConfigASTKey k = null) {
-        if (v.isConstant()) {
-            try {
+        try {
+            // TODO: Clean this up a lot. We're secure now, but....yeah.
+            if (v.value instanceof String && ((String)v.value).contains('$class')) {
```
abayer: Same issue as above with https://github.com/jenkinsci/pipeline-config-plugin/pull/1/files#diff-81e0fc9e9ef0ae7814d88c89b2ddcbc3R229

src/main/groovy/org/jenkinsci/plugins/pipeline/config/parser/JSONParser.groovy
michaelneale: why is it looking for $class? Is this trying to discern between: 

" foo is ${bar}"
and 'foo is $bar' if it was regular groovy?

PR #2, opened 2016-09-01
[FIXED JENKINS-37828] Properly reject parallel

abayer:
[JENKINS-37828](https://issues.jenkins-ci.org/browse/JENKINS-37828)

More notably, reject it when it's not the only step in a block. In the
JSON representation, we don't even use the term parallel - we just
have multiple branches. In the groovy parsing, we look for parallel as
the only step in a stage and treat it differently (resulting in
multiple branches). So any time we see parallel as a step by the time
we get to validation, well, it's wrong. So, yeah.

Also, tweaks to make sure we don't end up doing validation of steps we've already blacklisted, because that's silly.

cc @reviewbybees @i386 @michaelneale @oleg-nenashev

### DIFF
--- a/SYNTAX.md
+++ b/SYNTAX.md
@@ -97,10 +97,14 @@ stages {
 
 stages {
     stage('parallel-stage') {
-        parallel([
-            firstBlock: 'echo "First block of the parallel"',
-            secondBlock: 'echo "Second block of the parallel"'
-        ])
+        parallel(
+            firstBlock: {
+                echo "First block of the parallel"
+            },
+            secondBlock: {
+                echo "Second block of the parallel"
+            }
+        )
     }
 }
 ```
--- a/SYNTAX.md
+++ b/SYNTAX.md
@@ -70,6 +70,7 @@ to-be-released block-scoped `stage` syntax in base Pipeline.
 * *Takes a Closure*: Yes
 * *Closure Contents*: One or more Pipeline steps, including block-scoped steps and the special `script` block described below.
     * *NOTE*: Only the "declarative subset" of Groovy is allowed by default. See below for details on that subset.
+    * *NOTE*: The `parallel` step is a special case - it can only be used if it's the sole step in the `stage`.
 * *Examples*:
 
 ```
@@ -91,6 +92,15 @@ stages {
             checkout scm
             sh "mvn clean install"
         }
+    }
+}
+
+stages {
+    stage('parallel-stage') {
+        parallel([
+            firstBlock: 'echo "First block of the parallel"',
+            secondBlock: 'echo "Second block of the parallel"'
+        ])
     }
 }
 ```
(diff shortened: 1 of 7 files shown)


abayer:
ping @michaelneale - and if @jglick, @svanoort, @batmat or anyone else on @reviewbybees wants to give this a look, it'd be appreciated. =)

svanoort:
> In the groovy parsing, we look for parallel as
the only step in a stage and treat it differently (resulting in
multiple branches)

I'm assuming to keep things simple for Blue Ocean use?

abayer:
That and general model simplicity.

SYNTAX.md
```diff
@@ -93,6 +94,15 @@ stages {
         }
     }
 }
+
+stages {
+    stage('parallel-stage') {
+        parallel([
```
michaelneale: is this different to the script style of: 

```
parallel (
    "branch" : {...}
)
```

?

michaelneale:
yes the model has always had branches under stages (so you can have parallel branches, or single execution).

SYNTAX.md
```diff
@@ -93,6 +94,15 @@ stages {
...
+            firstBlock: 'echo "First block of the parallel"',
```
batmat: Somehow as a followup of Michael above, do those blocks really take a string, of shouldn't it be a closure?

src/test/resources/errors/rejectParallelMixedInSteps.groovy
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+pipeline {
+    agent none
+    stages {
+        stage("foo") {
+            echo "hello"
+            parallel([a: "echo '1'",
+                      b: "echo '2'"])
+        }
```
batmat: Just tested and 
```
parallel (
    bim: {
        echo "hey"    
    },
    bam: {
        echo "hoo"    
    }
)
```

(without the [ ] around the map elements), and this seems to be working fine with raw pipeline. Is there a reason you didn't choose the same syntax (or maybe both are working, didn't try it)?

michaelneale: @batmat originally I tried it with that syntax too and it worked fine in this. Have you tried the classic syntax? 
Yes I don't think it is warranted to change it like this. Parallel syntax isn't great, and this isn't any better so I don't see any point in it.

michaelneale:
OK with the enforcement, but -10 on the apparent syntax changes(?)

SYNTAX.md
abayer: I....screwed up the syntax! Fixing.

src/test/resources/errors/rejectParallelMixedInSteps.groovy
abayer: Yeah, I screwed up here too. Total brain fart. Didn't get caught (or matter particularly) because we're rejecting on the parallel anyway, but I'm fixing it now.

svanoort:
Mostly-comprehending 🐝

PR #4, opened 2016-09-02
[FIXED JENKINS-37928] Properly detect sections without values.

abayer:
[JENKINS-37928](https://issues.jenkins-ci.org/browse/JENKINS-37928)

The exact error message could be better - I need to understand what
other possible ways this could fail to be a method call expression
besides just not having arguments, but this should be general enough a
message for all possibilities for now.

cc @reviewbybees esp @oleg-nenashev @batmat @svanoort @michaelneale

### DIFF
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -118,35 +118,39 @@ class ModelParser {
         def sectionsSeen = new HashSet();
         eachStatement(pipelineBlock.body.code) { stmt ->
             def mc = matchMethodCall(stmt);
-            def name = parseMethodName(mc);
-            // Here, method name is a "section" name at the top level of the "pipeline" closure, which must be unique.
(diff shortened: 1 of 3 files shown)


abayer:
ping @reviewbybees ?

michaelneale:
Looks ok - although it may confuse people by calling "agent" a section (at least the wording) - vs suggesting what comes after it? 
Given that :bee:

abayer:
"section" is any of the top-level elements. I have no better word for them. =)

michaelneale:
no "section" is fine, I am just spitballing: 

"Not a valid section definition: 'agent'"

specifically for agent it would read like: 

"Not a valid section definition: 'agent'. You are missing configuration options like: label:'master'"

Perhaps there is a more generic way to express they should look up the docs on what to put after the section? (it isn't just agent...)

that is all I mean

abayer:
Feel free to open a ticket for the contextual messages - getting that kind of suggestion logic will take some work.

michaelneale:
"Not a valid section definition: 'agent'. Some extra configuration is required"

would that work?

abayer:
Works for me. Changing now.

PR #7, opened 2016-09-07
Demonstrating some stuff which does not work as I would expect

jglick:
Initial integration tests with https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/10 failed, but this seems to be due to preëxisting bugs here.

@reviewbybees

PR #8, opened 2016-09-07
Alternate tests of external library integration

jglick:
#7 is the simpler version, so just filing this for reference—feel free to discard.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.13</version>
+    <version>2.14</version>
   </parent>
 
   <groupId>org.jenkinsci.plugins</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -83,17 +83,7 @@
...
-            <property>
-              <name>hudson.udp</name>
-              <value>33849</value>
-            </property>
-          </systemProperties>
           <argLine>-Xmx4g -XX:MaxPermSize=256m</argLine>
-          <reuseForks>true</reuseForks>
-          <forkCount>${concurrency}</forkCount>
-          <!-- TODO: I hate the retrying tests but given the nondeterministic whackiness of JENKINS-37101... -->
-          <rerunFailingTestsCount>3</rerunFailingTestsCount>
```
jglick: All this stuff is defined in the parent POM (except works better because it is suppressed when `-Dtest=…` is specified).

pom.xml
```diff
@@ -104,7 +94,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-cps</artifactId>
-      <version>2.13</version>
+      <version>2.14-SNAPSHOT</version>
```
abayer: Might as well bump this to 2.15 once that gets released. =)

abayer:
So we've actually figured out why it doesn't work - named vs unnamed parameters fun. Question is do we change the test case to actually have named parameters on the library side, or do we move the test to be in a script block, where unnamed parameters are valid? Or do I do a PR to get rid of the named parameters requirement before then? =)

abayer:
fwiw, given that I can't figure out for the life of me how to get an error in these cases rather than silently not running due to the sandbox eating the `MissingMethodException` somehow or other, I'm leaning strongly towards re-writing to allow multiple unnamed parameters after all.

jglick:
> silently not running due to the sandbox eating the `MissingMethodException` somehow or other

Clearly we need to fix that.

PR #10, opened 2016-09-08
[FIXED JENKINS-38049] Stop using catchError.

abayer:
[JENKINS-38049](https://issues.jenkins-ci.org/browse/JENKINS-38049)

And make sure we throw failures properly. This still feels a bit weird
to me, but it's on the right track, at least.

@vivek - does this look to do the trick for you?

cc @reviewbybees


vivek:
@abayer Tested this PR, fixes the issue, 🐝:)

PR #11, opened 2016-09-09
[JENKINS-38097] Run empty stages for post-failure stages.

abayer:
[JENKINS-38097](https://issues.jenkins-ci.org/browse/JENKINS-38097)

To ensure that we always have the same stage execution order
(including the synthetic postBuild/notification stages), we now switch
to still running stages after a failure, but not running anything *in*
them. I don't know if there's some way we can set those post-failure
stages as "skipped"?

cc @reviewbybees esp @michaelneale @i386 and @vivek - if you've got any ideas on how to mark a stage skipped, that'd be great. =)

### DIFF
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -26,13 +26,10 @@ package org.jenkinsci.plugins.pipeline.modeldefinition;
 
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.NestedModel
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.StepBlockWithOtherArgs
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted
-import org.jenkinsci.plugins.workflow.actions.NotExecutedNodeAction
-import org.jenkinsci.plugins.workflow.actions.StageAction
-import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution
 import org.jenkinsci.plugins.workflow.cps.CpsScript
-import org.jenkinsci.plugins.workflow.cps.CpsThread
-import org.jenkinsci.plugins.workflow.graph.FlowNode
 
 import java.lang.reflect.ParameterizedType;
 
@@ -56,7 +53,7 @@ public class Utils {
      * @return True if the field exists and is of the given type.
      */
     @Whitelisted
-    static boolean isFieldA(Class fieldType, Class actualClass, String fieldName) {
+    public static boolean isFieldA(Class fieldType, Class actualClass, String fieldName) {
         def actualFieldName = actualFieldName(actualClass, fieldName)
         def realFieldType = actualClass.metaClass.getMetaProperty(actualFieldName)?.type
 
@@ -75,7 +72,7 @@ public class Utils {
      * @return The real field name, pluralized if necessary, or null if not found.
      */
     @Whitelisted
-    static String actualFieldName(Class actualClass, String fieldName) {
+    public static String actualFieldName(Class actualClass, String fieldName) {
         if (actualClass.metaClass.getMetaProperty(fieldName) != null) {
             return fieldName
         } else if (actualClass.metaClass.getMetaProperty("${fieldName}s") != null) {
@@ -93,7 +90,7 @@ public class Utils {
      * @return The class of the field in the inspected class, or the class contained in the list or map.
      */
     @Whitelisted
-    static Class actualFieldType(Class actualClass, String fieldName) {
+    public static Class actualFieldType(Class actualClass, String fieldName) {
         def actualFieldName = actualFieldName(actualClass, fieldName)
         if (actualFieldName == null) {
             return null
@@ -122,7 +119,7 @@ public class Utils {
      * @return True if the object is an instance of the class, false otherwise
      */
     @Whitelisted
-    static boolean instanceOfWrapper(Class c, Object o) {
+    public static boolean instanceOfWrapper(Class c, Object o) {
         return c.isInstance(o)
     }
 
@@ -134,17 +131,17 @@ public class Utils {
      * @return True if o can be assigned to c, false otherwise
      */
     @Whitelisted
-    static boolean assignableFromWrapper(Class c, Class o) {
+    public static boolean assignableFromWrapper(Class c, Class o) {
         return c.isAssignableFrom(o)
     }
 
     @Whitelisted
-    static Object[] toObjectArray(List<Object> origList) {
+    public static Object[] toObjectArray(List<Object> origList) {
         return origList.toArray()
     }
 
     @Whitelisted
-    static boolean hasScmContext(CpsScript script) {
+    public static boolean hasScmContext(CpsScript script) {
         try {
             // Just rely on SCMVar's own context-checking (via CpsScript) rather than brewing our own.
             script.getProperty("scm")
@@ -153,17 +150,5 @@ public class Utils {
             // If we get an IllegalStateException, "checkout scm" isn't valid, so return false.
             return false
         }
-    }
-
-    @Whitelisted
-    static void setSkippedStage(String stageName) {
-        CpsThread thread = CpsThread.current()
-        CpsFlowExecution execution = thread.execution
-
-        FlowNode currentNode = execution.currentHeads.find { n ->
-            n?.displayName?.equals(stageName)
-        }
-
-        currentNode.actions.add(new NotExecutedNodeAction())
     }
 }
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -26,10 +26,13 @@ package org.jenkinsci.plugins.pipeline.modeldefinition;
 
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
-import org.jenkinsci.plugins.pipeline.modeldefinition.model.NestedModel
-import org.jenkinsci.plugins.pipeline.modeldefinition.model.StepBlockWithOtherArgs
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted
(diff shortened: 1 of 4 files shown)


vivek:
@abayer How about marking these skipped stages with an Action, NotBuildAction, basically corresponding FlowNode.getAction(NotBuiltAction.class) gives non-null value? FlowNode has getError(), it should really have getResult() so that we know the state of such FlowNode.

abayer:
@vivek - is there a way to do that within the Pipeline script itself?

vivek:
@abayer Don't think there is way to do, guess this is question for @jglick or @svanoort?

abayer:
@vivek Didn't think so. I think that may end up making sense as a JIRA against `workflow-api` - there's already `NotExecutedNodeAction`, but I honestly can't tell what's up with that, so we may want to add a new `Action` there for a "skipped" node. I could then write probably write a `Step` that I could invoke to set that...hrm...

vivek:
@abayer Ah right NotExecutedNodeAction:) I think we can use it for this purpose. Reading doc of [NotExecutedNodeAction](https://github.com/jenkinsci/workflow-api-plugin/blob/ac98823111bef74073501a3656ce0dfe6c199655/src/main/java/org/jenkinsci/plugins/workflow/actions/NotExecutedNodeAction.java) suggests it can be used for nodes that are not executed yet?

abayer:
@vivek Will there be any UI side effects of that? I'm honestly not sure what `NotExecutedNodeAction` is used for in the first place...

vivek:
@abayer I am using it to return status of FlowNode back to UI, helps ui show such nodes muted. grep of stage-view plugin shows its used there as well. Having said that, it will be good to have it reviewed/tested to avoid regression.

abayer:
@vivek Got an idea - will be pushing a new commit here that tries this out (and makes a bunch of static methods package-scoped, but hey, while I'm here, right?)...

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -151,4 +154,16 @@ public class Utils {
...
         }
     }
+
+    @Whitelisted
+    static void setSkippedStage(String stageName) {
+        CpsThread thread = CpsThread.current()
+        CpsFlowExecution execution = thread.execution
+
+        FlowNode currentNode = execution.currentHeads.find { n ->
+            n?.displayName?.equals(stageName)
```
abayer: Not sure if there's a better/more secure way to be sure we've got the right `FlowNode`...

michaelneale:
I am ok with this but it depends on the side effects on stage viewer and stage graph (so they are prepared). Ideally they are shown as skipped somehow (failed would not be right, nor would passed) but in principle I like this approach. 

There is a design aspect though, if we even want to show postBuild or notifications up as a stage (are they as important as other stages? generally not, you only care about them if they failed or if you are looking into details)

svanoort:
@abayer Would not recommend use of NotExecutedNodeAction: it's designed for a very specific use case (resume from Checkpoints) and using it in this way will break UI rendering (we assume that all FlowNodes after the first executed one **will always** be executed).

Thus hard a hard 🐛 

If you want an empty stage, create a stage block with nothing in it.  Easy peasy.

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -108,6 +113,34 @@ public void twoStagePipeline() throws Exception {
...
+    public void allStagesExist() throws Exception {
+        prepRepoWithJenkinsfile("allStagesExist");
+
+        WorkflowRun b = getAndStartBuild();
+        j.assertBuildStatus(Result.FAILURE, j.waitForCompletion(b));
+        j.assertLogContains("[Pipeline] { (foo)", b);
+        j.assertLogContains("hello", b);
+        j.assertLogContains("[Pipeline] { (bar)", b);
+        FlowExecution execution = b.getExecution();
+        FlowGraphWalker walker = new FlowGraphWalker(execution);
```
svanoort: 🐜  Use DepthFirstWalker here, it is much faster.

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -108,6 +113,34 @@ public void twoStagePipeline() throws Exception {
...
+        boolean foundSkippedBar = false;
+        boolean foundExecutedFoo = false;
+
+        for (FlowNode node : walker) {
+            if (node.getDisplayName().equals("bar") && node.getAction(NotExecutedNodeAction.class) != null) {
+                foundSkippedBar = true;
+            }
+            if (node.getDisplayName().equals("foo") && node.getAction(NotExecutedNodeAction.class) == null) {
+                foundExecutedFoo = true;
+            }
```
svanoort: Could probably exit if you found both.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
svanoort: Ehhh.... you could look StepNode with StepDescriptor of StageStep

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -151,4 +154,16 @@ public class Utils {
...
+        }
+
+        currentNode.actions.add(new NotExecutedNodeAction())
```
svanoort: JENKINS-26522 is the way to do something like this, NotExecutedNodeAction has a very specific use case and it's not this, and this would break expected behavior.

abayer:
@svanoort And that's why we have code review. =) Thanks!

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Well, irrelevant now anyway since we're just doing an empty stage, but good to know!

svanoort:
No problemo -- :bee: I think (though I don't follow the groovy deeply here)

vivek:
@svanoort @abayer 

>@abayer Would not recommend use of NotExecutedNodeAction: it's designed for a very specific use case (resume from Checkpoints) and using it in this way will break UI rendering (we assume that all FlowNodes after the first executed one will always be executed).

In this case these stages are not executed. In BO we want to show such notes as 'not executed', because they are in fact not executed. Either NotExecutedNodeAction javadoc is misleading or stage view implemented it specifically for checkpoint resume. 

>Available so that views of a job flow node graph can be properly rendered i.e. the steps that were not executed are visibly muted in some way.

In any case we want deterministic way to determine a step/stage was not used, I hope it doesn't result in to 'NotExecutedNodeAction2' action;)

i386:
@vivek @abayer I'd like us to cover this PR today and what it means for the visualization

svanoort:
> In this case these stages are not executed. 

@vivek In this case, the stages *have* been executed, they are just no-ops (but you still have fully realized nodes).   They have their appropriate actions (LabelAction, TimingAction, logs, etc), whereas nodes before a checkpoint can lack this (literally nothing happened, they are just placeholders for previous execution). 

 A better handling would be to allow for skipping over (or visually muting) empty stages for visualization purposes.  It's trivial to track how many nodes are in a stage or block: <10 LoC with Bismuth (I did one of those in a prototype).  Or for the BO graph-based model you can do a quick peek-ahead to see if a BlockEndNode has just the BlockStartNode preceding it. 

> In any case we want deterministic way to determine a step/stage was not used,

I am very puzzled what you mean here -- what is your use case here, specifically, that is not met by empty stages and the above?   If we truly need more complex status rendering, the correct way is a JENKINS-26522 implementation -- which has a plan of execution and is not a ton of work.  This is directly blocked because we've frozen Stage View development, but could be rolled out on the backend and picked up by Blue Ocean.
 
The below is valid syntax and can denote something meaningful:

```groovy
stage('empty'){ }
```` 

@i386  I would strongly encourage not going down the NotExecutedNodeAction implementation path (it breaks back compatibility) until we can get together to discuss as a full group.

i386:
@svanoort agreed - I need to wrap my head around the implications

abayer:
Well, for now I've reverted back to empty stages.

vivek:
@svanoort 

>In this case, the stages have been executed, they are just no-ops (but you still have fully realized nodes). They have their appropriate actions (LabelAction, TimingAction, logs, etc), whereas nodes before a checkpoint can lack this (literally nothing happened, they are just placeholders for previous execution).

Maybe we are different terminology when we refer to not-executed. Let me try to explain
:  Lets say I have staged s1, s2, s3, postBuilld, if s1 fails then today the flowNodes I get are s1 and postBuild as s2 and s3 are not executed. With this PR you get s1, s2(empty), s3(empty), postBuild. Here s2 and s3 are represented empty and the fact is that s2 and s3 indeed haven not been implemented. It would be a good DRY pattern for Kyoto to mark such nodes with state not-executed instead of every visualization implementation try to put custom logic to check if they indeed have not been executed.

>I am very puzzled what you mean here -- what is your use case here, specifically, that is not met by empty stages and the above?

Use case is simple, FlowNode for a stage tells us whats the state of it. If stages are not executed - just state that. Kyoto is capable of decorating FlowNode with such information, it will be huge simplification.

michaelneale:
@svanoort @vivek the main thing is that from a Vidal perspective empty stages that were "skipped" are not presented as failure or success, but it is clear they are there, they just did no work (no comments on impl, but that is the end goal here)

michaelneale:
"Visual"

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: And it didn't work anyway - couldn't find a `StepNode` with the right name. Grr.

abayer:
@vivek @svanoort Mind re-reviewing? We'll approach the long-term handling of "skipped" stages separately.

svanoort:
@vivek @abayer To clarify, real problem is not use of NotExecutedNodeAction but the notion that you're mixing and matching executed and not executed sections in arbitrary order (which is not expected for NotExecutedNodeAction).  

Agree that the better solution is to add a new action.  

I think this part is already 🐝  able but let's discuss further on how to do the action.

PR #13, opened 2016-09-13
[JENKINS-38153] Adding synthetic stage marker.

abayer:
[JENKINS-38153](https://issues.jenkins-ci.org/browse/JENKINS-38153)

Also adding synthetic stages for pre-build stuff.

cc @reviewbybees esp @vivek


abayer:
Also cc @svanoort

svanoort:
Agree with adding a new action -- I think this is where we should do a trivial [JENKINS-26522](https://issues.jenkins-ci.org/browse/JENKINS-26522) implementation.  Give me a minute I'm going to hack together something for you :)

svanoort:
@abayer  and @vivek  I like the basic ideal but think a JENKINS-26522 implemetation around arbitrary annotations like this is better and more flexible: 

https://github.com/jenkinsci/workflow-api-plugin/pull/12  <-- POC concept

In that case, we'd use a tag "SYNTHETIC" on the action.

abayer:
@svanoort Well, that's part of it, but we did need more metadata than just the tag - we're also recording the execution context (currently the only options are pre-build/post-build, but I have a feeling we'll add more to that in the future).

svanoort:
> execution context

Yeah you're just annotating which synthetic stage type it is.   That's a SYNTHETIC tag and/or SYNTHETIC_PRE/SYNTHETIC_POST.

abayer:
Awwww, I have to translate everything to string tags? I'm back in AWS-land! =)

svanoort:
@abayer Okay, but implementation-nitpicking aside: don't tell me you don't want JENKINS-26522...  ;)  Or that you will get to be the trailblazer use case to justify it, with pipeline model definitions.

abayer:
Yes, I do want that. =)

jglick:
Seems hackish. Better to define a simple `Step` with a block and whatever `Action`s you need, possibly though not necessarily including `LabelAction`.

svanoort:
@abayer If we pull in the TagsAction here I think will be ready for merge?

abayer:
@svanoort Yeah, I can redo this with `TagsAction`.

abayer:
Now properly downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/12

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/SyntheticContext.java
```diff
@@ -0,0 +1,36 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.pipeline.modeldefinition.actions;
+
+public enum SyntheticContext {
```
svanoort: 🐛  I think you meant to remove this, since we don't need the whole Context/ContextAction bit after adding the TagsAction.

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/SyntheticStageMarkerAction.java
```diff
@@ -0,0 +1,48 @@
...
+import hudson.model.InvisibleAction;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Marker action for stages that aren't user-specified but are in fact created by the model directly.
+ *
+ * @author Andrew Bayer
+ */
+public class SyntheticStageMarkerAction extends InvisibleAction {
```
svanoort: 🐛  Not needed now that we can use TagsAction as our marker (and indeed a broader-use marker).

abayer:
FYI, current failure on ci.jenkins.io PR build is due to SNAPSHOT check staleness.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -167,6 +162,27 @@ public class Utils {
...
     }
 
+    /**
+     * Marks the containing stage with this name as a synthetic stage, with the appropriate {@link SyntheticContext}.
+     *
+     * @param stageName
+     * @param context
+     */
+    @Whitelisted
+    static void markSyntheticStage(String stageName, SyntheticContext context) {
```
svanoort: 🐛  Better to use static final strings for the markers, i.e. "PRE" and "POST" or similar.   Creating an enum just for two values is kind of silly, and if we're just getting string values for 'em we might as well use that.

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -145,6 +153,49 @@ public void validStepParameters() throws Exception {
...
+        DepthFirstScanner scanner = new DepthFirstScanner();
+
+        assertNotNull(scanner.findFirstMatch(heads, null, syntheticStagePredicate(SyntheticStageNames.toolInstall(), SyntheticContext.PRE)));
+        assertNotNull(scanner.findFirstMatch(heads, null, syntheticStagePredicate(SyntheticStageNames.checkout(), SyntheticContext.PRE)));
+        assertNotNull(scanner.findFirstMatch(heads, null, syntheticStagePredicate(SyntheticStageNames.postBuild(), SyntheticContext.POST)));
+        assertNotNull(scanner.findFirstMatch(heads, null, syntheticStagePredicate(SyntheticStageNames.notifications(), SyntheticContext.POST)));
+        assertNull(scanner.findFirstMatch(heads, null, syntheticStagePredicate(SyntheticStageNames.dockerPull(), SyntheticContext.PRE)));
+    }
+
+    private Predicate<FlowNode> syntheticStagePredicate(final String stageName,
```
svanoort: Would it make sense to expose this as part of the public API for Blue Ocean to use with the search/filter workflow APIs?

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/SyntheticContext.java
abayer: Nope, I meant to leave it - I want an enum of possible values for the SYNTHETIC_STAGE_TAG.

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/SyntheticStageMarkerAction.java
abayer: Yeah, forgot to delete that.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Sigh. Ok. =)

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
abayer: I'm betting y'all can write something better. =)

abayer:
@svanoort Ok, switched to `String`s. =)

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/SyntheticContext.java
abayer: ..but have now gotten rid of it. yay?

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -167,6 +161,27 @@ public class Utils {
...
+        CpsFlowExecution execution = thread.execution
+
+        FlowNode currentNode = execution.currentHeads.find { n ->
+            n?.displayName?.equals(stageName)
+        }
+
+        if (currentNode.getAction(TagsAction.class) == null) {
+            currentNode.actions.add(new TagsAction())
+        }
+        currentNode.getAction(TagsAction.class).addTag(SYNTHETIC_STAGE_TAG, context)
```
jglick: This is a hack. Better to have a block-scoped `Step` which adds this `Action` using the supported APIs.

src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -217,6 +226,10 @@ public class ModelInterpreter implements Serializable {
     def dockerOrWithout(Agent agent, Closure body) {
         if (agent.docker != null) {
             return {
+                script.stage(SyntheticStage.dockerPull()) {
+                    Utils.markSyntheticStage(SyntheticStage.dockerPull(), SyntheticStage.SYNTHETIC_PRE)
+                    script.getProperty("docker").image(agent.docker).pull()
+                }
                 script.getProperty("docker").image(agent.docker).inside {
```
jglick: BTW more directly written as

```groovy
script.docker.image…
```

abayer: So at least at some point in the past, that very much didn't work. Dunno if it would now, I'll give it a try.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Is there a way to limit the visibility/exposure of a `Step`? i.e., can I create one that only can be used in this plugin? If not, then hacks are really the only way I can see to get what we need - a `Step` that shows up in the snippet generator, etc that does this would be...flawed.

abayer:
ci.jenkins.io is still...questionable at times. Right now it's barfing due to git `user.name` and `user.email` not being set. Fun.

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/SyntheticContext.java
svanoort: Yay!

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
svanoort: You give people too much credit ;)

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -175,22 +176,22 @@ public void syntheticStages() throws Exception {
...
     }
 
     private Predicate<FlowNode> syntheticStagePredicate(final String stageName,
-                                                        final SyntheticContext context) {
+                                                        final String context) {
         return new Predicate<FlowNode>() {
             @Override
             public boolean apply(FlowNode input) {
                 return input.getDisplayName().equals(stageName) &&
                         input.getAction(TagsAction.class) != null &&
```
svanoort: 🐛  Upon further examination (sorry).  Fetching the action 3x is undesirable for performance (sigh).  Normally pessimization wouldn't be a bug but I happen to know this one can be fairly costly. 

Also worth checking for  'input instanceof StepNode && ((StepNode)input).getDescriptor == WhateverTheStageDescriptorIs' -- faster than string equals since it is a direct memory location compare.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
svanoort: Unfortunately I have to disagree with @jglick  about the block-scoped step -- it doesn't feel like synthetic stages are truly a different animal than a normal stage.  We're just attaching some additional helper information to that stage.

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
svanoort: Wait.  Nevermind, that's test-only.  It would be a bug if it were a public API.  

Going to blame that one on still being jetlagged post-JenkinsWorld.

svanoort:
In general: the context removal and string changeover look good.  

I think this one is 🐝  worthy -- though might suggest using your NotNull/CheckForNull annotations some more.

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
abayer: Also, I'm specifically trying to compare the `displayName` to make sure it's the right `stage`. I did try using `StageStartNode`, I think it was, whatever you suggested earlier, and it...did not work. At all. So I went back to this. =)

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: I strangely enough agree with @svanoort =)

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
svanoort: Display name can be the typeDisplayName if there isn't a LabelAction though, so it's not a safe check to be using in general (since you can get false positives).   Again, okay in the walled garden of tests, but I wouldn't want to see it elsewhere.  

In general, prefer type checks and then a LabelAction (or StageAction) check -- you're putting preconditions first that can be tested quickly.  Should be safer and a bit faster (since it'll return negative 99% of the time).

abayer:
Closing in favor of https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/47

PR #14, opened 2016-09-13
JENKINS-38152 - expose execution model on WorkflowRun

abayer:
[JENKINS-38152](https://issues.jenkins-ci.org/browse/JENKINS-38152)

Very much a work-in-progress at this point, but thought I'd get this up for initial feedback.

cc @reviewbybees esp @vivek

### DIFF
--- a/EXTENDING.md
+++ b/EXTENDING.md
@@ -1,0 +1,50 @@
+# Information on extending/integrating with Pipeline Model Definition
+
+## API Endpoints
+
+All API endpoints are exposed under `JENKINS_URL/pipeline-model-converter` currently.
+
+### JSON Schema
+* *URL*: `JENKINS_URL/pipeline-model-converter/schema`
+* *Parameters*: None
+* *Info*: JSON schema for the JSON representation of the model.
+* *Returns*: The JSON schema
+
+### Validation of `Jenkinsfile`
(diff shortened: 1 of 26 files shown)


svanoort:
Hmm.  This is basically like getting source to execution direct mapping options?  I like that.

abayer:
Ok, this isn't actually a work-in-progress now, I think. Test added, it works, I properly strip out source locations...wait, I should test that too.

abayer:
Source location stripping tests added. Oh, and cc @i386, 'cos you may be interested.

abayer:
@svanoort Yeah, not sure if I should be exposing the *entire* model or just the stages - the discussion this came out of was just talking about stages, but we may want to shift to exposing the whole model. Blue Ocean wants this. =)

abayer:
fwiw, I'd still ideally like to stash the `ModelASTPipelineDef` that's actually generated at `Jenkinsfile`-compilation time and reuse that, rather than re-parsing/re-compiling it at execution time, but...so far, I can't figure out how to do that. But I bet I will eventually. =)

i386:
😍😍😍

abayer:
Most recent commit also added `EXTENDING.md` with reference material for REST API endpoints and Java API stuff.

PR #15, opened 2016-09-13
Switching to depend on core 1.642.1

abayer:
This may break enough things that I don't actually merge it.

cc @reviewbybees esp @vivek @HRMPW


abayer:
aaaand closing - forgot that I need Groovy 2.x. =)

PR #17, opened 2016-09-20
[FIXED JENKINS-38242] Add the ability to specify agent parameters (docker)

kuhnroyal:
This adds the `args` parameter of the docker plugin to the agent directive.

### DIFF
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
@@ -46,6 +46,9 @@ public class Agent implements Serializable {
     String label
 
     @Whitelisted
+    String args
+
+    @Whitelisted
     Boolean any
 
     @Whitelisted
@@ -55,6 +58,9 @@ public class Agent implements Serializable {
         }
         if (args.containsKey("label")) {
             label = args.get("label")
+        }
+        if (args.containsKey("args")) {
+            this.args = args.get("args")
         }
     }
 
@@ -70,6 +76,6 @@ public class Agent implements Serializable {
 
     // TODO: Rewrite as an extension point and get this by extension discovery, but we knew that already.
     public static List<String> agentConfigKeys() {
-        return ["docker", "label"]
+        return ["docker", "label", "args"]
     }
 }
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
@@ -46,7 +46,7 @@ public class Agent implements Serializable {
     String label
 
     @Whitelisted
-    String args
+    String dockerArgs
 
     @Whitelisted
     Boolean any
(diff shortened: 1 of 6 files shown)


abayer:
Thanks for the PR! I'll respond in more detail in an hour or so. =)

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
```diff
@@ -46,6 +46,9 @@ public class Agent implements Serializable {
     String label
 
     @Whitelisted
+    String args
```
abayer: So this whole thing is going to get rewritten in the fairly near future, but I see no reason to reject this change in the meantime - it's a good addition that'll translate well into the pluggable approach we'll be moving towards. Buuuuut - could you go with `dockerArgs` instead of just `args` so that we don't end up with confusion down the road with combinations of keys, etc?

src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -214,9 +214,9 @@ public class ModelInterpreter implements Serializable {
     def dockerOrWithout(Agent agent, Closure body) {
         if (agent.docker != null) {
             return {
-                script.getProperty("docker").image(agent.docker).inside {
+                script.getProperty("docker").image(agent.docker).inside(agent.args, {
```
abayer: How does this work a null/empty args?

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
kuhnroyal: Done

src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
kuhnroyal: It works, the default parameter is an empty string but I didn't dig any deeper. I added test cases to catch if anything changes on the side of the docker plugin.

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AgentTest.java
```diff
@@ -121,4 +119,21 @@ public void agentNoneWithNode() throws Exception {
...
         j.assertLogContains("ONSLAVE=true", b);
     }
+
+    private WorkflowRun agentDocker(final String jenkinsfile) throws Exception {
+        assumeDocker();
+        // Bind mounting /var on OS X doesn't work at the moment
+        onAllowedOS(PossibleOS.LINUX);
+        prepRepoWithJenkinsfile(jenkinsfile);
+
+        assumeDocker();
```
abayer: This is redundant.

kuhnroyal: True, was there before, just a blind copy.

kuhnroyal: Fixed

abayer:
Merging - thanks!

PR #19, opened 2016-09-22
[JENKINS-38331] Per-stage configuration for agent

abayer:
[JENKINS-38331](https://issues.jenkins-ci.org/browse/JENKINS-38331)

- [x] Agreed on syntax.
- [x] Implemented runtime and parse-time.
- [x] Testing of success cases.
- [x] Testing of validation failure cases.
- [x] Update `SYNTAX.md`

Honestly, I'm doing this for JENKINS-38284 more, but I needed some
section to actually have be per-stage to test it, so...tada?

This is literally the result of three hours of slamming my head
against the wall over stupid mistakes, so I'm fairly sure there are
still a plethora of such mistakes in here. But I'm pretty sure it
works - I just added the JSON version and conversion for
perStageAgentConfig to this and the tests are actively running as I
type this, but I got impatient and made a commit now. So ha.

cc @reviewbybees esp @jglick @vivek @kohsuke @i386 @michaelneale @rsandell @stephenc


michaelneale:
so if you want to set per stage config, you have to put steps one level below in `steps` section? 

I can understand why, but this is not great

abayer:
@michaelneale Alternatively, we could always require the `steps` section even if you don't have a `config` section. I'm open to alternative approaches, but I just don't see any way to do this without either really bloated arguments to `stage` (I mean, `stage(name:'foo',config:[agent:[label:'some-label']]) { ... }`? Nah.) or a separate `config` section in the `stage`.

michaelneale:
yeah requiring steps always seems quite clear wrong... so will see...

abayer:
@michaelneale Requiring the `steps` block all the time would be preferable - much simpler parsing plus consistency. 

In general - requiring a `steps` block under `stage` doesn't seem to me like a worthy hill to die on. It's basically either that or no per-stage configuration. Pick yer poison.

HRMPW:
if we're being pedantic, then declaring a "steps" section makes it more declarative. :)

It's still possible to have two global variables (one called _config_ and one called _steps_ ) that completely bombs out but I think the odds of that happening are small.

abayer:
@HRMPW again, I'd still prefer to move to requiring the `steps` block in all cases, which gets us past that edge case forever. And makes parsing a lot simpler. =)

michaelneale:
I could possibly live with this (I would want the steps nested section only in cases of config). Would be nice if was possible at all to not require that though. That is my thinking

HRMPW:
I quickly polled @mslusarczyk and @mattstep. They thought having _steps_ without _config_ didn't make as much sense. I'm torn. I think not having _steps_ when there is no _config_ looks cleaner and is easier to write but writing _steps_ all the time is a better practice in general.

i386:
Throwing this out here:

```
pipeline {
  agent docker:'foo'

  stages {
    stage ('bar') {
      agent docker:'bar'
      sh './run.sh'
    }
  }

  notifications {
    stage ('bar') {
      always {
        send_hipchat 'bar stage is done'
      }
    }
    always {
      send_hipchat 'pipeline is done'
    }
  }
}
```

abayer:
Nope, sorry - can't mix config and steps.

michaelneale:
@abayer mixing config and steps would also count as: 

```
pipeline {
  agent docker:'foo'

  stages {
    stage ('bar') {
     config {
        agent docker:'bar'
     }
      
      sh './run.sh'
    }
  }


}
```

too?

michaelneale:
so - need to talk more about this... functionality wise, seems great.

michaelneale:
yeah the `steps` requirement with `config` is not going to fly. -1, need to find another way.

abayer:
Then I'm going to need suggestions for alternatives - the only other things
I can think of are either putting all the config for a stage in a map as a
parameter to stage, which would be very messy and ugly, or defining the
config for a stage outside of the stages block entirely.

michaelneale:
will think about it...

michaelneale:
@abayer is there anyting blocked on this (other features) as we could shelve this for now (under mvp/v1) IMO (and from talking to KK) for another time.

abayer:
So this is blocking, well, anything that needs to happen per-stage (so [JENKINS-38331](https://issues.jenkins-ci.org/browse/JENKINS-38331) obviously and [JENKINS-37792](https://issues.jenkins-ci.org/browse/JENKINS-37792) off the top of my head). Based on user feedback, I think those both need to be fairly high priorities for 1.0.

michaelneale:
yes patrick and I were talking, hopefully will catch you at the end of the day for this

michaelneale:
ok this will require updating of external docs etc - so @HRMPW - when should this go in?

src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy
```diff
@@ -101,28 +105,21 @@ public class ClosureModelTranslator implements MethodMissingWrapper, Serializabl
...
-                        if (Utils.instanceOfWrapper(Closure.class, thisArg)) {
-                            blockParams.add(createStepsBlock(thisArg))
-                        } else {
-                            // Otherwise, just add the parameter.
-                            blockParams.add(thisArg)
-                        }
+                    String n = origArgs[0]
+                    def ctm = new ClosureModelTranslator(actualType, script)
+
+                    if (Utils.instanceOfWrapper(Closure.class, origArgs[1])) {
```
rsandell: Risk of `ArrayIndexOutOfBounds` if someone has a brain fart and writes something like.
```groovy
stage "build"
steps {
...
}
```

src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -106,9 +107,10 @@ public class ModelInterpreter implements Serializable {
                                                 firstError = e
                                             }
                                         }
-                                    }
+                                    }.call()
```
rsandell: I have been wondering about all these "return closure and call it directly" approach that is scattered throughout. Why not just run the closure directly inside then method `nodeOrDockerOrNone`? Is it to clean up the stacktraces or what?

abayer: Honestly? Can't remember exactly. I think it was to deal with making sure we didn't execute at the wrong time...

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
```diff
@@ -343,19 +383,6 @@ public void importAndFunctionShouldNotSkipParsing() throws Exception {
     }
 
     private void failWithError(final String... errors) throws Exception {
```
rsandell: Should be called `assertFailWithError` it took me a while to get why you just failed the test with a message when you actually made an assertion that the build would fail with a specific message.

rsandell:
:bee: since I only have :ant:s and this is blocking my work ;)

rsandell:
Noticed now that `ModelASTStage.toGroovy()` is missing the `agent` output. So :bug:

src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -82,19 +82,20 @@ public class ModelInterpreter implements Serializable {
...
+                        for (int i = 0; i < root.stages.getStages().size(); i++) {
+                            Stage thisStage = root.stages.getStages().get(i)
 
-                                script.stage(thisStage.name) {
-                                    if (firstError == null) {
+                            script.stage(thisStage.name) {
+                                if (firstError == null) {
+                                    nodeOrDockerOrNone(thisStage.agent) {
                                         try {
                                             catchRequiredContextForNode(root.agent) {
```
rsandell: Shouldn't this be something like `thisStage.agent != null ? thisStage.agent : root.agent` ?

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
abayer: Renaming.

abayer:
@rsandell are you sure `agent` isn't getting into the generated groovy? Looks like it is to me...

michaelneale:
this is a fairly hefty change. I do like it now (without config block, but with steps{} always) - would like ot talk it over with @i386 but isn't around this week

rsandell:
Yes, even though I looked three times and coudn't see it, now, after a night of sleep, I do see it :) So :bee: again.

src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
rsandell: Yes, but you should be requiring the context for the node you are working on right? If a step requests the current workspace to be injected in this context it would get the node for the stage, not the outer one. And if it is something wrong you would check the context of the wrong agent.

abayer:
Good point. Need to think about source control autocheckout too...

abayer:
Well, merging for now. =)

PR #20, opened 2016-09-22
[FIXED JENKINS-38426] Require non-literal env vars to be single-quoted.

abayer:
[JENKINS-38426](https://issues.jenkins-ci.org/browse/JENKINS-38426)

This way, we can delay evaluation until runtime in the right context
and get the right values without any security exception etc.

This may not be ideal - but as of right now, it's the best approach
I've got.

cc @reviewbybees esp @vivek @rsandell @HRMPW @michaelneale @i386

### DIFF
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
@@ -103,12 +103,12 @@ public class Root implements NestedModel, Serializable {
      * Helper method for translating the key/value pairs in the {@link Environment} into a list of "key=value" strings
      * suitable for use with the withEnv step.
      *
-     * @return a list of "key=value" strings.
+     * @return a list of [key,value] lists.
      */
     @Whitelisted
-    List<String> getEnvVars() {
+    List<List<Object>> getEnvVars() {
         return environment.collect { k, v ->
-            "${k}=${v}"
+            [k, v]
         }
     }
 
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -196,6 +196,10 @@ class ModelParser {
                                 errorCollector.error(key, "Duplicate environment variable name: '${key.key}'")
                                 return
                             } else {
+                                if (exp.rightExpression instanceof GStringExpression) {
+                                    errorCollector.error(key, "Non-literal environment variable ${getSourceText(exp.rightExpression)} must be in single-quotes")
+                                    return
+                                }
                                 r.variables[parseKey(exp.leftExpression)] = parseArgument(exp.rightExpression)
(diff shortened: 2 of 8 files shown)


michaelneale:
I was hoping I could do this: 

```
pipeline {
    agent label:''
    
    stages {
        stage('build') {
            sh 'echo $YEAH'
        }
        
    }
    
    environment {
        YEAH = '${thing()}'
    }
    
}

def thing() {
    return "HEY"
}
```

But can't. 

Also: 

```
pipeline {
    agent label:''
    
    stages {
        stage('build') {
            sh 'echo $YEAH'
        }
        
    }
    
    environment {
        YEAH = thing()
    }
    
}

def thing() {
    return "HEY"
}
```

fails - should that be possible?

abayer:
The latter I'm not prepared for yet. =)

The former...hmm. Gotta think about how to do that. It'll work fine for a lot of things, but not for that use case right now.

Again, both will be much easier if we moved to 
```groovy
environment {
    foo bar()
}
```

HRMPW:
Would syntax without "=" work for all permutations?

`environment {
     foo bar() 
     HEY "Yah"
     BIG 'Bang'
}`

abayer:
@HRMPW As I mentioned on the JIRA, there may be issues with variable names in all caps - I remember hitting problems with that in the past. I'd have to play with it to be sure.

michaelneale:
@abayer I just want to test it with globals etc - so I just need to a spare day to find out how they work now.

HRMPW:
This might be a stupid question but for the environment block is it possible for the delegate to take each line as a string so you can manipulate any way you want and then _evaluate_ the string after manipulation?

abayer:
Sadly, no. Variable assignment as an operation messes things up.

michaelneale:
@abayer I wasn't able to make global libs work (not for this PR, but everything as described: https://github.com/jenkinsci/workflow-cps-global-lib-plugin doesn't actually seem to work) - so I don't know how to validate this.

michaelneale:
ok, finally was able to make it work (upgrading and multiple restarting). 

So I think this might be ok. 

The single quote may be confusing (but we are using single quotes most every where else) to a pipeline "expert". 

Also calling a function like 

```
AWS_KEY_ID = '${secret("thing")}'
```

instead of 
```
AWS_KEY_ID = secret("thing")
```

looks and is hard to explain. So many quotes to understand, no one will get it right. So I am not sure. I think this is an ok starting point, if it didn't block the way off in the future for direct functions that aren't in a string?

abayer:
So after further experimentation...

- `FOO "BAR"` syntax doesn't work because Groovy doesn't accept `FOO` as a method name. `"FOO" "BAR"` does work, though.
- neither `"FOO" currentBuild.getNumber()` nor `FOO = currentBuild.getNumber()` can work. In both cases, the `currentBuild.getNumber()` gets evaluated too early. Basically, any non-literal, non-closure expression has to be within a closure that we're not evaluating at parse-time or that expression gets evaluated at parse-time rather than execution-time.

So the problem isn't actually the variable assignment - I was wrong about that. I'll try to see if there are any alternatives.

abayer:
Closing in favor of #22

PR #21, opened 2016-09-23
Another attempt at JENKINS-38426 (non-literal env vars)

abayer:
This only works on top of #19. I am very confused and need to do further research, but putting this up here for reference purposes if nothing else.


abayer:
Oh wait, I think it's the whitelist change...

abayer:
Closing since I figured out how to do it on its own.

PR #22, opened 2016-09-23
[FIXED JENKINS-38426] Allow non-literal expressions in env vars.

abayer:
[JENKINS-38426](https://issues.jenkins-ci.org/browse/JENKINS-38426)

Allowing fall-through delegation and whitelisting getProperty(s) on
ClosureModelTranslator and PropertiesToMapTranslator lets us do env
vars with values like:
- "${currentBuild.getNumber()}"
- currentBuild.getNumber()

Can't do "${PREVIOUSLY_DEFINED_ENV}_something" yet, though. Still
trying to figure that one out.

This replaces #20 

cc @reviewbybees esp @rsandell @i386 @michaelneale @HRMPW

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/MethodMissingWrapperWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/MethodMissingWrapperWhitelist.java
@@ -41,7 +41,9 @@ import java.lang.reflect.Method;
 public class MethodMissingWrapperWhitelist extends Whitelist {
     @Override
     public boolean permitsMethod(@Nonnull Method method, @Nonnull Object receiver, @Nonnull Object[] args) {
-        return (method.getName().equals("invokeMethod") || method.getName().equals("setProperty"))
+        return (method.getName().equals("invokeMethod") ||
+                method.getName().equals("setProperty") ||
+                method.getName().equals("getProperty"))
                 && MethodMissingWrapper.class.isAssignableFrom(receiver.getClass());
     }
 
--- a/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy
@@ -195,7 +195,7 @@ public class ClosureModelTranslator implements MethodMissingWrapper, Serializabl
     private void resolveClosure(Object closureObj, Object translator) {
         Closure argClosure = closureObj
         argClosure.delegate = translator
-        argClosure.resolveStrategy = Closure.DELEGATE_ONLY
+        argClosure.resolveStrategy = Closure.DELEGATE_FIRST
         argClosure.call()
     }
 }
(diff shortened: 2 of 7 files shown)


abayer:
So using one env var inside another doesn't work with `withEnv` in the first place, so...yeah. Got an idea for a workaround, though!

abayer:
Workaround achieved! 
```groovy
environment {
    FOO = "BAR"
    CONTAINS_FOO = "\${FOO} is set"
}
```

Inelegant? Sure. But I can live with it.

HRMPW:
I didn't get into the code in great detail but the tests look good. If all of the tests are passing we can try out some more things. I'll let @michaelneale and @rsandell look at the code more closely. :bee:

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/MethodMissingWrapperWhitelist.java
```diff
@@ -41,7 +41,9 @@
 public class MethodMissingWrapperWhitelist extends Whitelist {
     @Override
     public boolean permitsMethod(@Nonnull Method method, @Nonnull Object receiver, @Nonnull Object[] args) {
-        return (method.getName().equals("invokeMethod") || method.getName().equals("setProperty"))
+        return (method.getName().equals("invokeMethod") ||
```
oleg-nenashev: Maybe usage of static constant set and <code>contains()</code> are not that bad in this case if we're aware about performance (seems so). 🐜

abayer: If we add anything more here in the future, I'll go that route.

michaelneale:
This seems :bug: as it doesn't seem to allow calling of functions anymore. 

Say I have some global called "secret", if I try to invoke it: 

```

pipeline {
    agent label:''
    
    stages {
        stage('build') {
            sh 'echo $YEAH $WOO'
        }
        
    }
    
    environment {
        YEAH = "${secret('bing')}" 
        WOO = secret('bop')
    }
    
}

```

I get this in the log: 

```
+ echo org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator@a72085a org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator@a72085a
org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator@a72085a org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator@a72085a
```

which is not what I expected (previously the string interpolation approach would work). 

Note that if I call "secret" direct from within a sh - it works. You could probably use some other global or something to try it.

abayer:
Oy, why is it calling my email replies review by bees!

Anyway - reproduced it, now to figure out what the hell is happening. Yay?

abayer:
So that's weird - `currentBuild.getNumber()` works, but `someFunction()` doesn't. Buh. Will poke at it.

abayer:
Oh, duh. It's `methodMissing` issues. Grr. I'll figure it out today.

abayer:
Attempted fix up. Lemme know how it works for you.

michaelneale:
:bee: tested with a global: https://github.com/michaelneale/annoying/blob/master/vars/secret.groovy and the pipeline above, and it works as I expect. 

nice

PR #23, opened 2016-09-23
Force GIT_COMMITTER_* env vars

abayer:
This is still not working right on ci.jenkins.io, so bugger it, let's hardcode it.


abayer:
Argh. I already had this and it didn't work anyway.

PR #24, opened 2016-09-26
Experimenting to speed up build.

abayer:
Also fix a test while I'm here. `@ClassRule` is magic.

cc @reviewbybees esp @rsandell @stephenc @batmat


abayer:
Ignore the jenkins.ci.cloudbees.com failure - that whole system was barfing.

PR #25, opened 2016-09-30
[FIXED JENKINS-37788] Use isLiteral instead of isConstant.

abayer:
[JENKINS-37788](https://issues.jenkins-ci.org/browse/JENKINS-37788)

cc @reviewbybees esp @rsandell @michaelneale


michaelneale:
:bee: nice one (as we discussed)

PR #26, opened 2016-10-05
[JENKINS-38564] API to convert json step blob to step syntax (and back) - one step at a time

rsandell:
@reviewbybees

### DIFF
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
@@ -37,6 +37,7 @@ import org.codehaus.groovy.control.CompilationUnit
 import org.codehaus.groovy.control.CompilerConfiguration
 import org.codehaus.groovy.control.SourceUnit
 import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTPipelineDef
+import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStep
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution
 import org.jenkinsci.plugins.workflow.cps.CpsScript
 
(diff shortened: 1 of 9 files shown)


rsandell:
Need some help with unit test scenarios that are expected to go bad

abayer:
fwiw, I've disabled the busted jenkins.ci.cloudbees.com job and am re-running the ci.jenkins.io one - I think that's transient noise, but I could be wrong.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
```diff
@@ -149,6 +150,31 @@ public class Converter {
         return model[0];
     }
 
+    public static List<ModelASTStep> scriptToPlainSteps(String script) {
```
abayer: So no complaints here, but I'm wondering if we can make this (and the related changes in `ModelParser` generic - i.e., parameterize the root element for parsing, rather than going into a hardcoded method. Just a thought.

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/endpoints/ModelConverterAction.java
```diff
@@ -142,6 +148,105 @@ public HttpResponse doToJson(StaplerRequest req) {
...
+
+        try {
+            List<ModelASTStep> steps = Converter.scriptToPlainSteps(groovyAsString);
+            JSONArray array = new JSONArray();
+            for (ModelASTStep step : steps) {
+                array.add(step.toJSON());
+            }
+            result.accumulate("result", "success");
+            if (array.size() == 1) {
+                result.accumulate("json", array.get(0)); //TODO decide if consistently returning an array is better
```
abayer: :ant: I'd say always return an array. Consistency and all that.

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/endpoints/ModelConverterAction.java
```diff
@@ -142,6 +148,105 @@ public HttpResponse doToJson(StaplerRequest req) {
...
+            } else {
+                result.accumulate("json", array);
+            }
+
+        } catch (MultipleCompilationErrorsException e) {
+            result.accumulate("result", "failure");
```
abayer: At this point, we may want to abstract the error handling logic out, since it's basically the same for all `Jenkinsfile` -> JSON conversations (and all JSON -> `Jenkinsfile` conversions have the same error handling themselves).

abayer:
Ok, something is failing consistently here in https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/JENKINS-38564/2/testReport/org.jenkinsci.plugins.pipeline.modeldefinition/EnvironmentTest/nonLiteralEnvironment/ - so :bug: until that's fixed.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
rsandell: So what would then be the return type?

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/endpoints/ModelConverterAction.java
rsandell: Done

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/endpoints/ModelConverterAction.java
rsandell: Done

rsandell:
Last build succeeded but

    Could not update commit status. Message: GitHub API rate limit exceeded

abayer:
wooo. =) Well, I'm a :bee: now anyway.

michaelneale:
nice work!

rsandell:
@reviewbybees done

abayer:
Will merge late today when I'm done flying and driving around Central
Europe. =)

PR #27, opened 2016-10-05
[FIXED JENKINS-37824] Adding job property-related functionality

abayer:
[JENKINS-37824](https://issues.jenkins-ci.org/browse/JENKINS-37824)

- [x] Implement runtime translation and execution.
- [x] Implement Groovy and JSON parsing.
- [x] Implement validation.
- [x] Test validation failure cases.
- [x] More elaborate success case tests.

cc @reviewbybees esp @oleg-nenashev @rsandell @michaelneale @HRMPW  @stephenc


src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -179,5 +206,45 @@ public class Utils {
...
+     * @param excludedSymbols Optional list of symbol names to exclude from the cache.
+     * @return A {@link LoadingCache} for looking up types from symbols.
+     */
+    static generateTypeCache(Class<? extends Descriptor> type, List<String> excludedSymbols = []) {
+        return CacheBuilder.newBuilder()
+            .expireAfterWrite(10, TimeUnit.MINUTES)
+            .build(new CacheLoader<Object, Map<String, String>>() {
+            @Override
+            Map<String, String> load(Object key) throws Exception {
+                return populateTypeCache(type, excludedSymbols)
```
rsandell: this does look strange. the load method is usually called when a specific key is missing in the cashe, but you aren't using the key.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -179,5 +206,45 @@ public class Utils {
...
+     * Actually populates the type cache.
+     *
+     * @param type The {@link Descriptor} class whose extensions we want to find.
+     * @param excludedSymbols Optional list of symbol names to exclude from the cache.
+     * @return A map of symbols or class names to class names.
+     */
+    private static Map<String,String> populateTypeCache(Class<? extends Descriptor> type, List<String> excludedSymbols = []) {
+        Map<String,String> knownTypes = [:]
+
+        ExtensionList.lookup(type).each { t ->
```
rsandell: `collect` seems more appropriate here than `each`

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: There were Reasons. =) I can't remember the exact reasoning at the time, but the idea was to take advantage of thread safety, periodic refreshes, etc with `LoadingCache`, but load everything into it at once rather than re-querying `ExtensionList` every time.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: In some cases, I'm adding two entries to the map for a single `Extension` - i.e., if the `Extension` has a `Symbol`, I'm adding the symbol *and* the class name as keys.

stephenc:
:bee:

SYNTAX.md
```diff
@@ -203,6 +204,55 @@ postBuild {
...
+* *Required*: No
+* *Parameters*: None
+* *Takes a Closure*: Yes
+* *Closure Contents*: A sequence of one or more job property configurations, using `@Symbol` names for constructors.
+    * Note that `[$class: 'Foo', arg1: 'something', ...]` syntax can not be used, only `booleanParam(...)` and the like.
+    * Note that the `parameters` and `pipelineTriggers` `@Symbol`s cannot be used here directly.
+* *Examples*:
+
+```
+jobProperties {
```
michaelneale: this is what it is called elsewhere - it just looks funny here. If this is the accepted term, that is ok. Always confusing "properties" vs "parameters". Also we say "job" here but elsewhere say pipeline. Would "pipelineProperties" make more sense? settings?

SYNTAX.md
```diff
@@ -203,6 +204,55 @@ postBuild {
...
+    string(defaultValue: '', description: '', name: 'SOME_STRING')
+}
+```
+
+### Job Properties
+* *Description*:
+* *Required*: No
+* *Parameters*: None
+* *Takes a Closure*: Yes
+* *Closure Contents*: A sequence of one or more job property configurations, using `@Symbol` names for constructors.
```
michaelneale: People won't be able to use the snippetizer for this will they...

SYNTAX.md
abayer: Consistency - they're job properties. Pipelines are a job type. Therefore, etc. I see no reason to make up a new name.

SYNTAX.md
abayer: They can, yes. The snippetizer will try to generate `@Symbol` based output first and only fall back on the old syntax if no symbol is available.

michaelneale:
I think this LTGM. Not 100% sold on "jobProperties" when we say pipeline all over, but eh... 


:bee:

SYNTAX.md
oleg-nenashev: yep. And it works like a charm

SYNTAX.md
```diff
@@ -203,6 +204,55 @@ postBuild {
...
+* *Examples*:
+
+```
+triggers {
+    cron('@daily')
+}
+```
+
+### Build Parameters
+* *Description*:
```
oleg-nenashev: I'm not sure what this empty field means

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -57,6 +56,14 @@ import java.lang.reflect.ParameterizedType;
 @SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
 public class Utils {
 
+    public static Map<String,String> blockedStepsBase() {
+        return [
+            "stage":      'The stage step cannot be used in Declarative Pipelines',
```
oleg-nenashev: Does it mean that we cannot have conditional stages in Declarative pipeline? E.g. "publish when a user sets up the Publish checkbox". Smells like a design 🐛 for me

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -127,6 +134,26 @@ public class Utils {
     }
 
     /**
+     * Finds the parameterized type argument for a {@link MethodsToList} class and returns it.
+     *
+     * @param c A class.
+     * @return The parameterized type argument for the class, if it's a {@link MethodsToList} class. Null otherwise.
```
oleg-nenashev: Would be nice to have since Javadoc for such methods since they are explicitly allowed for use in Pipeline

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -127,6 +134,26 @@ public class Utils {
...
+     */
+    @Whitelisted
+    public static Class<Describable> getMethodsToListType(Class c) {
+        Class retClass
+        c.genericInterfaces.each { Type t ->
+            if (t instanceof ParameterizedType) {
+                if (t.rawType.equals(MethodsToList.class) && t.getActualTypeArguments().first() instanceof Class) {
```
oleg-nenashev: Is there a way to have type without ActualTypeArguments? first() may throw NPE in such case IIRC

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -179,5 +206,52 @@ public class Utils {
...
 
+    /**
+     * Creates and sets the loading for a cache of {@link Describable}s descending from the given descriptor type.
+     *
+     * @param type The {@link Descriptor} class whose extensions we want to find.
+     * @param includeClassNames Optionally include class names as keys. Defaults to false.
+     * @param excludedSymbols Optional list of symbol names to exclude from the cache.
+     * @return A {@link LoadingCache} for looking up types from symbols.
+     */
+    static generateTypeCache(Class<? extends Descriptor> type, boolean includeClassNames = false,
```
oleg-nenashev: 🐜 Usages of this cache do not perform cache line invalidation, which may be a problem for dynamic plugin loading. Not sure if we care about it during the running build

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTMethodArg.groovy
```diff
@@ -0,0 +1,44 @@
...
+
+import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator
+
+/**
+ * A marker for classes that can serve as the argument for a method, either as part of a {@link ModelASTKeyValueOrMethodCallPair} or
+ * on its own in a list.
+ *
+ * @author Andrew Bayer
+ */
+public interface ModelASTMethodArg {
```
oleg-nenashev: 🐜 Undocumented interface with Object returns. Good opportunity for misusages

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTMethodArg.groovy
```diff
@@ -0,0 +1,44 @@
...
+    public String toGroovy()
+
+    public Object toJSON()
```
oleg-nenashev: And no option to throw the conversion expression, right?

src/test/resources/json/simpleParameters.json
```diff
@@ -0,0 +1,45 @@
+{"pipeline": {
```
oleg-nenashev: 🐜 Maybe you want to add license headers to JSONs as well. NIT for me

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidator.groovy
```diff
@@ -273,9 +285,168 @@ class ModelValidator {
...
     }
 
+    public boolean validateElement(@Nonnull ModelASTMethodCall meth) {
+        boolean valid = true
+        if (ModelASTMethodCall.blockedSteps.keySet().contains(meth.name)) {
+            errorCollector.error(meth,
+                "Invalid step '${meth.name}' used - not allowed in this context - ${ModelASTMethodCall.blockedSteps.get(meth.name)}")
+            valid = false
+        }
+        if (Jenkins.getInstance() != null) {
```
oleg-nenashev: I doubt this check is really required in Pipeline stuff now

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/JSONParser.groovy
```diff
@@ -170,6 +188,108 @@ class JSONParser {
...
+            errorCollector.error(pair, "Invalid value type")
+        }
+
+        return pair
+    }
+
+    public @CheckForNull ModelASTMethodCall parseMethodCall(Object o) {
+        ModelASTMethodCall meth = new ModelASTMethodCall(o)
+        if (o instanceof JSONObject) {
+            meth.name = o.getString("name")
```
oleg-nenashev: No field existence check

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/JSONParser.groovy
```diff
@@ -170,6 +188,108 @@ class JSONParser {
         return branch
     }
 
+    public @CheckForNull ModelASTJobProperties parseJobProperties(JSONObject j) {
```
oleg-nenashev: Design 🐜. I am a bit aware about having parse methods without error propagation via exceptions.

SYNTAX.md
abayer: Means I forgot to add the description!

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Currently we cannot. There's a JIRA open for that.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Does since actually make sense here? Hrm.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: I...don't think there is? But might be worth adding a `?`.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Yeah, my attitude is that the type cache gets refreshed every 10 minutes and that's good enough for me.

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTMethodArg.groovy
abayer: Oops. Will clean up.

PR #30, opened 2016-10-14
[JENKINS-37792] Per stage post build steps

rsandell:
@reviewbybees

### DIFF
--- a/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTBuildConditionsContainer.groovy
+++ b/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTBuildConditionsContainer.groovy
@@ -28,35 +28,57 @@ import groovy.transform.EqualsAndHashCode
 import groovy.transform.ToString
 import net.sf.json.JSONArray
 import net.sf.json.JSONObject
-import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
-import org.jenkinsci.plugins.pipeline.modeldefinition.model.Notifications
-import org.jenkinsci.plugins.pipeline.modeldefinition.model.PostBuild
-import org.jenkinsci.plugins.pipeline.modeldefinition.model.StepsBlock
 import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator
 
 /**
- * Represents a list of {@link BuildCondition} and {@link StepsBlock} pairs to be called, depending on whether the build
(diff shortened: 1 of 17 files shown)


abayer:
Yaaaay! =) Reviewing now!

rsandell:
I was planning to clean up all those `c.delegate = script ....` parts as well, but not sure if that should go into this PR or another one. This PR adds yet another one of those and I'm not a fan of redundant code :)

src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTBuildConditionsContainer.groovy
```diff
@@ -0,0 +1,84 @@
...
+        JSONArray a = new JSONArray()
+        conditions.each { c ->
+            a.add(c.toJSON())
+        }
+
+        return new JSONObject()
+                .accumulate("conditions", a)
+    }
+
+    protected void _validate(ModelValidator validator) {
```
abayer: I was going to ask why `_validate` and not just `super.validate(...)` calls, but I'm guessing this is to make sure it actually works right for each extension of the class?

abayer: Works for me.

src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/PostStageTest.java
```diff
@@ -0,0 +1,236 @@
...
+
+    @Test
+    public void failingNotifications() throws Exception {
+        expect(Result.FAILURE, "failingNotifications")
+                .logContains("hello", "Post stage", "farewell", "Error in stage post:", "I AM HERE STILL")
+                .logNotContains("world").go();
+    }
+
+
+    //Helper methods and builders, could possibly be moved to something more central for reuse
```
abayer: Ok, I change my mind, :bug: until this awesome stuff is moved to `AbstractModelDefTest`. =)

abayer:
Now to persuade you to go rewrite all the other tests to use this goodness later... =) :bee:

rsandell:
@reviewbybees done

marcusphi:
It seems this doesn't work all the way. 
The tests in this PR are too simplistic. If you would add a test of case as described in https://issues.jenkins.io/browse/JENKINS-57801 I think you would find that it fails.

PR #31, opened 2016-10-19
[JENKINS-39109] Add a configuration option for what label to use for docker agents

rsandell:
This adds a folder property and a global configuration where the administrator,
or someone configuring a folder, can specify what label expression to use with
a pipeline with `agent docker: "x"` and omitting the label.
The folder property can be specified on any folder in the hierarchy and the closest
parent to the run is consulted first and then up the tree, lastly the global configuration
is consulted for a value.
An extension point is also introduced with specified ordinals for the existing extensions
with ample room to spare in between in case something wants to jump in between.
Lastly a step that returns the value of the extension point that first returns a non null value.


rsandell:
@reviewbybees

abayer:
ping @stephenc @michaelneale in particular

rsandell:
I have a bit more motivation behind it in the ticket https://issues.jenkins-ci.org/browse/JENKINS-39109

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/config/GlobalConfig.java
```diff
@@ -0,0 +1,78 @@
...
+import org.jenkinsci.Symbol;
+import org.kohsuke.stapler.DataBoundSetter;
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
+ * The system config.
+ *
+ * For example the system level {@link DockerLabelProvider}.
+ */
+@Extension @Symbol("pmd")
```
abayer: :bug: `pmd` is already a thing, so I don't think we should be using it as a `Symbol`.

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/config/FolderConfig.java
```diff
@@ -0,0 +1,112 @@
...
+    public String getDockerLabel() {
+        return dockerLabel;
+    }
+
+    @DataBoundSetter
+    public void setDockerLabel(String dockerLabel) {
+        this.dockerLabel = dockerLabel;
+    }
+
+    @Extension @Symbol("pmd")
```
abayer: See other `pmd` `Symbol` concern. =)

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/config/GlobalConfig.java
rsandell: How about `pipeline-model`?

abayer:
Looks good other than the `Symbol` concern, but I want to see others' thoughts on the concept too.

src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/config/GlobalConfig.java
rsandell: Declarative inception ;) Using the word declarative in a declarative configuration that generates configuration for a declarative plugin.

abayer: Touché! Let's see if anyone else has naming suggestions, but given that, I guess `pipeline-model` may be the best bet.

stephenc:
So :bee: on the code... :ant: on us once again adding another plugin implementation of hierarchical configuration which really should be handled by core / a plugin (cc @HRMPW)

jglick:
Merge conflicts already? Life in the fast line here.

michaelneale:
Live fast or die trying? 

For extra points, see if you can have merge conflicts with yourself (working across machines). It's awesome.

stephenc:
Merge conflicts because bobby's changes were made on the pre-ast module split code base

abayer:
Tentative :bee: (i.e., assuming tests pass)

abayer:
re-:bee:ing 'cos I forgot the symbols too. =)

rsandell:
@reviewbybees done

PR #32, opened 2016-10-20
[FIXED JENKINS-38993] Add deterministic ordering of conditions.

abayer:
[JENKINS-38993](https://issues.jenkins-ci.org/browse/JENKINS-38993)
Ordering is:
* Always
* Changed
* Aborted/Failure/Success/Unstable

This is all done via ordinals.

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
@@ -23,6 +23,7 @@
  */
 package org.jenkinsci.plugins.pipeline.modeldefinition.model;
 
+import hudson.ExtensionComponent;
 import hudson.ExtensionList;
 import hudson.ExtensionPoint;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted;
@@ -47,8 +48,6 @@ public abstract class BuildCondition implements Serializable, ExtensionPoint {
     @Whitelisted
     public abstract boolean meetsCondition(WorkflowRun r);
 
-    public abstract double getOrdinal();
-
     /**
      * All the registered {@link BuildCondition}s.
      */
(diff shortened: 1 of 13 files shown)


stephenc:
:bee:

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
```diff
@@ -43,13 +47,40 @@
     @Whitelisted
     public abstract boolean meetsCondition(WorkflowRun r);
 
+    public abstract double getOrdinal();
```
rsandell: Why not use the ordinal that `@Extension` already provides?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
```diff
@@ -43,13 +47,40 @@
...
+
+        List<String> toSort = new ArrayList<>(conditionMethods.keySet());
+
+        Collections.sort(toSort, new Comparator<String>() {
+            @Override
+            public int compare(String c1, String c2) {
+                BuildCondition firstCondition = conditionMethods.get(c1);
+                BuildCondition secondCondition = conditionMethods.get(c2);
+
+                if (secondCondition == null) {
```
rsandell: :bug: if both are `null` you should return `0`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
abayer: ...good question. I straight-up didn't think about that.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
abayer: Good point.

abayer:
@rsandell Thoughts now?

rsandell:
:bee:

PR #33, opened 2016-10-20
[FIXED JENKINS-39011] Error out at parse time when pipeline step is n…

abayer:
…ested.

[JENKINS-39011](https://issues.jenkins-ci.org/browse/JENKINS-39011)

We don't want to support nesting the pipeline step within other blocks
- until JENKINS-38152, we never noticed that this was actually
possible, since the parse-time validation just ignored any Jenkinsfile
without a root-level pipeline step, but with JENKINS-38152, we not
only do another parse round but need to grab the stages from the model
to attach to the run, so we hit an NPE when there's a pipeline step
nested under other blocks.

So - this will now catch when the pipeline step is nested within
another block and give an error message at parse-time, rather than
NPEing at runtime. There is a valid use case for wanting to wrap the
entire build in something like a timeout or timestamper, but we're
going to address that with JENKINS-37823.

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -112,6 +112,19 @@ class ModelParser {
         }
     }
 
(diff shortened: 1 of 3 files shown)


stephenc:
:bee:

abayer:
ping @rsandell =)

rsandell:
:bee:

PR #34, opened 2016-10-21
A secret step

rsandell:
That can decrypt a string literal.
The Snippetizer page provides a way to encrypt a plain text string.

This can be considered a PoC for how I interpreted the examples of this step.

@reviewbybees 

The test case `worksInEnvironment` currently fails, I'm trying to figure out why. Somehow the descriptor is complaining that the parameter is of the wrong type when called from inside `environment {}`

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/SecretStep.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/SecretStep.java
@@ -94,7 +94,7 @@ public class SecretStep extends AbstractStepImpl implements Serializable {
 
         @Override
         public Step newInstance(@CheckForNull StaplerRequest req, @Nonnull JSONObject formData) throws FormException {
-            if (req.findAncestor(Snippetizer.class) != null) {
+            if (req != null && req.findAncestor(Snippetizer.class) != null) {
                 String text = formData.optString("text");
                 if (text != null) {
                     try {
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/SecretStep.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/SecretStep.java
@@ -1,0 +1,152 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/SecretStep.java
```diff
@@ -0,0 +1,152 @@
...
+
+        @Nonnull
+        @Override
+        public String getDisplayName() {
+            return "Provide an encrypted secret that will be decrypted in runtime";
+        }
+
+        @Override
+        public Step newInstance(@CheckForNull StaplerRequest req, @Nonnull JSONObject formData) throws FormException {
+            if (req.findAncestor(Snippetizer.class) != null) {
```
jglick: Are you sure you do not just want to declare the parameter to be of type `Secret`? Would need a little work in `structs` but probably easy enough.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/SecretStepTest.java
```diff
@@ -0,0 +1,106 @@
...
+        j.assertBuildStatusSuccess(j.waitForCompletion(build));
+        j.assertLogContains("FOO is " + plainText, build);
+    }
+
+    @Test
+    public void worksInEnvironment() throws Exception {
+        String plainText = "Bobby`s lil secret";
+        String script = "pipeline {\n" +
+                "    environment {\n" +
+                "        FOO = secret('"+encrypt(plainText)+"')\n" +
```
jglick: This is not being run as a regular `Step` in a `stage`, so I have no idea how this is implemented. Does not look like a bug in `SecretStep` per se.

abayer:
fwiw, I think `foo = "${secret('whatever')}"` would work. But we do need something better here. That said, I'm not sure if we actually want `secret` to be a `Step` and not just a method?

abayer:
Ok, found the problem - `PropertiesToMapTranslator.methodMissing(String, args)` was just passing `args` on untouched, but in every other place, I massage that to take the first element if it's a one-length array. So - fixed that, and fixed the `nonLiteralEnvironment` test accordingly. So - passes!

rsandell:
I guess we now need to figure out if this is what @kohsuke intended :)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/SecretStep.java
```diff
@@ -0,0 +1,152 @@
...
+
+    public static final class StepExecutionImpl extends AbstractSynchronousNonBlockingStepExecution<String> {
+        private static final long serialVersionUID = 1L;
+
+        @Inject
+        private SecretStep step;
+
+        @Override
+        public String run() throws Exception {
+            return DescriptorImpl.decrypt(step.getText());
```
jglick: This will leave the plaintext in `program.dat`. Also it does nothing to prevent you from accidentally disclosing the secret in the log.

rsandell: Well the intention of this step as I understand it is to avoid having plain text secrets in your repo.
Whatever happens to it on the Jenkins master is of a lot lower concern. And afaiu the same this could be said about `withCredentials`.

rsandell:
@jglick Implementing credentials is a different task from this one

jglick:
> the same this could be said about `withCredentials`

No, that step both masks log occurrences, and keeps the secret encrypted in `program.dat`—check its test coverage.

> Implementing credentials is a different task from this one

My argument is that implementing Credentials is the only reasonable way to accomplish the goal here. You want to not merely load some passwords in Groovy, you want interoperability with the wide array of plugins which expect secrets to be passed in as a `credentialsId` argument, and would not accept a “raw” password.

rsandell:
So descision is that this approach is not feasable.

PR #35, opened 2016-10-22
[JENKINS-38433] Make the agent section pluggable.

abayer:
[JENKINS-38433](https://issues.jenkins-ci.org/browse/JENKINS-38433)
- [x] Actually get it working.
- [x] Split the `DeclarativeAgent*` classes into their own plugin.
- [x] Figure out if we can avoid the full `workflow-cps` dependency for `pipeline-model-declarative-agent`. (EDIT: Aaaaactually we need that dependency for writing the `DeclarativeAgentScript` implementations anyway, so, hey, we'll keep the dependency!)
- [x] Decide if we want to go with Snippetizer support for `DeclarativeAgent` (spoiler: probably not) (EDIT: Not. Doesn't make a lot of sense, does it?)
- [x] Javadocs! Lots of them. 
- [x] Tests - we've already got tests for basic agent functionality, so we just need to add new tests what's new - validation, ordering, a custom agent type...

The DeclarativeAgent\* classes are in a separate plugin so that
they can be depended on by other plugins with minimal transitive
dependencies in the process.

cc @reviewbybees - especially looking for input from @stephenc @rsandell and @jglick

### DIFF
--- a/pipeline-model-api/pom.xml
+++ b/pipeline-model-api/pom.xml
@@ -28,7 +28,6 @@
     <groupId>org.jenkinsci.plugins</groupId>
     <artifactId>pipeline-model-parent</artifactId>
     <version>0.5-SNAPSHOT</version>
-    <relativePath>../pom.xml</relativePath>
   </parent>
 
   <groupId>org.jenkinsci.plugins</groupId>
--- a/pipeline-model-api/pom.xml
+++ b/pipeline-model-api/pom.xml
@@ -27,7 +27,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.13</version>
+    <version>2.17</version>
     <relativePath/>
   </parent>
 
@@ -36,7 +36,7 @@
   <version>0.5-SNAPSHOT</version>
   <packaging>hpi</packaging>
   <name>Pipeline: Model API</name>
-  <description>An opinionated, declarative Pipeline</description>
+  <description>Model API for Declarative Pipeline</description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Model+Definition+Plugin</url>
 
   <licenses>
@@ -90,23 +90,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <configuration>
-          <systemProperties>
-            <property>
-              <name>hudson.udp</name>
-              <value>33849</value>
-            </property>
-          </systemProperties>
-          <argLine>-Xmx4g -XX:MaxPermSize=256m</argLine>
-          <reuseForks>true</reuseForks>
-          <forkCount>0.5C</forkCount>
-          <!-- TODO: I hate the retrying tests but given the nondeterministic whackiness of JENKINS-37101... -->
-          <rerunFailingTestsCount>3</rerunFailingTestsCount>
-        </configuration>
-      </plugin>
-
     </plugins>
   </build>
 
--- a/pipeline-model-api/pom.xml
+++ b/pipeline-model-api/pom.xml
@@ -25,41 +25,18 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.jenkins-ci.plugins</groupId>
-    <artifactId>plugin</artifactId>
-    <version>2.17</version>
-    <relativePath/>
+    <groupId>org.jenkinsci.plugins</groupId>
+    <artifactId>pipeline-model-parent</artifactId>
+    <version>0.5-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
   </parent>
 
   <groupId>org.jenkinsci.plugins</groupId>
   <artifactId>pipeline-model-api</artifactId>
-  <version>0.5-SNAPSHOT</version>
   <packaging>hpi</packaging>
   <name>Pipeline: Model API</name>
   <description>Model API for Declarative Pipeline</description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Model+Definition+Plugin</url>
-
-  <licenses>
-    <license>
-        <name>MIT</name>
-        <url>http://opensource.org/licenses/MIT</url>
-        <distribution>repo</distribution>
-    </license>
-  </licenses>
-
-  <developers>
-    <developer>
-      <id>abayer</id>
-      <name>Andrew Bayer</name>
-    </developer>
-  </developers>
-
-  <scm>
-    <connection>scm:git:git://git@github.com/jenkinsci/pipeline-config-plugin.git</connection>
-    <developerConnection>scm:git:git@github.com:jenkinsci/pipeline-config-plugin.git</developerConnection>
-    <url>https://github.com/jenkinsci/pipeline-config-plugin</url>
-    <tag>HEAD</tag>
-  </scm>
 
   <dependencies>
     <dependency>
@@ -69,53 +46,5 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-  
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.codehaus.gmaven</groupId>
-        <artifactId>gmaven-plugin</artifactId>
-        <version>1.4</version>
-        <configuration>
-          <providerSelection>1.8</providerSelection>
-        </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>generateStubs</goal>
-              <goal>compile</goal>
-              <goal>generateTestStubs</goal>
-              <goal>testCompile</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
-    </plugins>
-  </build>
-
-  <properties>
-    <jenkins.version>2.7.1</jenkins.version>
-    <java.level>7</java.level>
-    <groovy.version>2.4.7</groovy.version>
-  </properties>
-  
-  <repositories>
-    <repository>
-      <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
-      <releases>
-        <enabled>true</enabled>
-      </releases>
-      <snapshots>
-        <enabled>true</enabled>
-      </snapshots>
-    </repository>
-  </repositories>
-  <pluginRepositories>
-    <pluginRepository>
-      <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
-    </pluginRepository>
-  </pluginRepositories>
   
 </project>
--- a/pipeline-model-declarative-agent/pom.xml
+++ b/pipeline-model-declarative-agent/pom.xml
@@ -28,7 +28,6 @@
     <groupId>org.jenkinsci.plugins</groupId>
     <artifactId>pipeline-model-parent</artifactId>
     <version>0.5-SNAPSHOT</version>
-    <relativePath>../pom.xml</relativePath>
   </parent>
 
   <groupId>org.jenkinsci.plugins</groupId>
--- a/pipeline-model-declarative-agent/pom.xml
+++ b/pipeline-model-declarative-agent/pom.xml
@@ -36,7 +36,7 @@
   <version>0.5-SNAPSHOT</version>
   <packaging>hpi</packaging>
   <name>Pipeline: Declarative Agent API</name>
-  <description>An opinionated, declarative Pipeline</description>
+  <description>API for declarative agents in Declarative Pipeline</description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Model+Definition+Plugin</url>
 
   <licenses>
@@ -89,23 +89,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <configuration>
-          <systemProperties>
-            <property>
-              <name>hudson.udp</name>
-              <value>33849</value>
-            </property>
-          </systemProperties>
-          <argLine>-Xmx4g -XX:MaxPermSize=256m</argLine>
-          <reuseForks>true</reuseForks>
-          <forkCount>0.5C</forkCount>
-          <!-- TODO: I hate the retrying tests but given the nondeterministic whackiness of JENKINS-37101... -->
-          <rerunFailingTestsCount>3</rerunFailingTestsCount>
-        </configuration>
-      </plugin>
-
     </plugins>
   </build>
 
--- a/pipeline-model-declarative-agent/pom.xml
+++ b/pipeline-model-declarative-agent/pom.xml
@@ -25,41 +25,19 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.jenkins-ci.plugins</groupId>
-    <artifactId>plugin</artifactId>
-    <version>2.13</version>
-    <relativePath/>
+    <groupId>org.jenkinsci.plugins</groupId>
+    <artifactId>pipeline-model-parent</artifactId>
+    <version>0.5-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
   </parent>
 
   <groupId>org.jenkinsci.plugins</groupId>
   <artifactId>pipeline-model-declarative-agent</artifactId>
-  <version>0.5-SNAPSHOT</version>
   <packaging>hpi</packaging>
   <name>Pipeline: Declarative Agent API</name>
   <description>API for declarative agents in Declarative Pipeline</description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Model+Definition+Plugin</url>
 
-  <licenses>
-    <license>
-        <name>MIT</name>
-        <url>http://opensource.org/licenses/MIT</url>
-        <distribution>repo</distribution>
-    </license>
-  </licenses>
-
-  <developers>
-    <developer>
-      <id>abayer</id>
-      <name>Andrew Bayer</name>
-    </developer>
-  </developers>
-
-  <scm>
-    <connection>scm:git:git://git@github.com/jenkinsci/pipeline-config-plugin.git</connection>
-    <developerConnection>scm:git:git@github.com:jenkinsci/pipeline-config-plugin.git</developerConnection>
-    <url>https://github.com/jenkinsci/pipeline-config-plugin</url>
-    <tag>HEAD</tag>
-  </scm>
 
   <dependencies>
     <dependency>
@@ -68,53 +46,5 @@
       <version>2.18</version>
     </dependency>
   </dependencies>
-  
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.codehaus.gmaven</groupId>
-        <artifactId>gmaven-plugin</artifactId>
-        <version>1.4</version>
-        <configuration>
-          <providerSelection>1.8</providerSelection>
-        </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>generateStubs</goal>
-              <goal>compile</goal>
-              <goal>generateTestStubs</goal>
-              <goal>testCompile</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
-    </plugins>
-  </build>
-
-  <properties>
-    <jenkins.version>2.7.1</jenkins.version>
-    <java.level>7</java.level>
-    <groovy.version>2.4.7</groovy.version>
-  </properties>
-  
-  <repositories>
-    <repository>
-      <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
-      <releases>
-        <enabled>true</enabled>
-      </releases>
-      <snapshots>
-        <enabled>true</enabled>
-      </snapshots>
-    </repository>
-  </repositories>
-  <pluginRepositories>
-    <pluginRepository>
-      <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
-    </pluginRepository>
-  </pluginRepositories>
-  
+    
 </project>
--- a/pipeline-model-declarative-agent/pom.xml
+++ b/pipeline-model-declarative-agent/pom.xml
@@ -32,10 +32,10 @@
   </parent>
 
   <groupId>org.jenkinsci.plugins</groupId>
-  <artifactId>pipeline-model-definition</artifactId>
+  <artifactId>pipeline-model-declarative-agent</artifactId>
   <version>0.5-SNAPSHOT</version>
   <packaging>hpi</packaging>
-  <name>Pipeline: Model Definition</name>
+  <name>Pipeline: Declarative Agent API</name>
   <description>An opinionated, declarative Pipeline</description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Model+Definition+Plugin</url>
 
@@ -60,6 +60,14 @@
     <url>https://github.com/jenkinsci/pipeline-config-plugin</url>
     <tag>HEAD</tag>
   </scm>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-cps</artifactId>
+      <version>2.18</version>
+    </dependency>
+  </dependencies>
   
   <build>
     <plugins>
@@ -100,135 +108,7 @@
 
     </plugins>
   </build>
-  
-  <dependencies>
-    <dependency>
-      <groupId>org.jenkinsci.plugins</groupId>
-      <artifactId>pipeline-model-api</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkinsci.plugins</groupId>
-      <artifactId>pipeline-model-declarative-agent</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-job</artifactId>
-      <version>2.6</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-multibranch</artifactId>
-      <version>2.8</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-scm-step</artifactId>
-      <version>2.2</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-basic-steps</artifactId>
-      <version>2.1</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-durable-task-step</artifactId>
-      <version>2.4</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>pipeline-stage-step</artifactId>
-      <version>2.2</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>docker-workflow</artifactId>
-      <version>1.8</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>structs</artifactId>
-      <version>1.5</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>script-security</artifactId>
-      <version>1.22</version>
-    </dependency>
 
-    <!-- JSON schema stuff -->
-    <dependency>
-      <groupId>com.github.fge</groupId>
-      <artifactId>json-schema-validator</artifactId>
-      <version>2.0.4</version>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.datatype</groupId>
-      <artifactId>jackson-datatype-json-org</artifactId>
-      <version>2.2.3</version>
-    </dependency>
-
-    <!-- TEST deps --> 
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>git</artifactId>
-      <version>2.6.0</version>
-      <scope>test</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>org.apache.httpcomponents</groupId>
-          <artifactId>httpclient</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-
-    <dependency>
-      <groupId>org.jenkins-ci.main</groupId>
-      <artifactId>jenkins-test-harness-tools</artifactId>
-      <version>2.0</version>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-cps-global-lib</artifactId>
-      <version>2.3</version>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>git</artifactId>
-      <version>2.6.0</version>
-      <classifier>tests</classifier>
-      <scope>test</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>org.apache.httpcomponents</groupId>
-          <artifactId>httpclient</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-scm-step</artifactId>
-      <version>2.2</version>
-      <classifier>tests</classifier>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-        <groupId>org.jenkins-ci.plugins.workflow</groupId>
-        <artifactId>workflow-support</artifactId>
-        <version>2.2</version>
-        <classifier>tests</classifier>
-        <scope>test</scope>
-    </dependency>
-    
-  </dependencies>
-  
   <properties>
     <jenkins.version>2.7.1</jenkins.version>
     <java.level>7</java.level>
--- a/pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
+++ b/pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
@@ -24,21 +24,19 @@
 
 package org.jenkinsci.plugins.pipeline.modeldefinition.agent;
 
-import groovy.lang.GroovyCodeSource;
-import groovy.lang.GroovyShell;
 import hudson.ExtensionPoint;
 import hudson.model.AbstractDescribableImpl;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted;
-import org.jenkinsci.plugins.structs.describable.DescribableModel;
 import org.jenkinsci.plugins.workflow.cps.CpsScript;
 import org.jenkinsci.plugins.workflow.cps.CpsThread;
 
-import javax.annotation.Nonnull;
 import java.io.Serializable;
-import java.net.URL;
 
-import static org.apache.commons.lang.exception.ExceptionUtils.getFullStackTrace;
-
+/**
+ * Implementations for {@link DeclarativeAgentDescriptor} - pluggable agent backends for Declarative Pipelines.
+ *
+ * @author Andrew Bayer
+ */
 public abstract class DeclarativeAgent extends AbstractDescribableImpl<DeclarativeAgent> implements Serializable, ExtensionPoint {
 
     /**
--- a/pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
+++ b/pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
@@ -1,0 +1,68 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.pipeline.modeldefinition.agent;
+
+import groovy.lang.GroovyCodeSource;
+import groovy.lang.GroovyShell;
+import hudson.ExtensionPoint;
+import hudson.model.AbstractDescribableImpl;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted;
+import org.jenkinsci.plugins.structs.describable.DescribableModel;
+import org.jenkinsci.plugins.workflow.cps.CpsScript;
+import org.jenkinsci.plugins.workflow.cps.CpsThread;
+
+import javax.annotation.Nonnull;
+import java.io.Serializable;
+import java.net.URL;
+
+import static org.apache.commons.lang.exception.ExceptionUtils.getFullStackTrace;
+
+public abstract class DeclarativeAgent extends AbstractDescribableImpl<DeclarativeAgent> implements Serializable, ExtensionPoint {
+
+    /**
+     * ONLY TO BE RUN FROM WITHIN A CPS THREAD. Parses the script source and loads it.
+     * TODO: Decide if we want to cache the resulting objects or just *shrug* and re-parse them every time.
+     *
+     * @return The script object for this declarative agent.
+     * @throws Exception if the script source cannot be loaded or we're called from outside a CpsThread.
+     */
+    @SuppressWarnings("unchecked")
+    @Whitelisted
+    public DeclarativeAgentScript getScript(CpsScript cpsScript) throws Exception {
+        CpsThread c = CpsThread.current();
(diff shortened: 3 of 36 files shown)


michaelneale:
tests???

abayer:
Oh yeah, I should write some tests. =)

abayer:
Though fwiw, the syntax is identical now to how it was pre-pluggable agent. The tests that are needed aren't syntax checks but verifying that proper ordering occurs.

michaelneale:
@abayer right - fair point. I guess to have test would need a fake extension to test fictitious config for

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
```diff
@@ -0,0 +1,66 @@
...
+import org.jenkinsci.plugins.workflow.cps.CpsThread;
+
+import java.io.Serializable;
+
+/**
+ * Implementations for {@link DeclarativeAgentDescriptor} - pluggable agent backends for Declarative Pipelines.
+ *
+ * @author Andrew Bayer
+ */
+public abstract class DeclarativeAgent extends AbstractDescribableImpl<DeclarativeAgent> implements Serializable, ExtensionPoint {
```
rsandell: Why implement `ExtensionPoint`? It's not common for `Describable` to be that. Normally it's the `descriptor` that is the `ExtensionPoint`

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
```diff
@@ -40,47 +43,39 @@ import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted
...
+    /**
+     * Get the appropriate instantiated {@link DeclarativeAgent} corresponding to our arguments.
+     *
+     * @return The instantiated declarative agent or null if not found.
+     */
     @Whitelisted
-    public boolean hasAgent() {
-        return any || docker != null || label != null
+    public DeclarativeAgent getDeclarativeAgent() {
+        String foundType = DeclarativeAgentDescriptor.orderedNames.find { arguments.containsKey(it) }
```
rsandell: So this makes it so you can no longer have both `label` and `docker`?

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
abayer: That is a mistake. Fixing!

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -231,60 +233,15 @@ public class ModelInterpreter implements Serializable {
...
-    def dockerOrWithout(Agent agent, Closure body) {
-        if (agent.docker != null) {
-            return {
-                script.getProperty("docker").image(agent.docker).inside(agent.dockerArgs, {
-                    body.call()
-                })
-            }
         } else {
-            return {
+            return agent.getDeclarativeAgent().getScript(script).run {
```
rsandell: Does this work when the sandbox is on?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public Closure run(Closure body) {
+        String targetLabel = declarativeAgent.label
+        if (targetLabel == null) {
+            targetLabel = script.dockerLabel()?.trim()
+        }
+        LabelScript labelScript = (LabelScript) Label.DescriptorImpl.instanceForName("label", [label: targetLabel]).getScript(script)
+        return labelScript.run {
```
rsandell: You seem to be breaking the model here and not returning a closure?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Yup - the overwhelming majority of tests are using `CpsScmFlowDefinition`, which has sandbox on. And I'm adding a change to make sure the ones using `CpsFlowDefinition` turn on sandbox too.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
abayer: Ah, but I *am* returning a closure! I'm returning the closure from `LabelScript.run`!

abayer: And yes, I get confused by the `.call()` stuff too. I honestly don't remember exactly why I need it everywhere I've got it, but I do. I'll try to figure that out soon. =)

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
abayer: Yeah, everything I can see is this way - a lot more `Describable`s as `ExtensionPoint`s than `Descriptor`s as `ExtensionPoint`s.

jglick: Yes historically @kohsuke has marked `Describable`s as `ExtensionPoint`s even though logically it is the `Descriptor` which should be, since that is where the `@Extension` would go. “Feel free” to fix it everywhere. Until then, go with the convention.

pipeline-model-declarative-agent/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+    <version>2.13</version>
+    <relativePath/>
+  </parent>
+
+  <groupId>org.jenkinsci.plugins</groupId>
+  <artifactId>pipeline-model-declarative-agent</artifactId>
+  <version>0.5-SNAPSHOT</version>
+  <packaging>hpi</packaging>
+  <name>Pipeline: Declarative Agent API</name>
+  <description>An opinionated, declarative Pipeline</description>
```
jglick: Misleading description of this module.

pipeline-model-declarative-agent/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <systemProperties>
+            <property>
+              <name>hudson.udp</name>
+              <value>33849</value>
```
jglick: You do not need to override all this stuff. It should get inherited from the parent.

pipeline-model-declarative-agent/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+            </property>
+          </systemProperties>
+          <argLine>-Xmx4g -XX:MaxPermSize=256m</argLine>
+          <reuseForks>true</reuseForks>
+          <forkCount>0.5C</forkCount>
+          <!-- TODO: I hate the retrying tests but given the nondeterministic whackiness of JENKINS-37101... -->
+          <rerunFailingTestsCount>3</rerunFailingTestsCount>
```
jglick: Why are you overriding this? It is on by default, and in a better way than you have configured. Best to delete this whole `plugin` section.

pipeline-model-declarative-agent/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+    <jenkins.version>2.7.1</jenkins.version>
+    <java.level>7</java.level>
+    <groovy.version>2.4.7</groovy.version>
+  </properties>
+  
+  <repositories>
+    <repository>
+      <id>repo.jenkins-ci.org</id>
+      <url>http://repo.jenkins-ci.org/public/</url>
+      <releases>
```
jglick: Specifying `releases` and `snapshots` here is unnecessary AFAIK.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Descriptor for {@link DeclarativeAgent}.
+ *
+ * @author Andrew Bayer
+ */
+public abstract class DeclarativeAgentDescriptor extends Descriptor<DeclarativeAgent> {
+
+    /**
+     * The name for this agent type. Should match the {@code Symbol} on the class.
```
jglick: So why not make it non-`abstract`?

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+     * @return The name.
+     */
+    public abstract @Nonnull String getName();
+
+    /**
+     * The full package and class name for the {@link DeclarativeAgentScript} class corresponding to this.
+     *
+     * @return The class name.
+     */
+    public abstract @Nonnull String getDeclarativeAgentScriptClass();
```
jglick: Could make a default implementation that just appends something to the agent class name.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * An ordered (by Extension ordinal, in descending order) list of all descriptors.
+     * @return A list of descriptors.
+     */
+    public static List<DeclarativeAgentDescriptor> getOrderedDescriptors() {
+        List<DeclarativeAgentDescriptor> orderedDescriptors = new ArrayList<>();
+
+        List<ExtensionComponent<DeclarativeAgentDescriptor>> extensionComponents = new ArrayList<>(all().getComponents());
+        Collections.sort(extensionComponents);
```
jglick: Delete this method. `ExtensionList` does this automatically, so `all()` should suffice.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Get a map of name-to-{@link DescribableModel} of all known/registered descriptors.
+     *
+     * @return A map of name-to-{@link DescribableModel}s
+     */
+    public static Map<String,DescribableModel> getDescribableModels() {
+        Map<String,DescribableModel> models = new HashMap<>();
+
+        for (DeclarativeAgentDescriptor d : getOrderedDescriptors()) {
+            Set<String> symbolValues = SymbolLookup.getSymbolValue(d);
```
jglick: Uh, do you not mean to call `getName` here?

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+
+            if (!symbolValues.isEmpty()) {
+                models.put(symbolValues.iterator().next(), new DescribableModel<>(d.clazz));
```
jglick: You should rather loop over all the symbols. Actually easier than what you are doing.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+            }
+
+        }
+
+        return models;
+    }
+
+    /**
+     * Get the map of the subset of descriptors with single-arguments - i.e., "none" and "any".
```
jglick: The naming here seems very misleading. IIUC these are _zero_-argument implementations.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+     * An ordered list of all descriptor names.
+     *
+     * @return A list of names
+     */
+    public static List<String> getOrderedNames() {
+        List<String> orderedNames = new ArrayList<>();
+
+        for (DeclarativeAgentDescriptor d : getOrderedDescriptors()) {
+            Set<String> symbolValues = SymbolLookup.getSymbolValue(d);
+            if (symbolValues != null && !symbolValues.isEmpty()) {
```
jglick: Again, simpler and better to just add all the symbols.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Get the descriptor for a given name or null if not found.
+     *
+     * @param name The name for the descriptor to look up
+     * @return The corresponding descriptor or null if not found.
+     */
+    @Whitelisted
+    public static @Nullable DeclarativeAgentDescriptor byName(@Nonnull String name) {
+        for (DeclarativeAgentDescriptor d : all()) {
+            if (d.getName().equals(name)) {
```
jglick: If you delete `getName` and use symbols directly, you could simplify this to `SymbolLookup.findDescriptor`.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,190 @@
...
+     * @throws Exception
+     */
+    @Whitelisted
+    public static @Nullable DeclarativeAgent instanceForName(@Nonnull String name,
+                                                             Map<String,Object> arguments) throws Exception {
+        DeclarativeAgentDescriptor descriptor = byName(name);
+
+        if (descriptor != null) {
+            if (singleArgModels().keySet().contains(name)) {
+                return descriptor.newInstance(new HashMap<String, Object>());
```
jglick: Uh, what? This makes no sense to me at all. What if the arguments are nonempty?

pipeline-model-definition/pom.xml
```diff
@@ -108,6 +108,11 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.jenkinsci.plugins</groupId>
+      <artifactId>pipeline-model-declarative-agent</artifactId>
+      <version>${project.version}</version>
```
jglick: FWIW I do not recommend releasing reactors of plugins with synchronized versions. But if you really want to, you are doing it wrong here, since the `pipeline-model-declarative-agent` POM uses a fixed `version` rather than one inherited from an aggregator.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
stephenc: Technically, `ExtensionPoint` is just any class that that is extensible. You are supposed to see the actual class/interface for details as to how to register... for a large chunk of them you register by `@Extension` but that is not a requirement for to be called an `ExtensionPoint`... and `Descriptor` already extends `ExtensionPoint`...

Of course the javadoc could make this distinction clearer... but...

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
jglick: Rather roundabout to delegate to `LabelScript`. Easier to say

```groovy
{
  def main = {
    script.getProperty("docker").image(declarativeAgent.docker).inside(declarativeAgent.dockerArgs, {
      body.call()
    })
  }
  if (targetLabel != null) {
    script.node(targetLabel) {main()}
  } else {
    main()
  }
}
```

pipeline-model-declarative-agent/pom.xml
abayer: If we only ever have `release` dependencies, yeah, but I prefer to play it safe and include both.

pipeline-model-declarative-agent/pom.xml
abayer: Done.

pipeline-model-declarative-agent/pom.xml
abayer: Fixed.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Done.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Seems reasonable.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Good point.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Better point. =)

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Agreed.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Are you sure about that? I did some experiments and found that `ExtensionList.lookup(SomeDescriptor.class)` didn't order by ordinals.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Done.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Thought about that - is a good point.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Hrm, thought I'd done something differently here. Fixing.

pipeline-model-definition/pom.xml
abayer: I need to talk to @stephenc - this is his setup. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
abayer: So I'm doing it this way at least in part as an example of how you can combine agent scripts together - but I'll think about it.

pipeline-model-definition/pom.xml
stephenc: @abayer actually it is more your setup... as you set up the parent hierarchy.

I recommend that all poms in this repo have the *exact same version* lock step.That will probably mean that the parent pom will need to be in the hierarchy... which means that we will have to deploy it too

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AgentTest.java
```diff
@@ -126,6 +125,27 @@ public void perStageConfigAgent() throws Exception {
         j.assertLogContains("ONSLAVE=true", b);
     }
 
+    @Test
+    public void agentTypeOrdering() throws Exception {
+        prepRepoWithJenkinsfile("agentTypeOrdering");
```
rsandell:
```java
expect("agentTypeOrdering").logContains("Running in labelAndOtherField with otherField = banana",
                                    "[Pipeline] { (foo)", "ONSLAVE=true").go();
```

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
```diff
@@ -378,6 +378,34 @@ public void globalLibraryObjectMethodCall() throws Exception {
     }
 
     @Test
+    public void unknownAgentType() throws Exception {
+        prepRepoWithJenkinsfile("errors", "unknownAgentType");
```
rsandell:
```java
expect(Result.FAILURE, "errors", "unknownAgentType")
    .logContains("No agent type specified. Must contain one of [otherField, docker, label, any, none]").go();
```

rsandell:
:bee:

abayer:
Because I forgot to remove it! Will do so.

abayer:
ping @jglick - thoughts on the current state?

abayer:
ping @jglick again

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,194 @@
...
+public abstract class DeclarativeAgentDescriptor extends Descriptor<DeclarativeAgent> {
+
+    /**
+     * The name for this agent type. Should match the {@code Symbol} on the class.
+     *
+     * @return The name.
+     */
+    public @Nonnull String getName() {
+        Set<String> symbolValues = SymbolLookup.getSymbolValue(this);
+        return symbolValues.iterator().next();
```
jglick: Will throw a confusing `NoSuchElementException` if you forget the annotation.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * The full package and class name for the {@link DeclarativeAgentScript} class corresponding to this.
+     *
+     * @return The class name.
```
jglick: Specify the default value.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,194 @@
...
+    @Whitelisted
+    public static @Nullable DeclarativeAgent instanceForName(@Nonnull String name,
+                                                             Map<String,Object> arguments) throws Exception {
+        DeclarativeAgentDescriptor descriptor = byName(name);
+
+        if (descriptor != null) {
+            if (zeroArgModels().keySet().contains(name)) {
+                return descriptor.newInstance();
+            } else {
+                return descriptor.newInstance(arguments);
```
jglick: Why does `zeroArgModels` and no-arg `newInstance()` exist? If the arguments list is empty, `newInstance(Map)` would receive an empty map and work the same. IOW why does this need to be a special case?

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,194 @@
...
+    public static ExtensionList<DeclarativeAgentDescriptor> all() {
+        return ExtensionList.lookup(DeclarativeAgentDescriptor.class);
+    }
+
+    /**
+     * Get a map of name-to-{@link DescribableModel} of all known/registered descriptors.
+     *
+     * @return A map of name-to-{@link DescribableModel}s
+     */
+    public static Map<String,DescribableModel> getDescribableModels() {
```
jglick: Could probably be deleted—see comment below.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        return models;
+    }
+
+    /**
+     * Get the map of the subset of descriptors with no additional arguments - i.e., "none" and "any".
+     * @return A map of descriptors with no arguments.
+     */
+    public static Map<String,DescribableModel> zeroArgModels() {
```
jglick: Could probably be deleted—see comment below.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,194 @@
...
+        return new DescribableModel<>(clazz).instantiate(arguments);
+    }
+
+    /**
+     * Creates an instance of the corresponding {@link DeclarativeAgent} with no arguments.
+     *
+     * @return An instantiated {@link DeclarativeAgent}
+     * @throws Exception
+     */
+    public DeclarativeAgent newInstance() throws Exception {
```
jglick: Could probably be deleted—see comment below.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
jglick: Exactly, should be deleted.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Get the descriptor for a given name or null if not found.
+     *
+     * @param name The name for the descriptor to look up
+     * @return The corresponding descriptor or null if not found.
+     */
+    @Whitelisted
+    public static @Nullable DeclarativeAgentDescriptor byName(@Nonnull String name) {
```
jglick: Use `@CheckForNull`.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * For a given name and map of arguments, find the corresponding descriptor and return an instance using those arguments.
+     *
+     * @param name The name of the descriptor
+     * @param arguments A map of arguments
+     * @return The instantiated {@link DeclarativeAgent} instance, or null if the name isn't found.
+     * @throws Exception
+     */
+    @Whitelisted
+    public static @Nullable DeclarativeAgent instanceForName(@Nonnull String name,
```
jglick: Use `@CheckForNull`.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
```diff
@@ -40,47 +43,39 @@ import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted
...
     }
 
+    /**
+     * Special constructor for the no-additional-arguments agent types, i.e., none and any
+     */
     @Whitelisted
-    public Agent(boolean b) {
-        this.any = b
+    public Agent(String s) {
+        this.arguments.put(s, "true")
```
jglick: Huh??

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
```diff
@@ -40,47 +43,39 @@ import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted
...
+     *
+     * @return The instantiated declarative agent or null if not found.
+     */
     @Whitelisted
-    public boolean hasAgent() {
-        return any || docker != null || label != null
+    public DeclarativeAgent getDeclarativeAgent() {
+        String foundType = DeclarativeAgentDescriptor.orderedNames.find { arguments.containsKey(it) }
+        if (foundType != null) {
+            return DeclarativeAgentDescriptor.instanceForName(foundType, arguments)
```
jglick: Would be faster, simpler, and more reliable to call `byName` and then `newInstance`, rather than searching the list of extensions twice.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -533,15 +534,36 @@ class ModelValidatorImpl implements ModelValidator {
...
+            Map<String,DescribableModel> zeroArgModels = DeclarativeAgentDescriptor.zeroArgModels()
+            if (!zeroArgModels.containsKey(singleArg.value.getValue())) {
+                errorCollector.error(agent.args, "Invalid argument for agent - '${singleArg.value.toGroovy()}' - must be map of config options or bare ${zeroArgModels.keySet().sort()}.")
+                valid = false
+            }
+        } else if (agent.args instanceof ModelASTNamedArgumentList) {
+            ModelASTNamedArgumentList namedArgs = (ModelASTNamedArgumentList)agent.args
+            List<String> argKeys = namedArgs.arguments.collect { k, v ->
+                k.key
+            }
```
jglick: Could just collapse the two cases here, passing `[:]` in case of `ModelASTSingleArgument`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/Any.java
```diff
@@ -0,0 +1,48 @@
...
+    @DataBoundConstructor
+    public Any() {
+        super(null);
+    }
+
+    @Extension(ordinal = -900) @Symbol("any")
+    public static class DescriptorImpl extends DeclarativeAgentDescriptor {
+        public @Nonnull String getDeclarativeAgentScriptClass() {
+            return "org.jenkinsci.plugins.pipeline.modeldefinition.agent.impl.LabelScript";
+        }
```
jglick: Could now be deleted IIUC.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy
```diff
@@ -47,12 +47,12 @@ public class ClosureModelTranslator implements MethodMissingWrapper, Serializabl
...
      * Placeholder to make sure 'agent none' works.
      */
-    boolean none = false
+    String none = "none"
 
     /**
      * Placeholder to make sure 'agent any' works.
      */
-    boolean any = true
+    String any = "any"
```
jglick: Are these actually being used, or could they be deleted?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
```diff
@@ -0,0 +1,54 @@
...
+                    body.call()
+                }
+            }
+        } else {
+            return {
+                script.node {
+                    body.call()
+                }
+            }
+        }
```
jglick: Could probably be simplified to just

```groovy
{
    script.node(declarativeAgent.label) {
        body.call()
    }
}
```

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
```diff
@@ -0,0 +1,54 @@
...
+
+public class LabelScript extends DeclarativeAgentScript {
+
+    public LabelScript(CpsScript s, DeclarativeAgent a) {
+        super(s, a)
+    }
+
+    @Override
+    public Closure run(Closure body) {
+        if (declarativeAgent?.label != null) {
```
jglick: `declarativeAgent` may not be null here, may it?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelAndOtherFieldAgent.java
```diff
@@ -0,0 +1,67 @@
...
+    public String getLabel() {
+        return label;
+    }
+
+    @Whitelisted
+    public String getOtherField() {
+        return otherField;
+    }
+
+    @Extension(ordinal = 1100) @Symbol("otherField")
```
jglick: Generally clearer to make this a `TestExtension` inside the test suite.

pipeline-model-api/pom.xml
```diff
@@ -25,42 +25,19 @@
...
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.jenkins-ci.plugins</groupId>
-    <artifactId>plugin</artifactId>
-    <version>2.13</version>
-    <relativePath/>
+    <groupId>org.jenkinsci.plugins</groupId>
+    <artifactId>pipeline-model-parent</artifactId>
+    <version>0.5-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
```
abayer: Done.

pom.xml
```diff
@@ -57,25 +64,30 @@
```
abayer: Done.

pom.xml
```diff
@@ -57,25 +64,30 @@
 
   <modules>
     <module>pipeline-model-api</module>
+    <module>pipeline-model-declarative-agent</module>
     <module>pipeline-model-definition</module>
   </modules>
 
   <!-- project is just for aggregating, never as a parent, we force skip to ensure we never deploy or install it -->
```
abayer: Done.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Got a tweak done.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Done.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Where/how?

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: `zeroArgModels` is a convenience method for finding, well, zero-arg models. The problem is that even in the case of zero-arg models, there's still an entry in the arg map - the key identifying the model. And for some reason I can't remember off the top of my head, even passing an empty `Map` still didn't work for instantiating them...

Oh, and `zeroArgModels` is also used at validation time to make sure any identifying-key-only calls are actually using a valid zero-arg model.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
abayer: Okiedokie.

abayer: Hrm, actually, doesn't quite work - I need to find the first descriptor (ordered by ordinal-descending) with its name in the arguments list, not just a `byName`. I'll see if there's something I can do...

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
abayer: huh? The check is different.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/Any.java
abayer: Nope, 'cos `Any` reuses the `Label` script.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy
abayer: Annoyingly, they are still being used - otherwise `agent none` and `agent any` won't work.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
abayer: I feel like there is a way it could be but can't remember off the top of my head.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
abayer: Ah, found an approach.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
jglick: Do not use `@Nullable` unless you really know what it means, and even then generally not.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
jglick: Yeah, either way the `@DataBoundConstructor` will be called with a null argument.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -54,13 +54,17 @@
      */
     public @Nonnull String getName() {
```
jglick: Do you even still need `getName` at all?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
jglick: I mean get rid of the special-case handling of zero-argument models, except to the extent that the AST is trivially different so `argKeys` would have to be set to `[:]`.

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
abayer: Okiedokie, so I'll remove `@Nullable`, since I've just been cargo-culting it. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
abayer: Ah. I'll think as to whether I can do that and still achieve the goals...

abayer:
@jglick Yeah, forgot to nuke `getOrderedDescriptors()` but have done so now. And we do still want `getName()` so that it *can* optionally be overridden with something that isn't a `@Symbol` for some reason (and so that we don't have to throw the symbol-handling stuff everywhere).

pipeline-model-declarative-agent/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentDescriptor.java
```diff
@@ -0,0 +1,175 @@
...
+ * @author Andrew Bayer
+ */
+public abstract class DeclarativeAgentDescriptor extends Descriptor<DeclarativeAgent> {
+
+    /**
+     * The name for this agent type. Should match the {@code Symbol} on the class.
+     *
+     * @return The name.
+     */
+    public @Nonnull String getName() {
```
jglick: I do feel like everywhere you are calling this method it is a mistake, since you are either

* enumerating all names, in which case you could equally enumerate all symbols
* checking for the presence of a particular name, i.e., first symbol, in which case there is some bug involving people using a subsequent symbol

Probably not a big deal, just a hint.

abayer: I'll think on it a bit more.

rsandell:
:bee:

PR #36, opened 2016-10-24
[FIXED JENKINS-37823] Added "wrappers" section.

abayer:
[JENKINS-37823](https://issues.jenkins-ci.org/browse/JENKINS-37823)

Trying to figure out what more tests are needed - note that this doesn't accept the `wrap` step, since it requires a `node` context and we explicitly reject any step that requires a `node`. 

Also cc @kuhnroyal since I know you're interested in this functionality. =)

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/WrappersToMapTranslator.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/WrappersToMapTranslator.groovy
@@ -1,0 +1,82 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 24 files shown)


rsandell:
:bee:

rsandell:
:bee:

michaelneale:
Very nice - not 100% sold on the name "wrappers" but that is a minor thing. Can this be done at stage level too potentially?

pipeline-model-definition/src/test/resources/multipleWrappers.groovy
```diff
@@ -0,0 +1,41 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    agent none
+    wrappers {
```
abayer: Find me a better one? =)

michaelneale:
LGTM

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/WrappersToMapTranslator.groovy
```diff
@@ -0,0 +1,82 @@
...
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.Wrappers
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.WrappersToMap
+import org.jenkinsci.plugins.workflow.cps.CpsScript
+
+/**
+ * Translates a closure containing a sequence of wrapper calls into a {@link WrappersToMap} implementation.
+ *
+ * @author Andrew Bayer
+ */
+public class WrappersToMapTranslator implements MethodMissingWrapper, Serializable {
```
rsandell: I'm starting to see a lot of similar translators, we should extract common code into something that can be inherited and reused instead of copy and paste.

rsandell:
:bee:

PR #38, opened 2016-10-24
[JENKINS-39216] Add "dockerfile" agent backend

abayer:
[JENKINS-39216](https://issues.jenkins-ci.org/browse/JENKINS-39216)

Not happy with the naming - but I needed a map key/value pair for this, so went with `dockerfileImage` as the unique identifier for this backend. The actual image name will be that string plus the `Dockerfile`'s sha1 hash. I'm open to better naming.

cc @reviewbybees esp @michaelneale @HRMPW @stephenc @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
@@ -48,7 +48,7 @@ public class DockerPipelineFromDockerfileScript extends DeclarativeAgentScript {
             try {
                 def hash = Utils.stringToSHA1(script.readFile(declarativeAgent.getDockerfileAsString()))
                 def imgName = "${hash}"
-                def img = script.getProperty("docker").build(declarativeAgent.getDockerfileAsString())
+                def img = script.getProperty("docker").build(imgName, "-f ${declarativeAgent.getDockerfileAsString()}")
                 img.inside(declarativeAgent.dockerArgs, {
                     body.call()
                 })
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
@@ -1,0 +1,60 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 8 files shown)


abayer:
Note that this is on top of #35 - so it's got a lot more commits in it currently than are actually relevant. First relevant commit is da91209. Also, the instances we're building the PRs on on ci.jenkins.io don't have docker, so you can't actually see the results of the new test. =) But I promise, it passes! Or at least it did for me once!

rsandell:
How about `Dockerfile` as the name? It's quite distinguished from `docker` and people should recognise it as something else and related to an actual Dockerfile.

rsandell:
Really hard to revew untill pluggable agent backend has been merged, so the diff is understandable.

michaelneale:
exciting!

Ok does 
`agent dockerfileImage:"something"`
mean that it is looking for a file called `something`? 

Given 99.9% of cases Dockerfiles are called dockerfile, and if not, they would be "Dockerfile.whatever" - this doesn't make sense. 

Also the term "dockerfileImage" is super confusing. 

My suggestions: 

`agent dockerfile:'Dockerfile'`

simple, explicit.

michaelneale:
I am ok with this but would prefer `agent dockerfile:'something'` but otherwise looks great to me.

abayer:
Ok, I can do that - alternatively, we might be able to do `agent dockerfile:true` as well as `agent dockerfile:'/path/to/a/Dockerfile'`, but we can't just do `agent dockerfile` 'cos I'm not adding any more zero-arg models due to the pain inherent in supporting them in the backend.

abayer:
@michaelneale @rsandell @HRMPW Updated to use `dockerfile` as the key, and to allow either a `String` for a different filename to use or a `boolean` (honestly, it could be `true`, it could be `false`, it could be `12573456987` - any value other than a `String`) to just use `Dockerfile`.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
```diff
@@ -0,0 +1,60 @@
...
+        if (targetLabel == null) {
+            targetLabel = script.dockerLabel()?.trim()
+        }
+        LabelScript labelScript = (LabelScript) Label.DescriptorImpl.instanceForName("label", [label: targetLabel]).getScript(script)
+        return labelScript.run {
+            script.checkout script.scm
+            try {
+                def hash = Utils.stringToSHA1(script.readFile(declarativeAgent.getDockerfileAsString()))
+                def imgName = "${hash}"
+                def img = script.getProperty("docker").build(imgName)
```
rsandell: aren't you just building the dockerfile in the cwd here? or am I missing some docker knowledge with the hash magic?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AgentTest.java
```diff
@@ -139,13 +139,31 @@ public void agentTypeOrdering() throws Exception {
...
         prepRepoWithJenkinsfile("agentAnyInStage");
-
         WorkflowRun b = getAndStartBuild();
         j.assertBuildStatusSuccess(j.waitForCompletion(b));
         j.assertLogContains("[Pipeline] { (foo)", b);
         j.assertLogContains("THIS WORKS", b);
     }
 
+    @Test
+    public void fromDockerfile() throws Exception {
```
rsandell: should add a test with a dockerfile in a different location as well.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
abayer: Yup, that'd be a mistake. Fixing.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AgentTest.java
abayer: Good point.

michaelneale:
very nice - thanks! this is so great actually. Really great.

PR #39, opened 2016-10-25
Remove whitelisting

abayer:
We apparently don't need it any more, so let's get rid of it.

Note that this is on top of #35.

cc @reviewbybees esp @jglick @stephenc @rsandell


rsandell:
:bee:

rsandell:
:bee:

PR #40, opened 2016-10-25
[FIXED JENKINS-39244, FIXED JENKINS-39245] Add env and tools to stages

abayer:
[JENKINS-39244](https://issues.jenkins-ci.org/browse/JENKINS-39244), [JENKINS-39245](https://issues.jenkins-ci.org/browse/JENKINS-39245)

Adds support for tools and environment sections inside individual stages.

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/EnvironmentTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/EnvironmentTest.java
@@ -55,6 +55,16 @@ public class EnvironmentTest extends AbstractModelDefTest {
     }
 
     @Test
+    public void environmentInStage() throws Exception {
+        prepRepoWithJenkinsfile("environmentInStage");
+
+        WorkflowRun b = getAndStartBuild();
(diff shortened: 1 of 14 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/EnvironmentTest.java
```diff
@@ -55,6 +55,16 @@ public void simpleEnvironment() throws Exception {
     }
 
     @Test
+    public void environmentInStage() throws Exception {
+        prepRepoWithJenkinsfile("environmentInStage");
```
rsandell:
```java
expect("environmentInStage").logContains("[Pipeline] { (foo)", "FOO is BAR").go()
```

rsandell:
Only one missed debug print otherwise a be, but :bug:

rsandell:
:bee:

rsandell:
it would be nice to get this merged asap so I don't have to redo half of the work I'm currently doing when it is.

rsandell:
Merged, cause I need it and I see no problems with it and no one else has for 24h.
We can always tweak later before 1.0

PR #41, opened 2016-10-26
[JENKINS-37781] Adding Stage.when to specify if a stage should run

rsandell:


rsandell:
@reviewbybees 

I need to write a few more tests and I'm not sure the json structure is good

rsandell:
@reviewbybees 
@abayer

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTScriptBlock.java
```diff
@@ -7,13 +7,17 @@
  */
 public class ModelASTScriptBlock extends ModelASTStep {
     public ModelASTScriptBlock(Object sourceLocation) {
+        this(sourceLocation, "script");
+    }
+
+    protected ModelASTScriptBlock(Object sourceLocation, String name) {
```
abayer: :bug: I'm not a huge fan of overloading `ModelASTScriptBlock` - I'd rather keep it as just for `script` blocks. I'd say go with making a new class that has this as the main constructor, with `ModelASTWhen` and `ModelASTScriptBlock` both extending that new class.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -95,49 +95,50 @@ public class ModelInterpreter implements Serializable {
...
-                                                                try {
-                                                                    for (int ni = 0; ni < postClosures.size(); ni++) {
-                                                                        setUpDelegate(postClosures.get(ni)).call()
-                                                                    }
-                                                                } catch (Exception e) {
-                                                                    script.echo "Error in stage post: ${e.getMessage()}"
-                                                                    script.getProperty("currentBuild").result = Result.FAILURE
-                                                                    if (firstError == null) {
-                                                                        firstError = e
+                                runStageOrNot(thisStage, firstError) {
```
abayer: :bug: Flip with the line below. We always want to have the `stage` even if we don't do anything in it. This is so that we have a consistent, predictable execution model.

rsandell: If we somehow could mark the stage as `NOT_BUILT` then I would be all for it.

svanoort: Andrew has the right of it, unfortunately -- if the actual stage set changes, there's not any way to easily map one run with another.   

Stage View particularly suffers from this case - https://issues.jenkins-ci.org/browse/JENKINS-33430 -- I suspect that the final Blue Ocean implementation may handle it differently if needed, but probably will also not like it.

abayer: Yup - pinging in @vivek for his thoughts

vivek: I agree with @abayer. Blueocean has does have projected graph thats union with last successful run in case current build fails. This is different, I think its going to create more problems. We should preserve the stages and mark them NOT_BUILT, so that visualization can render it correctly.

abayer: Yeah, I'd like this to wait and be on top of #47 using `SKIPPED_FOR_CONDITIONAL`...

abayer:
Actually, get the merge conflict resolved and we'll merge this. I'll integrate it into #47

rsandell:
@abayer merge conflict resolved

abayer:
:bee: pending build success

PR #44, opened 2016-11-01
Explain that steps section is needed in stages.

abayer:
We're not supporting backwards compatibility for syntax that we change
before 1.0, but we should give clear error messages explaining what
has changed. At this point, this is the only compatibility-breaking
syntactic change.

cc @reviewbybees esp @rsandell @HRMPW @michaelneale


rsandell:
:bee:

PR #45, opened 2016-11-02
JENKINS-39394 - Get rid of notifications, rename postBuild to post

abayer:
[JENKINS-39394](https://issues.jenkins-ci.org/browse/JENKINS-39394)

Gets rid of `notifications`, renames `postBuild` to `post`, gives meaningful errors if it sees those old section names used, and while we're here, moves `ast-schema.json` to `pipeline-model-api`, since that's where we're defining the schema's contents in the first place!

cc @reviewbybees esp @rsandell @michaelneale @HRMPW @i386


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ASTSchema.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public String getDisplayName() {
+        return null;
+    }
+
+    @SuppressWarnings("unused")
+    public void doJson(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException {
+        rsp.serveFile(req, getClass().getResource("/ast-schema.json"), TimeUnit2.DAYS.toMillis(1));
```
rsandell: is it possible to do more here? if jenkins is restarted with a newer version we would want to make sure they aren't using a day old cache.

rsandell: maybe redirecting to the actual resource path?

rsandell:
:bee:

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ASTSchema.java
abayer: Honestly, there's probably no need for the caching at all. Not sure why I even did that. Removing it - I don't know if there's a better way than `serveFile` to get at the file, though. If there is, I'd love to hear it. =)

rsandell:
:bee:

rsandell:
:bee:

abayer:
ping @HRMPW @i386 @michaelneale - I'd like your signoff before merging this.

i386:
Can't really speak for the code but as the PR is described LGTM

abayer:
This is now on top of #51 so don't worry about reviewing it until that's merged.

michaelneale:
nice - is wiki updated with this (or will be)?

abayer:
@michaelneale Wiki will be updated when this gets released.

abayer:
ping @rsandell - could use a re-review now that all its ancestors are merged into master.

rsandell:
:bee:

kuhnroyal:
Can the `currentBuild` result be accessed somehow?
I tried `currentBuild.result` but that's always `null`.

abayer:
Null means success - it'll be null until either the end of the build or something causes it to get set to aborted/failed/unstable.

kuhnroyal:
@abayer thanks that makes sense

PR #46, opened 2016-11-03
[JENKINS-37778] Support credentials in environment

rsandell:
Adds a `credentials` step mostly suited for use in `environment {}` but can be used outside as well.
Supports a slightly opinionated way of binding credentials.

@reviewbybees esp. @stephenc and @abayer 

Some things left todo

- [x] A few more tests
- Ssh credentials doesn't have a binding, so skipping that.


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -82,80 +79,84 @@ public class ModelInterpreter implements Serializable {
             // Entire build, including notifications, runs in the withEnv.
             withEnvBlock(root.getEnvVars()) {
                 inWrappers(root.wrappers) {
-                // Stage execution and post-build actions run in try/catch blocks, so we still run post-build actions
```
rsandell: Hmm, I had some issues getting the indentation correct at first, through that I had fixed it but apparently not.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/util/HasArchived.java
```diff
@@ -0,0 +1,123 @@
...
+import org.hamcrest.TypeSafeMatcher;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+/**
+ * {@link Matcher} for {@link Run}s containing an archived file
+ */
+public class HasArchived extends TypeSafeMatcher<Run> {
```
rsandell: This might be useful to contribute to the Jenkins test harness?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/util/HasArchived.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private boolean matchesFile(VirtualFile file, VirtualFile root) throws IOException {
+        if (file.isDirectory()) {
+            for (VirtualFile child : file.list()) {
+                if (matchesFile(child, root)) {
+                    return true;
+                }
+            }
+        } else if (nameMatcher.matches(fullName(file, root))) {
+            if (contentMatcher.matches(file.open())) {
```
rsandell: I would have liked to do a check here if the contentMatcher's type was `InputStream` or `VirtualFile` to either send in `file` or `file.open()` depending. But I couldn't figure out how.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
rsandell: Ah, I actually accidentally fixed a previous indentation error.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/CredentialWrapperStep.java
```diff
@@ -0,0 +1,112 @@
...
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nonnull;
+import java.io.Serializable;
+import java.util.Collections;
+
+/**
+ * Returns a {@link CredentialWrapper} for simplified credentials binding in {@code environment {}}
+ */
+public class CredentialWrapperStep extends AbstractStepImpl implements Serializable {
```
rsandell: So while working on this I've been more and more wanting to have this as a function to only be callable within `environment{}` so I don't pollute the step namespace with something as "common" as `credentials` but I don't know how to achieve that.

abayer: Interesting question. Probably don't make it a `Step`, then, and let me think about where it'd fit in...

abayer: Oh, I think I have an idea - add it as a method on `Environment`!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
```diff
@@ -106,11 +109,23 @@ public class Stage implements NestedModel, Serializable {
...
+        return environment.findAll{k, v -> !(v instanceof CredentialWrapper)}.collect { k, v ->
             "${k}=${v}"
         }
     }
 
+    @Whitelisted
+    @Nonnull
+    Map<String, CredentialWrapper> getEnvCredentials() {
+        Map<String, CredentialWrapper> m = [:]
+        environment.each {k, v ->
```
jglick: Is this code implicitly `@NonCPS`?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/CredentialsBindingHandler.java
```diff
@@ -0,0 +1,211 @@
...
+        @Override
+        public Class<? extends StandardCredentials> type() {
+            return StandardUsernamePasswordCredentials.class;
+        }
+
+        @Nonnull
+        @Override
+        public List<Map<String, Object>> getWithCredentialsParameters(String credentialsId) {
+            Map<String, Object> map = new HashMap<>();
+            map.put("$class", "UsernamePasswordBinding");
```
jglick: Safer to specify a FQN: `UsernamePasswordBinding.class.getName()`

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
rsandell: I don't know, it seems to be since it works as well as the code in the method above that has been working for a long time.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/CredentialWrapperStepTest.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Test
+    public void mixedEnvInFolder() throws Exception {
+
+        expect("credentials", "mixedEnv").runFromRepo(false).inFolder(folder)
+                .logContains("SOME_VAR is SOME VALUE",
+                             "INBETWEEN is Something in between",
+                             "OTHER_VAR is OTHER VALUE")
+                .archives("cred1.txt", mixedEnvInFolderCred1Secret)
```
stephenc: well lookupStores is supposed to give you all the stores that are contributing to credentials available at the context object... so you would need to find the one that is bound to the folder if you want to inject into the folder level store

rsandell: OK, I thought they would be returned in order of priority for where to look.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
rsandell: Yes, they are in `src/main/groovy` and not in `resources`

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/CredentialWrapperStep.java
rsandell: OK, so due to the way that delegate resolution is setup for all the closures "just adding a method" didn't work, so I had to do some hackery in `methodMissing`

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/CredentialWrapperStepTest.java
rsandell: Turned out that a good old fashioned `configRoundTrip` was needed and then just fetch the store from the property directly.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
```diff
@@ -143,11 +143,22 @@ public class Root implements NestedModel, Serializable {
      */
     @Whitelisted
     List<String> getEnvVars() {
-        return environment.collect { k, v ->
+        return environment.findAll{k, v -> !(v instanceof CredentialWrapper)}.collect { k, v ->
             "${k}=${v}"
         }
     }
 
+    @Whitelisted
```
abayer: No need for `Whitelisted` - ive got that pending PR to remove it everywhere, so let's not add it in new places too. =)

abayer:
One question - what validation are we doing in re the credentials method?

rsandell:
@abayer since I'm not sure what you are talking about; probably nothing ;)

There are some error states that tries to throw exceptions with human readable messages and explanations.

abayer:
@rsandell At runtime or validation time?

rsandell:
@abayer runtime

rsandell:
@abayer since the json/jenkinsfile endpoints aren't connected to a Job, we can't really do a lookup for the credentials to see if they exists or are of the correct type. There is a risk of too many false negatives.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
rsandell: merge that pr, then I'll fix it with the other merge problems that I'll get.

abayer:
@rsandell Gotcha.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
rsandell: Done

abayer:
Ok, then :bee:

rsandell:
Why isn't the build starting?

PR #47, opened 2016-11-04
[FIXED JENKINS-38153] Added synthetic stages using TagsAction.

abayer:
[JENKINS-38153](https://issues.jenkins-ci.org/browse/JENKINS-38153)
[JENKINS-39522](https://issues.jenkins-ci.org/browse/JENKINS-39522)

This is downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/24

We're doing two things here - adding metadata to mark "synthetic" stages (i.e., putting all the possibly-time-consuming work we do that isn't inside a user-defined stage into automatically generated stages that Blue Ocean will treat differently than user-defined stages) and marking stages that are skipped due to failure of earlier stages with metadata for their status.

@vivek - the artifacts for this are available for a Blue Ocean PR to depend on as well - version `0.6-synthetic-stage-SNAPSHOT`.

cc @reviewbybees esp @jglick @rsandell @svanoort @vivek


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -198,6 +204,44 @@ public class Utils {
         r.addAction(new ExecutionModelAction(stages))
     }
 
+    static boolean withinAStage() {
+        CpsThread thread = CpsThread.current()
+        CpsFlowExecution execution = thread.execution
+
+        DepthFirstScanner scanner = new DepthFirstScanner();
```
abayer: Ok, changing. Can you explain why one will give correct results and the other won't so I understand for future purposes?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -198,6 +204,44 @@ public class Utils {
...
+     */
+    static void markSyntheticStage(String stageName, String context) {
+        CpsThread thread = CpsThread.current()
+        CpsFlowExecution execution = thread.execution
+
+        FlowNode currentNode = execution.currentHeads.find { n ->
+            n?.displayName?.equals(stageName)
+        }
+
+        if (currentNode.getAction(TagsAction.class) == null) {
```
svanoort: 🐛 Broken persistence.  The right way:
 
```java
myAction = currentNode.getAction(TagsAction.class)
if (myAction == null) { 
  myAction = new TagsAction(); 
  myAction.addTag( stuff );  // persists it automatically
} else {
   myAction.addTag(stuff); 
  currentNode.save();  // persists
}
```

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -82,37 +82,40 @@ public class ModelInterpreter implements Serializable {
...
-                                                } catch (Exception e) {
-                                                    script.echo "Error in stages execution: ${e.getMessage()}"
-                                                    script.getProperty("currentBuild").result = Result.FAILURE
-                                                    if (firstError == null) {
-                                                        firstError = e
-                                                    }
-                                                } finally {
-                                                    // And finally, run the post stage steps.
-                                                    List<Closure> postClosures = thisStage.satisfiedPostStageConditions(root, script.getProperty("currentBuild"))
+                                script.stage(thisStage.name) {
```
svanoort: why why why are we doing whitespace changes

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Don't I need to add `myAction` to `currentNode.actions`?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -295,4 +301,69 @@ public void globalLibrarySuccess() throws Exception {
...
+        j.assertLogContains("[Pipeline] { (foo)", b);
+        j.assertLogContains("hello", b);
+        j.assertLogContains("[Pipeline] { (Post Build Actions)", b);
+        j.assertLogContains("I AM A POST-BUILD", b);
+
+        FlowExecution execution = b.getExecution();
+
+        Collection<FlowNode> heads = execution.getCurrentHeads();
+
+        DepthFirstScanner scanner = new DepthFirstScanner();
```
svanoort: Bloody NO.  Again: BlockHoppingLinearScanner.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Because I'm fixing some indent problems while I'm here so as not to get so hateful at everything. =)

vivek:
@abayer noted. thanks

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
svanoort: see the current version, I was typing fast.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
abayer: Fun fact: this doesn't work with `LinearBlockHoppingScanner` - just tried it. We're post-build at this point, so as is said in that class's Javadoc, "ONLY use this with nodes inside the flow graph". =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
svanoort: also note: it needs to be exactly this way.  If you add the empty action first, you're persisting to disk 2x -- that's expensive, don't do it.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
svanoort: I could, but it's in their JavaDocs, I encourage you to look at those ;)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Yup, tried that and it didn't work, so switching back to how I was doing things but with an added `currentNode.save()` call.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: I looked, and didn't understand. =P

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Okiedokie.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Or rather, I didn't understand what the differences between the two *mean* in this case.

abayer:
Added JENKINS-39522 (marking skipped stages) as well.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -205,6 +212,71 @@ public class Utils {
...
+
+        LinearBlockHoppingScanner scanner = new LinearBlockHoppingScanner();
+
+        FlowNode currentNode = scanner.findFirstMatch(execution, isStageWithOptionalName(stageName))
+
+        TagsAction tagsAction = currentNode.getAction(TagsAction.class)
+        if (tagsAction == null) {
+            tagsAction = new TagsAction()
+            tagsAction.addTag(tagName, tagValue)
+            currentNode.addAction(tagsAction)
```
rsandell: why don't you save the `currentNode` here?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -82,37 +79,40 @@ public class ModelInterpreter implements Serializable {
...
+                                                        script.getProperty("currentBuild").result = Result.FAILURE
+                                                        if (firstError == null) {
+                                                            firstError = e
+                                                        }
+                                                    } finally {
+                                                        // And finally, run the post stage steps.
+                                                        List<Closure> postClosures = thisStage.satisfiedPostStageConditions(root, script.getProperty("currentBuild"))
                                                         catchRequiredContextForNode(thisStage.agent != null ? thisStage.agent : root.agent, false) {
                                                             if (postClosures.size() > 0) {
                                                                 script.echo("Post stage")
```
rsandell: shouldn't post stage then also be a synthetic stage within the current stage?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
```diff
@@ -44,6 +45,19 @@ public class DockerPipelineFromDockerfileScript extends DeclarativeAgentScript {
         }
         LabelScript labelScript = (LabelScript) Label.DescriptorImpl.instanceForName("label", [label: targetLabel]).getScript(script)
         return labelScript.run {
+            if (!Utils.withinAStage()) {
+                script.stage(StageTagsMetadata.agentSetup()) {
+                    Utils.markSyntheticStage(StageTagsMetadata.agentSetup(), StageTagsMetadata.SYNTHETIC_PRE)
+                    buildImage(body).call()
```
rsandell: looks like you are creating a nested stage here? is that intended?

abayer: Nope - only creates a new stage when we aren't already in one.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Ideally, yes, but until Blue Ocean is comfortable with nested stages, we're avoiding them.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: @svanoort told me I don't need to. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
rsandell: Yes, but wont `buildImage(body).call()` enter stages in turn?

abayer: You are correct, sir! Will fix that.

rsandell: Some of the below once probably have the same issue.

abayer: I think the others are ok, but I'll check.

rsandell:
:bee:

pipeline-model-definition/pom.xml
```diff
@@ -69,7 +69,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-api</artifactId>
-      <version>2.6-SNAPSHOT</version>
+      <version>2.6-add-tagsaction-SNAPSHOT</version>
```
svanoort: you can bump to 2.7-add-tagsaction-SNAPSHOT if you're feeling frisky now!  🐐

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
svanoort: @rsandell FYI if curious, b/c `addAction` in `FlowNode` does it automagically by override of `getActions()` with a custom list whose mutators trigger `save()`

svanoort:
🐝  from me now after changes, minus some groovy magic I don't 100% understand

rsandell:
:bee:

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
```diff
@@ -46,8 +46,8 @@ public class DockerPipelineScript extends DeclarativeAgentScript {
         LabelScript labelScript = (LabelScript) Label.DescriptorImpl.instanceForName("label", [label: targetLabel]).getScript(script)
         return labelScript.run {
             if (!Utils.withinAStage()) {
-                script.stage(StageTagsMetadata.agentSetup()) {
-                    Utils.markSyntheticStage(StageTagsMetadata.agentSetup(), StageTagsMetadata.SYNTHETIC_PRE)
+                script.stage(SyntheticStageNames.agentSetup()) {
+                    Utils.markSyntheticStage(SyntheticStageNames.agentSetup(), Utils.getSyntheticStageMetadata().pre)
```
rsandell: we could potentially implement a new pipeline step in `tags-metadata` that allows "normal" pipeline users to tag their own stages.

rsandell:
:bee:

rsandell:
:bee:

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageNames.groovy
```diff
@@ -28,22 +28,26 @@ package org.jenkinsci.plugins.pipeline.modeldefinition
 
 public class SyntheticStageNames {
     public static String checkout() {
-        return "Checkout SCM"
+        return "Declarative: Checkout SCM"
```
abayer: Good point, but I really do want to make sure we have uniqueness for the synthetic stages. Any ideas?

abayer:
This is now on top of both #51 and #45 so it won't get merged until those both get merged.

vivek:
@abayer @svanoort blueocean PR for tag action: https://github.com/jenkinsci/blueocean-plugin/pull/608.

jglick:
The apparent use case would be served more simply and with better type safety by defining an `Action` in `pipeline-model-definition` which marks something as synthetic; perhaps also by using a different `Step` rather than `stage`. In any case, `blueocean-pipeline-api-impl` already depends directly on `pipeline-model-definition` so why is any API needed in `workflow-api`?

abayer:
My first attempt at this was exactly that - I added a new `Action` to `pipeline-model-definition`. Sam, in my opinion, rightly pointed out that this is likely to be something that can be of use in other areas, not just `pipeline-model-definition`/Blue Ocean, and proposed adding `TagsAction` instead. I agree with that approach.

Definitely not going to create a new `Step` instead of `stage` - that'll just cause a lot of pain for Blue Ocean for no real benefit.

abayer:
@jglick Note that we're also using `TagsAction` to relay status information - in this case, why a stage was skipped. 

Again, @svanoort has made the case that this is of more general use than just between Declarative and Blue Ocean, and I agree, so it sure feels like `workflow-api` is the right place to put it. I swear we've had this discussion already...

jglick:
> this is of more general use than just between Declarative and Blue Ocean

*Could* be of more general use, but no such more general use has been so much as prototyped AFAICT.

> not going to create a new `Step` instead of `stage` - that'll just cause a lot of pain for Blue Ocean

Really? Seems like the new block would simply be ignored by default, but readable by special-case code, which is sort of what you are looking for here.

To be clear, I do not really have the time to do detailed reviews of the use cases and implementation here, and did not ask to be on the critical path. Since my opinion was requested, I am offering general comments about the API design, which from a casual review seems out of proportion to the actual feature being delivered.

svanoort:
Yes, I agree with the statement that workflow-api is the right place to add custom step metadata -- commenting there.   The alternative is worse:  building custom logic for pipeline model definition that is *not* portable back to standard pipeline.

abayer:
We still want to be actual `stage`s - we just want to treat them specially in reporting in Blue Ocean. Adding a new step just seems silly. And @jglick, since you're on the critical path for https://github.com/jenkinsci/workflow-api-plugin/pull/24, we need that for this and the Blue Ocean PR, and you wanted a full use case before approving https://github.com/jenkinsci/workflow-api-plugin/pull/24, you're on the critical path here too. =)

rsandell:
re:bee:

lieutdan13:
Is there any documentation on how to use this feature? I'm very much interested in how to implement it to see if it meets our needs.

rsandell:
@lieutdan13 The only user facing part of this feature is that BO will mark skipped and failed stages.

PR #48, opened 2016-11-07
Refactoring tests to use the expect(...) tooling.

abayer:
Bobby wrote this nice stuff a few weeks ago and we should really be
using it everywhere here, so, let's use it everywhere here. There are
still a couple weird cases where I couldn't figure out an easy way to
translate the tests into the expect(...) approach, but 95% or so of
the run-from-repo tests are now expect(...)-style.


rsandell:
looks good except for the archived files check that I have a more elaborate implementation of in #46

abayer:
@rsandell Yup, updated to use that.

rsandell:
:bee:

rsandell:
:bee:

PR #49, opened 2016-11-08
Refactoring ModelInterpreter for easier maintenance

abayer:
Moved blocks of logic out into their own methods for actual
stage/post-stage execution, notifications and postBuild.

Switched all ModelInterpreter methods previously returning Closures to
instead call() those Closures. This *drastically* simplified the main
method's readability.

Still need to Javadoc all the ModelInterpreter methods and move a bit
more into separate methods from the main method.

Note that we're still using call() on Closures and returning the
results because...well, otherwise things just don't work. More
experimentation is needed to remember exactly why we need to do this,
but I believe it's to do with getting execution timing right.

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -64,31 +64,18 @@ public class ModelInterpreter implements Serializable {
         Throwable firstError
 
         if (root != null) {
-            def jobProps = []
-
-            if (root.jobProperties != null) {
-                jobProps.addAll(root.jobProperties.properties)
-            }
-            if (root.triggers != null) {
-                jobProps.add(script.pipelineTriggers(root.triggers.triggers))
-            }
-            if (root.parameters != null) {
-                jobProps.add(script.parameters(root.parameters.parameters))
-            }
-            if (!jobProps.isEmpty()) {
-                script.properties(jobProps)
-            }
+            executeProperties(root)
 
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -216,38 +153,57 @@ public class ModelInterpreter implements Serializable {
...
+     * @return The return of the resulting executed closure
+     */
     def withEnvBlock(List<String> envVars, Closure body) {
         if (envVars != null && !envVars.isEmpty()) {
             return {
                 script.withEnv(envVars) {
                     body.call()
                 }
-            }
+            }.call()
```
rsandell: do you really need to `return` if you `call()`?

abayer: Good question - let's find out!

abayer: Apparently yes. Hrm.

rsandell: But that means that you return the return value of the `call` not the closure itself. So that would indicate to me that something is funky somewhere else.

rsandell:
:bee: anyway :)

abayer:
Merging - @rsandell, could you open a JIRA to investigate the `return`/`call()` stuff in more detail?

PR #51, opened 2016-11-10
[FIXED JENKINS-39631] Last step in failed stage has an error.

abayer:
[JENKINS-39631](https://issues.jenkins-ci.org/browse/JENKINS-39631)

cc @reviewbybees esp @vivek @svanoort @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -107,6 +107,7 @@ public class ModelInterpreter implements Serializable {
                                                                 } catch (Exception e) {
                                                                     script.echo "Error in stages execution: ${e.getMessage()}"
                                                                     script.getProperty("currentBuild").result = Result.FAILURE
+                                                                    Utils.attachErrorToStep(e)
                                                                     if (firstError == null) {
                                                                         firstError = e
                                                                     }
@@ -124,6 +125,7 @@ public class ModelInterpreter implements Serializable {
                                                                             } catch (Exception e) {
                                                                                 script.echo "Error in stage post: ${e.getMessage()}"
                                                                                 script.getProperty("currentBuild").result = Result.FAILURE
+                                                                                Utils.attachErrorToStep(e)
                                                                                 if (firstError == null) {
                                                                                     firstError = e
                                                                                 }
@@ -154,6 +156,7 @@ public class ModelInterpreter implements Serializable {
                                 } catch (Exception e) {
                                     script.echo "Error in postBuild execution: ${e.getMessage()}"
                                     script.getProperty("currentBuild").result = Result.FAILURE
+                                    Utils.attachErrorToStep(e)
                                     if (firstError == null) {
                                         firstError = e
                                     }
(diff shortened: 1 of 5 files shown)


i386:
🐛 from the users PoV we are now returning two steps for just the single step that they have defined in the Jenkinsfile. There should be a single failed step.

BTW I am using the [example on the ticket](https://issues.jenkins-ci.org/browse/JENKINS-39631).

jenkins-39463 1 2016-11-11 08-05-05

abayer:
@svanoort Yeah, that's why I changed it a few minutes before you reviewed to get rid of the superfluous `echo` that was causing the confusion, so that the last step in the `stage` is correctly the one that failed. See `BasicModelDefTest.allStagesExist` - it tests using `StatusAndTiming`.

vivek:
@abayer tested, LGTM 🐝

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/PostStageTest.java
```diff
@@ -121,7 +121,7 @@ public void withAllLocalChanged() throws Exception {
     @Test
     public void failingNotifications() throws Exception {
         expect(Result.FAILURE, "failingNotifications")
-                .logContains("hello", "Post stage", "farewell", "Error in stage post:", "I AM HERE STILL")
+                .logContains("hello", "Post stage", "farewell", "I AM HERE STILL")
```
rsandell: This test doesn't really test that a failing notification is communicated any more.

abayer:
Updated on top of the refactoring of `ModelInterpreter` - @rsandell @svanoort @vivek Would appreciate reviews as soon as convenient. Thanks!

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/PostStageTest.java
abayer: Yeah, but that test's gonna end up removed in a bit when #45 comes in anyway, so I just threw up my hands and moved on for now. =)

vivek:
@abayer No extra steps added if stage is first or second, however I notice if failed stage is the last one, there is extra step added to it.

```
pipeline {
    agent label:''
    stages {
        stage ('Build') {
            steps{
                sh 'echo "Building"'
            }
        }
        stage ('Test') {
            steps{
                sh 'echo "Testing"'
            }
        }
        stage ('Deploy') {
            steps{
                sh 'echo1 "Deploying"'
            }
        }
    }
}
```
See the image below, there is extra print step in Deploy stage. Its marked as error step but its a synthetic one added by declarative.

image

cc:@i386

abayer:
Ah, yeah, I forgot that post everything error message. I'll remove it
tomorrow.

vivek:
@abayer 👍

i386:
@abayer @vivek we don't want a synthetic step though?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/PostStageTest.java
rsandell: Not really, because this test PostStage, not post build

rsandell:
:bee:

abayer:
That last `echo` has been removed.

@svanoort Can you re-review, please?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -365,77 +380,74 @@ public class ModelInterpreter implements Serializable {
...
                     }
                 }
-            }
-        } catch (Exception e) {
-            script.echo "Error in notifications execution: ${e.getMessage()}"
-            script.getProperty("currentBuild").result = Result.FAILURE
-            if (firstError == null) {
-                firstError = e
+            } catch (Exception e) {
+                script.getProperty("currentBuild").result = Result.FAILURE
```
svanoort: Be really really cautious with doing this.  I don't see an error here but explicitly setting status can lead to some... odd behaviors.

michaelneale:
:bee:

vivek:
🐝

PR #52, opened 2016-11-14
Add "when" JSON conversion test.

abayer:
We need to make sure we've got a JSON conversion test for `when`, and it's good to have a visible example of the syntax for others to see.

cc @reviewbybees esp @michaelneale @rsandell

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
@@ -304,5 +304,17 @@ public class BasicModelDefTest extends AbstractModelDefTest {
         j.assertBuildStatusSuccess(j.waitForCompletion(b));
         j.assertLogContains("[Pipeline] { (One)", b);
         j.assertLogContains("[Pipeline] { (Two)", b);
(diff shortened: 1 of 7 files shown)


michaelneale:
:bee: when passes. Nice one.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -295,4 +295,14 @@ public void globalLibrarySuccess() throws Exception {
...
+
+    @Test
+    public void basicWhen() throws Exception {
+        prepRepoWithJenkinsfile("basicWhen");
+
+        WorkflowRun b = getAndStartBuild();
+        j.assertBuildStatusSuccess(j.waitForCompletion(b));
+        j.assertLogContains("[Pipeline] { (One)", b);
+        j.assertLogContains("[Pipeline] { (Two)", b);
+    }
```
rsandell: I think you also need to test that a stage is actually also skipped.

abayer: Fair enough. Will add one.

rsandell:
:bee:

PR #53, opened 2016-11-15
Make docker-workflow an optional dependency

abayer:
Prompted by [JENKINS-39723](https://issues.jenkins-ci.org/browse/JENKINS-39723)

Allow installing `pipeline-model-definition` without autoinstalling `docker-workflow` as well. Also makes the `DockerPipeline` and `DockerPipelineFromDockerfile` `Extension`s `optional`, so that they won't have classloading errors if `docker-workflow` isn't installed, and attempts to use `agent docker:...` or `agent dockerfile:...` will error out at validation time.

cc @reviewbybees esp @rsandell @michaelneale

### DIFF
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -100,6 +100,7 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>docker-workflow</artifactId>
       <version>1.9</version>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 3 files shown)


michaelneale:
Ambivalent as to if this is a good idea or not. *shrug emoji*

abayer:
> Ambivalent as to if this is a good idea or not

So here's my reasoning - not everyone's going to use `docker-workflow`. We want to actively encourage developers of similar plugins (i.e., the kubernetes plugin) to write their own `DeclarativeAgent` backends, and not everyone's going to be using containers at all.

michaelneale:
Well the solution based documentation uses `agent docker` a fair bit, so this would mean they have to install something else (which has a slippery name) - I would like @i386 to comment.

i386:
🐛 I think this is a bad idea as people will have to figure out at runtime that `docker-workflow` is needed to run the examples. Now, if we had a `declarative` aggregator that we told everyone to install and it brings in things like `docker-workflow` and `pipeline-model-definition` became the definition for the model I'd like this change better.

rsandell:
I'm ok with this if the documentation is updated to not have docker examples unless explicitly stated that they need another plugin installed.

abayer:
Fair enough - closing this.

jglick:
Optional dependencies are generally evil anyway.

PR #56, opened 2016-11-17
[FIXED JENKINS-39799] Properly handle bad content in post blocks.

abayer:
[JENKINS-39799](https://issues.jenkins-ci.org/browse/JENKINS-39799)

We actually would have errored out properly-ish if it wasn't for the
fact that we ended up with a ModelASTBuildCondition with null
condition and null branch, leading to the NPE during compilation. I
decided to improve the error messaging and make sure we don't attempt
to validate further when condition or branch are null.


rsandell:
:bee:

rsandell:
clicked the wrong "comment" button.

PR #57, opened 2016-11-17
Run empty stages for conditionally-skipped stages.

abayer:
This is just a band-aid until JENKINS-38153 lands with a better
solution involving tags metadata and all that jazz, but this'll at
least make sure our execution plan matches reality for now.

cc @reviewbybees esp @rsandell @HRMPW @i386


i386:
so this goes and creates a flow node for a conditional stage even if it doesn't really get run?

abayer:
Yes, same as we've long been doing for stages after a failed stage.

i386:
Sounds good to me 💯

rsandell:
Well by principal I don't like it. ;)

abayer:
@rsandell =P

PR #58, opened 2016-11-21
[FIXED JENKINS-39923] Add a CLI endpoint for the linter

abayer:
[JENKINS-39923](https://issues.jenkins-ci.org/browse/JENKINS-39923)

Can be used like:

```
ssh -p [sshd port on master] localhost declarative-linter < Jenkinsfile
```

or

```
curl -X POST -H `curl 'JENKINS_URL/crumbIssuer/api/xml?xpath=concat(//crumbRequestField,":",//crumb)'` -F 'jenkinsfile="@Jenkinsfile"' JENKINS_URL/pipeline-model-converter/validate
```
(the `crumbIssuer` bit is needed if your Jenkins master has CRSF enabled, as it should)

- [x] Write the guts.
- [x] Switch to taking a file path as parameter. 
- [x] Decide on permissions.
- [x] Add tests.

Example output:

Invalid:
```
ssh -p 8675 localhost declarative-linter < src/test/resources/errors/emptyAgent.groovy
Errors encountered validating Jenkinsfile:
WorkflowScript: 25: Not a valid section definition: 'agent'. Some extra configuration is required. @ line 25, column 1.
   pipeline {
   ^

WorkflowScript: 25: Missing required section 'agent' @ line 25, column 1.
   pipeline {
   ^
```

Valid:
```
ssh -p 8675 localhost declarative-linter < src/test/resources/basicWhen.groovy
Jenkinsfile successfully validated.
```

cc @reviewbybees esp @rsandell @HRPMW @i386 @michaelneale

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommand.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommand.java
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.pipeline.modeldefinition;
 
 import hudson.Extension;
 import hudson.cli.CLICommand;
+import hudson.cli.util.ScriptLoader;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.apache.commons.io.IOUtils;
@@ -47,32 +48,16 @@ import java.util.List;
 
 @Extension
 public class DeclarativeLinterCommand extends CLICommand {
-    @Argument(metaVar="JENKINSFILE", usage="Path to the Jenkinsfile to validate", required=true)
-    String jenkinsfile;
-
     @Override
     public String getShortDescription() {
         return Messages.DeclarativeLinterCommand_ShortDescription();
     }
 
     protected int run() throws Exception {
-        File jf = new File(jenkinsfile);
-
         int retVal = 0;
         List<String> output = new ArrayList<>();
 
-        String script = null;
-
-        try {
-            FileInputStream fif = new FileInputStream(jf);
-            script = IOUtils.toString(fif);
-        } catch (FileNotFoundException e) {
-            retVal = 1;
-            output.add("Jenkinsfile '" + jenkinsfile + "' does not exist or cannot be read.");
-        } catch (IOException e) {
-            retVal = 1;
-            output.add("IOException reading Jenkinsfile '" + jenkinsfile + "': " + e.getMessage());
-        }
+        String script = script = IOUtils.toString(stdin);
 
         if (script != null) {
             try {
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommand.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommand.java
@@ -34,7 +34,12 @@ import org.codehaus.groovy.control.MultipleCompilationErrorsException;
 import org.codehaus.groovy.control.messages.SyntaxErrorMessage;
 import org.jenkinsci.plugins.pipeline.modeldefinition.endpoints.ModelConverterAction;
 import org.jenkinsci.plugins.pipeline.modeldefinition.parser.Converter;
+import org.kohsuke.args4j.Argument;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


i386:
Holy crap @abayer you are a machine!! ping @michaelneale @PaddyOH this needs to be documented!

i386:
@abayer what does the output look like?

abayer:
Added examples above.

mpailloncy:
Really cool ! thx @abayer

i386:
You know this deserves its own blog post *hint*

abayer:
Not til it's released!

jimklimov:
Looking forward to it... the blog... and release :)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommand.java
```diff
@@ -0,0 +1,107 @@
...
+    protected int run() throws Exception {
+        File jf = new File(jenkinsfile);
+
+        int retVal = 0;
+        List<String> output = new ArrayList<>();
+
+        String script = null;
+
+        try {
+            FileInputStream fif = new FileInputStream(jf);
```
rsandell: Since this code is running on the master ir would only work if you specify a path to a Jenkinsfile on the master

rsandell: Example https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/cli/CreateJobCommand.java

rsandell: https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/cli/UpdateJobCommand.java

rsandell: https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/cli/GroovyCommand.java

rsandell: Otherwise you would need to send a FileCallable back to the client, and then it won't work with ssh.

rsandell:
:bug: for only working on master.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommand.java
abayer: Hrm - my test doesn't work with the `loadScript()` approach...

abayer: Yeah, looks like I gotta switch to just `stdin` if I want to work via `CLICommandInvoker` or ssh. Ah well!

abayer:
Switched back to taking the content from stdin due to Bobby noticing the obvious that I missed. =)

rsandell:
:bee: so far, but some form of permission check needs to be done, otherwise it could be used to probe the master for installed plugins/vulnerabilities.

abayer:
@rsandell Any idea of what the permission should be?

stephenc:
If you can't build the job you shouldn't need to lint... so how does `Job.BUILD` sound

rsandell:
@stephenc there is no job context in the command, so no job to check permission on. That is also why we can't do proper linting on whether a credential id exists or not.
Maybe a new permission that is implied by `EXTENDED_READ`?

abayer:
Ick, strong -1 on a new permission - that's more complexity than we need. I'd say just go with `EXTENDED_READ` in that case.

abayer:
Wait, no, `EXTENDED_READ` doesn't fit either, since that only applies to an `Item`. Woof. This is complicated.

abayer:
@daniel-beck made a good point to me - we already have `Overall/Read` for `/pipeline-syntax`, so we're not actually exposing anything additional here if we use the same permissions. So, it's gonna be that.

stephenc:
If we do not have the job context then we cannot validate things like credential ids

Perhaps let the command take an optional context?

rsandell:
:bee: should possibly also include a test under secured jenkins to check the permission holds.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommandTest.java
```diff
@@ -75,6 +79,18 @@ public void invalidJenkinsfile() throws Exception {
         assertThat(result.stdout(), containsString("Not a valid section definition: 'agent'. Some extra configuration is required"));
     }
 
+    @Test
+    public void invalidUser() throws Exception {
+        File testPath = writeJenkinsfileToTmpFile("simplePipeline");
+
+        j.jenkins.setAuthorizationStrategy(new GlobalMatrixAuthorizationStrategy());
```
rsandell: There is no need to add matrix-auth as a dependency, when you have access to MockAuthorisationStrategy
Example:
```java
j.jenkins.setSecurityRealm(j.createDummySecurityRealm());
j.jenkins.setAuthorizationStrategy(new MockAuthorizationStrategy()
                                                   .grant(Jenkins.ADMINISTER).everywhere().to("bob")
                                                   .grant(Jenkins.READ, 
                                                             Item.READ, 
                                                             Item.EXTENDED_READ).everywhere().to("alice"));
```

abayer: Okiedokie, fixing.

rsandell:
:bee:

jimklimov:
Just in case for posterity, (and maybe an issue with my system or `curl` version), but the `curl ... -F 'jenkinsfile="@Jenkinsfile"' ...` syntax did not work for me; instead, I had to use 
````
--form "jenkinsfile=`cat Jenkinsfile`"
````
to get the job done.

Overall my command was like this (for the older similar service - but this hint is rather about curl, not the fix):
````
BASEURL='http://username:passw0rd@appserver:8080/jenkins'

curl -H "`curl "$BASEURL/crumbIssuer/api/xml?xpath=concat(//crumbRequestField,%22:%22,//crumb)"`" \
    -X POST --form "jenkinsfile=`cat Jenkinsfile`" \
    "$BASEURL/pipeline-model-converter/validateJenkinsfile"
````

rsandell:
@jimklimov From what I can see you are calling the http api and not the CLI command that this PR is about.

jimklimov:
Yes, this was just a hint for people googling into this issue due to http-api use-case (as it is posted in the initial comment) which they need (or think so).

ssbarnea:
Somehow I find the comments from the first comment quite incomplete, I was hoping to see script that works regardless if Jenkins has crumble protection enabled or not.

PR #59, opened 2016-11-29
[JENKINS-39134] Fix CreationException when resuming inside script {}

jglick:
[JENKINS-39134](https://issues.jenkins-ci.org/browse/JENKINS-39134)

My initial thought was that a missing `optional = true` attribute on `@Inject` would solve the issue; but it did not, and that would not explain the `withMaven` problem anyway. At any rate, this fix works for this plugin, and I will track down the `pipeline-maven` bug separately.

@reviewbybees


abayer:
Interesting. :bee:

rsandell:
:bee:

PR #61, opened 2016-12-01
[FIXED JENKINS-40136] Properly support failFast in parallel.

abayer:
[JENKINS-40136](https://issues.jenkins-ci.org/browse/JENKINS-40136)

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
@@ -21,6 +21,7 @@ public final class ModelASTStage extends ModelASTElement {
     private ModelASTWhen when;
     private ModelASTTools tools;
     private ModelASTEnvironment environment;
+    private Boolean failFast;
 
     public ModelASTStage(Object sourceLocation) {
         super(sourceLocation);
@@ -35,6 +36,9 @@ public final class ModelASTStage extends ModelASTElement {
         JSONObject o = new JSONObject();
         o.accumulate("name", name);
         o.accumulate("branches", a);
+        if (failFast != null) {
+            o.accumulate("failFast", failFast);
+        }
         if (agent != null) {
             o.accumulate("agent", agent.toJSON());
         }
@@ -109,6 +113,9 @@ public final class ModelASTStage extends ModelASTElement {
                 }
                 result.append('\n');
                 result.append(branch.getName()).append(": {\n").append(branch.toGroovy()).append("\n}");
+            }
+            if (failFast != null && failFast) {
+                result.append(",\nfailFast: true");
             }
             result.append("\n)\n");
         } else if (branches.size() == 1) {
@@ -205,6 +212,14 @@ public final class ModelASTStage extends ModelASTElement {
         this.environment = environment;
     }
 
+    public Boolean getFailFast() {
+        return failFast;
+    }
+
+    public void setFailFast(Boolean f) {
+        this.failFast = f;
+    }
+
     @Override
     public String toString() {
         return "ModelASTStage{" +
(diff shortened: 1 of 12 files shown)


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
```diff
@@ -21,6 +21,7 @@
     private ModelASTWhen when;
     private ModelASTTools tools;
     private ModelASTEnvironment environment;
+    private Boolean failFast;
```
rsandell: So we should not provide the ability to say `failFast = env.BRANCH != "master"` ? >:)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -306,7 +306,25 @@ class ModelParser {
                             break
                         case 'steps':
                             def stepsBlock = matchBlockStatement(s);
-                            stage.branches.addAll(parseStepsBlock(asBlock(stepsBlock.body.code)))
+                            BlockStatement block = asBlock(stepsBlock.body.code)
```
rsandell: Why are you inlining this block? didn't we have the code creep issue refactored once already?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
abayer: =P

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: Because I needed to have `failFast` available at this point, not just the branches.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
rsandell: I can bet a beer on that someone will try it and then complain that it doesn't work :)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
rsandell: But why can't you extract this to a separate method?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
abayer: Hrm.

abayer: Yeah, thought about it - I say keep it simple for now.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: I prefer to avoid multi-return methods, which is what I'd ned to get both the branches and `failFast` returned from a method. This code's literally used nowhere but here, so inline is fine, IMO.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
rsandell: Agreed

rsandell:
reluctant :bee:

abayer:
=)

PR #62, opened 2016-12-01
Make internal-only steps advanced.

abayer:
This is so they don't show up in the Snippet Generator or the eventual
Pipeline Editor step choice UI, since they only exist for internal purposes.

@kzantow - this covers one of the things we talked about today.


rsandell:
:bee: Well they will still show up in the snippet generator, but at the bottom under the `advanced/deprecated` section.

PR #63, opened 2016-12-08
Switch to requiring no Launcher in getRequiredContext() for wrappers

abayer:
More narrowly scoped than the previous requiring it to be empty. That
ended up bugging out with TimeoutStep from workflow-basic-steps 2.3
(probably due to some other change elsewhere in the dependency tree,
since TimeoutStep itself didn't change between 2.2 and 2.3, though
TimeoutStepExecution did). So switching!


rsandell:
:bee:

PR #64, opened 2016-12-08
[FIXED JENKINS-40226] Make sure non-within-stage failures are real.

abayer:
[JENKINS-40226](https://issues.jenkins-ci.org/browse/JENKINS-40226)

More notably, make sure we still end up doing a `post` for them if
needed and that the build status is correct at that time.

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
@@ -25,6 +25,7 @@
 
 package org.jenkinsci.plugins.pipeline.modeldefinition.agent.impl
 
+import hudson.model.Result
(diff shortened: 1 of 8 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
```diff
@@ -48,14 +49,30 @@ public class DockerPipelineFromDockerfileScript extends DeclarativeAgentScript {
             if (!Utils.withinAStage()) {
                 script.stage(SyntheticStageNames.agentSetup()) {
                     Utils.markSyntheticStage(SyntheticStageNames.agentSetup(), Utils.getSyntheticStageMetadata().pre)
-                    buildImage().call()
+                    try {
+                        buildImage().call()
+                    } catch (Exception e) {
+                        script.getProperty("currentBuild").result = Result.FAILURE
```
rsandell: Shouldn't there be some information printed to the build log explaining that building the image failed?

rsandell:
Slight concern about not giving the user enough information, but :bee:

PR #65, opened 2016-12-12
Fixed some test failures

rsandell:
For some reason the dockerfile and malformed verification tests
didn't succeed on my new dev machine when I made a fresh clone
of the workspace. While fixing them I wonder how they ever could have worked.

@reviewbybees esp. @abayer


abayer:
:bee:

PR #67, opened 2016-12-13
JENKINS-40337 - add top-level `options` section and `skipCheckout` option

abayer:
[JENKINS-40337](https://issues.jenkins-ci.org/browse/JENKINS-40337)

Right now, there's just the one possible thing that can go in `options`, but I have a feeling there could well be more. I opted to not go with a `Map` underlying the `options` (unlike `tools`, `environment`, et al) because this way, it'll be a *lot* easier in the future if we need to add a complex option - i.e., a nested map, etc - while still getting good type and name validation. That said, if there's a better way to do it, I'm not saying no!

cc @reviewbybees esp @rsandell @michaelneale @HRMPW @i386

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTOptions.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTOptions.java
@@ -1,0 +1,104 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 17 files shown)


i386:
@abayer how are you explaining the difference between `options` and `properties`? when is something supposed to be a `option` rather than a `property`?

abayer:
Properties mean job properties. Don't blame me for the fact that that's a thing. =) We could change options to config?

i386:
Job properties are defined as:
> Other job properties, such as build discarding, limiting concurrent builds, and more.

That sounds a lot like `options` to me. Perhaps `properties` are a mistake and we should be moving these into well documented `options`?

abayer:
No. `properties` aka `jobProperties` are all extensions of the `JobProperty` extension point in core. They are relevant outside of Declarative. `options` would only contain things specific to Declarative alone.

michaelneale:
A user won't see the difference between properties and options. A UI (which PMD is) shouldn't be a lecture in the legacy decisions of design of Jenkins if at all possible. 

options OR properties make sense.

i386:
@abayer that distinction doesn't make a lot of sense from the users perspective since it requires a history lesson in the implementation of Jenkins and Pipeline.

What `JobProperties` are there that are most likely going to apply to multi-branch pipeline? How commonly will they be used? We should categorise and proprieties these and turn them into options and remove `properties`.

rsandell:
Do we need an options block at all couldn't it be properties on the pipeline block itself?
And the syntax is different from the other "variable like" assignments. In `environment` for example the syntax is `name = value` while in options you've gone for `name "value"` how are users going to know when to use what?

Ex:

```
pipeline {
  agent none
  skipCheckout = true

  environment {
    FOO = "bar"
  }

  stages {
    ...
  }
}

rsandell:
When I'm writing my PMD pipelines by hand I always do the mistake of writing `agent: any` and have to go back and change it :)

abayer:
the `foo = bar` in `environment` wasn't my first choice. =) I wanted to do `foo bar` everywhere, and that's what we're using in every case other than `environment`.

abayer:
Closed in favor of #70

PR #68, opened 2016-12-13
[FIXED JENKINS-40418] Fix validation for a number of extended types

abayer:
[JENKINS-40418](https://issues.jenkins-ci.org/browse/JENKINS-40418)

Turns out we do need validate methods on child classes - the parent
validate method will end up calling
validator.validateElement(ParentClass) which is...not what we
want. So, fixing that in a bunch of places and adding tests.

Also, added an expectError convenience method.


pipeline-model-api/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorTest.java
```diff
@@ -44,4 +48,18 @@ public void ensure_postStage_calls_own_validator_and_super() {
...
+    public void triggerValidateCallsRightMethod() throws Exception {
+        ModelValidator validator = Mockito.mock(ModelValidator.class);
+
+        ModelASTTrigger trigger = new ModelASTTrigger(null);
+        ModelASTMethodArg arg = ModelASTValue.fromConstant("bar", null);
+        trigger.setArgs(Collections.singletonList(arg));
+        trigger.setName("foo");
+        trigger.validate(validator);
+        InOrder inOrder = Mockito.inOrder(validator);
+        inOrder.verify(validator, Mockito.times(1)).validateElement(trigger);
```
rsandell: I think you should go with `.validateElement(same(trigger));` just to be sure.

rsandell:
:bee:

PR #70, opened 2016-12-14
[JENKINS-40337] Add Declarative-specific options.

abayer:
[JENKINS-40337](https://issues.jenkins-ci.org/browse/JENKINS-40337)

Specifically adding skipCheckout, but also renamed "properties"
section to "options" and put both JobPropertys and DeclarativeOptions
in that section. Yes, we just renamed "jobProperties" to "properties",
but I prefer "options" when we're not just straight-up containing
JobPropertys. So. Yeah.

cc @reviewbybees esp @rsandell @michaelneale @i386 @HRMPW

PR #72, opened 2016-12-17
[FIXED JENKINS-40462] Move wrappers into options

abayer:
[JENKINS-40462](https://issues.jenkins-ci.org/browse/JENKINS-40462)

This depends on
https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/70
so shouldn't be merged until that's in.

I'm not a huge fan of the hoops we have to jump through to support
wrappers in the same section as Describables, so I may tweak this.


rsandell:
I'm holding off my review until #70 is merged.

abayer:
@rsandell Mind reviewing?

SYNTAX.md
```diff
@@ -269,8 +269,8 @@ parameters {
 ```
 
 ### Options
-* *Description*: Other options exclusive to Declarative, such as `skipDefaultCheckout`, and traditional `JobProperty`s, such as 
-build discarding, limiting concurrent builds, and more.
+* *Description*: Other options exclusive to Declarative, such as `skipDefaultCheckout`, as well as traditional `JobProperty`s, such as 
+build discarding, limiting concurrent builds, and more, and "wrappers" that should wrap the entire build, such as `timeout`.
```
rsandell: I think I got an aneurysm reading this sentence.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Options.groovy
```diff
@@ -72,31 +83,52 @@ public class Options implements Serializable {
...
 
+    public static Map<String,String> getEligibleWrapperStepClasses() {
+        Map<String,String> stepNames = [:]
+
+        // TODO: Figure out if we want to support metasteps? Don't think so.
+        ExtensionList.lookup(StepDescriptor.class).each { s ->
+            if (s.takesImplicitBlockArgument()
+                && !(s.getFunctionName() in ModelASTMethodCall.blockedSteps.keySet())
+                && !(Launcher.class in s.getRequiredContext())
+                && !(FilePath.class in s.getRequiredContext())) {
```
rsandell: I'm not sure about the cost of calling `s.getRequiredContext()` but if it does some kind of lookup you should consider cashing it in a local variable first.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/JSONParser.groovy
```diff
@@ -113,9 +113,6 @@ class JSONParser implements Parser {
                 case 'parameters':
                     pipelineDef.parameters = parseBuildParameters(pipelineJson.getJSONObject("parameters"))
                     break
-                case 'wrappers':
```
rsandell: Deprecation warning?

SYNTAX.md
abayer: Rephrased. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/JSONParser.groovy
abayer: Nah - we don't really care about that for the JSON at this point. Plus, `wrappers` would actually fail at schema validation time, before we even got this far.

abayer:
ping @rsandell

abayer:
Gonna merge this - if it turns out there are problems with the new caching approach, we can fix them later.

PR #73, opened 2016-12-17
[FIXED JENKINS-40524] Reworked agent syntax to be more extensible.

abayer:
[JENKINS-40524](https://issues.jenkins-ci.org/browse/JENKINS-40524)

In order to get a better syntax for `when`, we're also going to spruce up the syntax for `agent` for consistency's sake, killing off the `agent label: 'foo'` in favor of 
```
agent {
  label 'foo'
}
```

This still supports the bare `agent none` and `agent any`, though. So it's a bit weird.

cc @reviewbybees esp @rsandell @i386 @michaelneale @HRMPW


HRMPW:
:bee: for the syntax.

rsandell:
It seems awkward to have each argument on a separate line.

kuhnroyal:
Regarding https://issues.jenkins-ci.org/browse/JENKINS-39684 maybe something more generic is required. Something to the likes of this.

```
agent docker {
     label 'has-docker'
     image 'ubuntu:lts'
     registry 'docker.example.com'
     credentialsId 'foo'
     args '-v /var/run/docker.socks:/var/run/docker.socks -p 8080:8080'
}
```

```
agent dockerFile {
     label 'has-docker'
     name 'Dockerfile'
     registry 'docker.example.com'
     credentialsId 'foo'
     args '-v /var/run/docker.socks:/var/run/docker.socks -p 8080:8080'
}
```

```
agent jenkins? {
     label 'foo'
}
```

Not sure if that is possible.

abayer:
@rsandell Yeah, but I wanted consistency with the syntax we're using for every other section (with the weirdo exception of `environment`, but that is still pretty similar).

@kuhnroyal The syntax in the PR can actually support the more complicated/generic key combos already.

kuhnroyal:
Nice!
I was just wondering what your thoughts are on:
```
agent docker {
     label 'has-docker'
     image 'ubuntu:lts'
     args '-v /var/run/docker.socks:/var/run/docker.socks -p 8080:8080'
}
```
vs
```
agent docker {
     label 'has-docker'
     docker 'ubuntu:lts'
     dockerArgs '-v /var/run/docker.socks:/var/run/docker.socks -p 8080:8080'
}
```

I think the first one is easier to understand.

kuhnroyal:
@abayer Sounds good, whatever works and is easier to implement.

michaelneale:
I like what I am seeing

i386:
Quite like that `agent docker { label 'foo' }` proposal.

HRMPW:
That will really mess up consistency. We tried that before and needed the parens.  Also, `agent docker {label 'foo'} ` would require `image 'my-image` in the closure anyway so you would be adding verbiage not reducing it.

i386:
@HRMPW ack

michaelneale:
Will this also support registries in a nicer way: https://issues.jenkins-ci.org/browse/JENKINS-39684 (or at least allow that, in a future PR?)

abayer:
This will enable JENKINS-39684, yes.

abayer:
FYI, I briefly pushed a change that started adding support for nested maps within `agent`, i.e., to support something like:
```
agent {
  label "foo"
  kubernetes {
    someK8Conf "bar"
    someOtherConf "baz"
  }
}
```
...but that ended up being a lot harder than I wanted it to be, so I yanked it for now. Given that adding that in the future doesn't mean breaking compatibility, I'm declaring this PR ready to go.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
```diff
@@ -1,65 +1,107 @@
...
-        this.args = args;
+    public void setVariables(Map<ModelASTKey, ModelASTValue> variables) {
+        this.variables = variables;
     }
 
     @Override
     public String toString() {
         return "ModelASTAgent{" +
-                "args=" + args +
+                "variables=" + variables +
```
rsandell: Don't think this would produce a very good output.

abayer: How so?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
```diff
@@ -67,13 +67,9 @@ public class Root implements NestedModel, Serializable {
...
-    Root agent(Map<String,String> args) {
-        this.agent = new Agent(args)
-        return this
-    }
-
-    Root agent(String s) {
-        this.agent = new Agent(s)
+    Root agent(Agent a) {
+        this.agent = a.convertZeroArgs()
+        System.err.println("arg: ${this.agent}")
```
rsandell: remove debug prints

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
```diff
@@ -67,13 +67,9 @@ public class Root implements NestedModel, Serializable {
...
     }
 
-    Root agent(Map<String,String> args) {
-        this.agent = new Agent(args)
-        return this
-    }
-
-    Root agent(String s) {
-        this.agent = new Agent(s)
+    Root agent(Agent a) {
```
rsandell: Can `a` be `null`?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
abayer: oops.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
abayer: Yup.

abayer: Wait, no, it can't be null - we require the `agent` section at the top level.

abayer: But it could be in `Stage`, so I just made that null-safe. =)

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
rsandell: It seems like it would produce something like `variables =LinkedHashMap@12345`

SYNTAX.md
```diff
@@ -30,13 +30,27 @@ These are sections that are specified directly within the `pipeline` argument cl
...
 * *Examples*:
-    * `agent label:'has-docker', docker:'ubuntu:lts'`
-    * `agent docker:'ubuntu:lts'`
-    * `agent label:'hi-speed'`
-    * `agent none`
-    * `agent any`
+
+```groovy
+agent {
+    label 'has-docker'
```
kzantow: Is the only reason for this to have a way to run on an agent with access to docker?

SYNTAX.md
```diff
@@ -30,13 +30,27 @@ These are sections that are specified directly within the `pipeline` argument cl
...
+    docker 'ubuntu:lts'
+}
+
+agent {
+    docker 'ubuntu:lts'
+}
+ 
+agent {
```
kzantow: I prefer the original proposal, also... this would make sense to be: `agent any { label: 'hi-speed' }`

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
abayer: we've got the same thing in `ModelASTEnvironment` and elsewhere - I'll poke at it.

SYNTAX.md
abayer: Or to otherwise specify the agent, yes.

SYNTAX.md
abayer: `agent any` is syntactic sugar for `agent { label '' }` - i.e., any agent.

kzantow: I see, now I understand what you were referring to before. From my perspective, `none` and `any` could be some agent model that just accepts no and an optional closure, respectively. If I wanted to run on any agent with the label `hi-speed`, I would use `agent any { label: 'hi-speed' }` but if I wanted to run on a docker container, I would use `agent docker { image: 'maven' }` the `docker` agent type would accept a label, also, to filter to agents with specific labels. This would seem to not require any sort of special handling for `none` and `any` and would allow things like the `docker` agent to automatically filter without a label to only docker-enabled agents. Am I missing something? This would be analogous to steps: select the step type, fill out the step parameters (although admittedly that's somewhat of a UI thing)...

abayer: So take a look at the discussion on the JIRA as well - but a couple main points.

One - `agent any { label "foo" }` isn't something I'm going to go with in the first place. I *really* don't like the look of `something somethingelse { closure }`. And...well, yeah, everything else is in the JIRA. I don't *love* this approach, but it works without a drastic rewrite of the backend implementation.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -172,8 +172,7 @@ public class ModelInterpreter implements Serializable {
             } catch (MissingContextVariableException e) {
                 if (FilePath.class.equals(e.type) || Launcher.class.equals(e.type)) {
                     if (!agent.hasAgent()) {
-                        script.error("Attempted to execute a step that requires a node context while 'agent none' was specified. " +
-                                "Be sure to specify your own 'node { ... }' blocks when using 'agent none'.")
+                        script.error(Messages.ModelInterpreter_NoNodeContext())
```
rsandell: Not really necessary since this won't execute on a user request thread you'd most likely get the default language anyway.

rsandell:
:bee:

abayer:
Closed in favor of #85

PR #74, opened 2016-12-19
[FIXED JENKINS-40547] Verify durability of Declarative runs

abayer:
[JENKINS-40547](https://issues.jenkins-ci.org/browse/JENKINS-40547)

Was 99% sure this would work, and, well, that was correct.

cc @reviewbybees esp @rsandell and @jglick - any ideas for additional things we should be testing re durability?


abayer:
@rsandell Generally agreed -not worried about {{script}} since that's braindead simple, but otherwise, yes. I'll add more tests to this PR.

PR #75, opened 2016-12-19
[JENKINS-40188] Test to verify that boolean params can be passed on.

abayer:
[JENKINS-40188](https://issues.jenkins-ci.org/browse/JENKINS-40188)

Can't actually reproduce the issue with 'env.PARAM' or 'params.PARAM',
but can with '"${env.PARAM}"'. If there's a deeper problem here, it's
that we're not actually validating the parameters to the parameters of
steps themselves, but that's a different matter.


herau:
+1

rsandell:
:bee:

PR #78, opened 2016-12-20
[JENKINS-40370] Make stage.when more declarative

rsandell:
[JENKINS-40370](https://issues.jenkins-ci.org/browse/JENKINS-40370)

cc @reviewbybees esp @kzantow @i386 @HRMPW @michaelneale


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStep.java
```diff
@@ -15,7 +15,7 @@
  */
 @ToString(includeSuper = true, includeSuperProperties = true)
 @SuppressFBWarnings(value = "SE_NO_SERIALVERSIONID")
-public class ModelASTStep extends ModelASTElement {
+public class ModelASTStep extends ModelASTElement implements ModelASTWhenCondition {
```
rsandell: This does not look correct from a model perspective.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTWhenCondition.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.jenkinsci.plugins.pipeline.modeldefinition.ast;
+
+import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator;
+
+/**
+ * A marker for special-casing validation of a {@link ModelASTStep} inside {@link ModelASTWhen}
+ * @author Andrew Bayer
+ */
+public interface ModelASTWhenCondition {
```
rsandell: This looks more like a base AST element to me.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ExpressionConditional.java
```diff
@@ -0,0 +1,74 @@
...
+    @DataBoundConstructor
+    public ExpressionConditional(StepsBlock block) {
+        this.block = block;
+    }
+
+    public StepsBlock getBlock() {
+        return block;
+    }
+
+    public static boolean booleanFromReturn(Object o) throws AbortException {
```
rsandell: Well if you insists of calling groovy scripts then this should be removed and use the groovy evaluation to determine if something is true or not. Then we can get the benefits back for for example `if "not empty"` etc.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -404,6 +405,23 @@ public class ModelInterpreter implements Serializable {
...
     /**
+     *
+     */
+    def evaluateWhen(StageConditionals when) {
+        if (when == null) {
+            return true
+        } else {
+            for (int i = 0; i < when.conditions.size(); i++) {
+                DeclarativeStageConditional c = when.conditions.get(i)
+                if (!c.getScript(script).evaluate()) {
```
rsandell: this seems rather excessive to me, the only benefit I can see is my comment above on the boolean expression.

rsandell:
The changes to agent are throwing me off, I guess they are realted to the WithScript thingie, but at the same time should perhaps be done separately to not do too much in the same pr.

abayer:
@rsandell I thought about that, but I needed to make sure that `WithScript` would work for both `DeclarativeAgent` and `DeclarativeStageConditional` so it made sense to do it all here.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -351,4 +352,29 @@ public class Utils {
...
+     */
+    public static boolean isOfType(UninstantiatedDescribable ud, Class<?> base) {
+        Descriptor d = SymbolLookup.get().findDescriptor(base, ud.symbol)
+        return d != null
+    }
+
+    /**
+     * @param c The closure to wrap.
+     */
+    public static StepsBlock createStepsBlock(c) {
```
abayer: A `Closure` - I feel like there was a reason I didn't specify the type originally, but I'll try it again with it typed.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ExpressionConditional.java
abayer: Not sure what you mean?

abayer:
@rsandell Got rid of `ModelASTWhenCondition` and switched to a different name for the "validate `ModelASTStep` from within a `ModelASTWhen`" validation method.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ExpressionConditional.java
abayer: Ah, I kinda know what you mean now - I want to avoid having to `evaluate` the expression, since we need to run it in the `CpsScript` context etc, and that's...a pain to pull off via `evaluate`. Hence the move to `WithScript`. But then we need to make sure that the return of the closure is actually a boolean.

rsandell: Well I don't think we need to make sure that it actually is a boolean, just that Groovy can evaluate it to `true`/`false` and would be much more convenient if we let groovy handle that then trying to emulate it in Java imho it is much more convenient to just do `{ return env.MY_ENV }` instead of `{return env.My_ENV != null && !env.MY_ENV,isEmpty()}`

rsandell:
I would like to see a more pure Groovy eval of the scripted when condition, but intermediate :bee: anyway.

abayer:
@rsandell We can always change the backend implementation later - right now we just need the syntax locked in and things working. =)

SYNTAX.md
```diff
@@ -144,6 +144,33 @@ stages {
...
+* *Takes a Closure*: Yes
+* *Closure Contents*: One or more of the possible `when` conditions, provided by an extension point.
+    * *Built-In `when` Conditions*:
+        * `branch` - run when the branch being built matches the branch pattern given.
+        * `environment` - run when the environment contains a given environment variable with a given value.
+        * `expression` - run when an arbitrary Groovy expression returning a boolean is true.
+* *Examples*:
+```groovy
+when {
+    branch "*/master"
```
kzantow: this looks excellent

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTWhen.java
```diff
@@ -25,29 +25,73 @@
...
-        return o;
+        result.append("}\n");
+        return result.toString();
     }
 
     @Override
-    public void validate(ModelValidator validator) {
-        super.validate(validator);
+    public void removeSourceLocation() {
+        super.removeSourceLocation();
```
kzantow: what is this for?

abayer: To get the source location information (i.e., the `ASTNode` or `JSONObject`) out so that we can serialize as a graph - e.g., https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy#L53 calling https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy#L192-L201

rsandell:
@abayer well in this case fixing the backend means removing a unrestricted public static method, hence me wanting to iron it out before the merge.

abayer:
How would you like it to work? I'm honestly not entirely clear on what you want. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ExpressionConditionalScript.groovy
```diff
@@ -0,0 +1,51 @@
...
+        Object retVal = null
+        Closure c = describable.block?.closure
+
+        if (c != null) {
+            c.delegate = script
+            c.resolveStrategy = Closure.DELEGATE_FIRST
+            retVal = c.call()
+        }
+
+        return ExpressionConditional.booleanFromReturn(retVal)
```
rsandell: What I'm saying is change this line to

```groovy
    if (retVal) {
        return true
    } else {
        return false
    }
```

Then we would say `true` if what was returned was for example a non empty string or a int of value `0` etc. So it behaves like it did in previous version(s).

abayer: aaaaaaaah. That makes sense. Doing so.

rsandell:
:bee: Since I own this PR, I can't approve it :)

PR #79, opened 2016-12-20
[FIXED JENKINS-40580] Quote parallel branch names

abayer:
[JENKINS-40580](https://issues.jenkins-ci.org/browse/JENKINS-40580)

Decided to be defensive and always quote when converting, rather than
trying to figure out when we could avoid quoting.

cc @reviewbybees esp @kzantow @rsandell

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+import org.apache.commons.lang.StringEscapeUtils;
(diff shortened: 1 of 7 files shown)


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
```diff
@@ -112,7 +112,7 @@ public String toGroovy() {
                     result.append(',');
                 }
                 result.append('\n');
-                result.append(branch.getName()).append(": {\n").append(branch.toGroovy()).append("\n}");
+                result.append('"' + branch.getName() + '"').append(": {\n").append(branch.toGroovy()).append("\n}");
```
kzantow: 🐛  defensively, I would also here (and any other places we're quoting things) replace quotes. e.g. `.replaceAll("\"", "\\\"")`; probably there's a utility function for this somewhere.

rsandell: Yea, and add a test where the name has a `"` in it, for fun and profit.

abayer:
And updated.

PR #80, opened 2016-12-22
[FIXED JENKINS-40642] Add additional imports for parse/translate

abayer:
[JENKINS-40642](https://issues.jenkins-ci.org/browse/JENKINS-40642)

@NonCPS and @Library were the most glaring missing import, but I
replicated the standard CpsGroovyShell imports - @NonCPS's package,
hudson.model, jenkins.model, and anything provided by a GroovyShellDecorator.

cc @reviewbybees esp @rsandell @kohsuke and @imod

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
@@ -43,7 +43,6 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStep
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution
 import org.jenkinsci.plugins.workflow.cps.CpsScript
 import org.jenkinsci.plugins.workflow.cps.GroovyShellDecorator
-import org.jenkinsci.plugins.workflow.libs.Library
 
 import java.security.CodeSource
 import java.security.cert.Certificate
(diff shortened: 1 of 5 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
```diff
@@ -108,14 +112,28 @@ public class Converter {
...
 
+    private static CompilerConfiguration makeCompilerConfiguration() {
+        CompilerConfiguration cc = new CompilerConfiguration();
+
+        ImportCustomizer ic = new ImportCustomizer();
+        ic.addStarImports(NonCPS.class.getPackage().getName());
+        ic.addStarImports("hudson.model","jenkins.model");
+        for (GroovyShellDecorator d : GroovyShellDecorator.all()) {
+            d.customizeImports(null, ic);
+        }
```
imod: I sure must be missing something, but where does `org.jenkinsci.plugins.workflow.libs.Library` get added to the imports?

abayer: Through the magic of `GroovyShellDecorator`s - https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryDecorator.java =)

imod:
👍

rsandell:
non needed :bee: but making the bot happy.

PR #82, opened 2016-12-31
[FIXED JENKINS-40721] Don't require parens for no-arg method calls

abayer:
[JENKINS-40721](https://issues.jenkins-ci.org/browse/JENKINS-40721)

(note that this is on top of #72)

This is...weird. I'm a little shocked that this works, honestly. But!
You can now do something like:

```
options {
  timestamps
}
```

And it'll work. Note that this is only applicable in the `options`,
`triggers`, and `parameters` sections - no-arg steps still need their
parentheses even with this change.

I'm still not 100% sure this is a good idea, but...dang. It works.

cc @reviewbybees esp @rsandell @michaelneale @i386 @HRMPW

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -41,6 +41,9 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTPipelineDef
 import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStages
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.MethodsToList
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.StepsBlock
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.Options
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.Parameters
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.Triggers
 import org.jenkinsci.plugins.pipeline.modeldefinition.parser.Converter
 import org.jenkinsci.plugins.structs.SymbolLookup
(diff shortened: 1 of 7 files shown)


pipeline-model-definition/src/test/resources/json/errors/invalidWrapperType.json
```diff
@@ -16,7 +16,7 @@
     "isLiteral": true,
     "value": "none"
   },
-  "wrappers": {"wrappers": [  {
+  "options": {"options": [  {
```
michaelneale: How come this is an object ? (I know, same as before)

abayer: Clarity/legacy/ease of translation...this makes it a lot simpler for parsing and translating to/from groovy, IMO.

rsandell:
Looks like this needs a rebase since it still contains the aneurysm inducing sentence.

rsandell:
So is this going in at some point?

abayer:
@rsandell I'm still debating, honestly. =)

michaelneale:
is a nice touch, if not too scary.

bitwiseman:
Okay, I'll play devils advocate.  

Unless you plan to make this change work everywhere in Declarative, definitely don't do it. 

It is documentation nightmare and a support issue and bug farm.  The last thing we need is more special cases: you don't need empty parens here, but you do need them there, oh, but not that other place. 

Groovy overall makes parens optional but not for empty methods. I see zero benefit to innovating away from that.  Let's try to stick with either "It works just like it does in plain old Groovy or it is not allowed".

This is something that should at best be post-1.0 - it is a non-breaking change to make the empty parens optional later.  On the other hand, if you make them optional now, it is a breaking change if you find some reason why they are needed later.  This feature is already coming in hot enough.

abayer:
Aaaand I've decided not to merge this. It's a nice idea but is too inconsistent with everything else.

michaelneale:
good choice I think and @bitwiseman makes good points. Although I worry that people using PMD will have to think too much about groovy (if so, we have failed)

PR #85, opened 2017-01-06
[FIXED JENKINS-40524] Reworked agent syntax to be more extensible.

abayer:
[JENKINS-40524](https://issues.jenkins-ci.org/browse/JENKINS-40524)

Note - this replaces #73.

- [x] Implementation
- [x] Some additional validation tests for preventing multiple agent types (i.e., not allowing more than one thing at the top level of `agent`) and some other details.
- [x] Update SYNTAX.md

Moves agent to a block (except for any and none) and moves any agent
type with more than one argument into a nested block as well. For example:

```groovy
agent {
    label "foo"
}

agent {
    docker {
       nodeLabel "foo"
       image "some-image"
       args "--hi-there"
    }
 }
 ```

Worth mentioning - this lays the groundwork for nesting further maps of config for things like Docker registry config, Kubernetes config, etc - still got a little bit to figure out there, but nothing that blocks this.

Also renamed arguments for docker and dockerfile - label to nodeLabel,
dockerArgs to args, docker to image. Still thinking about how to
handle dockerfile more elegantly.

This also simplifies the JSON model considerably.

cc @reviewbybees esp @kzantow @rsandell @HRMPW @i386 @michaelneale

### CODE REVIEW

**Comment by kzantow** [comment]
Only question, can I do `agent { docker 'maven' }` to get the `maven` image?

**Comment by abayer** [comment]
Yes. I didn't even mean to support that, but it just kinda happened!

File: SYNTAX.md [hunk 1]
```diff
@@ -30,13 +30,27 @@ These are sections that are specified directly within the `pipeline` argument cl
...
+* *Takes a Closure*: yes
 * *Examples*:
-    * `agent label:'has-docker', docker:'ubuntu:lts'`
-    * `agent docker:'ubuntu:lts'`
-    * `agent label:'hi-speed'`
-    * `agent none`
-    * `agent any`
+
+```groovy
+agent {
```
kzantow: All these examples are no longer pertinent, right?

abayer: Correct, updating task list...

kuhnroyal:
I like it!

HRMPW:
@kzantow @michaelneale @i386 for the editor would it be possible to create the long form by default but still be able to read the short form? 

`agent any` 
is readable
but the longer form is what is created for a new Pipeline.
`agent { label '' } ` 

For new users it is easier to explain how to use label, docker, etc in the long form and as they read more documentation and get better acquainted they can start using the sugar and shortcuts.

kzantow:
@HRMPW the editor just shoves the jenkinsfile through the converter to JSON, so it handles all forms the converter handles, it should, at that point, be converted to the long/consistent form for editing, then shoved back through the converter it could handle optimizing forms to get the resulting jenkinsfile, this is irrespective of the editor, though

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
```diff
@@ -9,57 +10,87 @@
...
     @Override
-    public Object toJSON() {
-        return args.toJSON();
+    public JSONObject toJSON() {
+        final JSONObject j = new JSONObject();
+        j.accumulate("type", agentType.toJSON());
+
+        if (variables != null) {
+            if (variables instanceof ModelASTClosureMap &&
+                    !((ModelASTClosureMap) variables).getVariables().isEmpty()) {
```
rsandell: could `variables.getVariables()` return null?

abayer: Nope - https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/85/files#diff-0b9ec62a152fbd9760c7f76b17ac7380R40

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/JSONParser.groovy
```diff
@@ -455,10 +455,42 @@ class JSONParser implements Parser {
         return tools
     }
 
-    public @CheckForNull ModelASTAgent parseAgent(Object j) {
+    public @CheckForNull ModelASTClosureMap parseClosureMap(JSONArray j) {
+        ModelASTClosureMap map = new ModelASTClosureMap(j)
+
+        j.each { rawEntry ->
```
rsandell: Unnecessary cast, do

```
j.each { JSONObject entry ->
```
instead

abayer: Good point.

rsandell:
I couldn't find any changes to the default docker label configuration, but I'm guessing that it would need some touching. Maybe you've forgotten and the test coverage of that feature isn't good enough, or it actually still works. Please check, :bee: otherwise ;)

abayer:
It still works. =) Nothing actually had to change for it - that logic is entirely in the scripts.

HRMPW:
I thought we agreed to change {{nodeLabel}} to just {{label}} to make it consistent in all cases? @abayer

kzantow:
@HRMPW yes, it was switched back: https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/85/commits/b2614b193c78c673bd9336beaef2b3eda66a3a37

PR #87, opened 2017-01-08
[FIXED JENKINS-40573] Re-do step arguments to always be named in JSON representation.

abayer:
[JENKINS-40573](https://issues.jenkins-ci.org/browse/JENKINS-40573)

While still serializing to Groovy as, e.g., "echo('foo')" when
appropriate (i.e., sole required parameter, just one parameter
specified, no closure involved).


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
```diff
@@ -0,0 +1,139 @@
...
+    private transient Map<String, StepDescriptor> stepMap;
+    private transient Map<String, DescribableModel<? extends Step>> modelMap;
+    private transient Map<String, Descriptor<? extends Describable>> describableMap;
+    private transient Map<String, DescribableModel<? extends Describable>> describableModelMap;
+
+    public static DescriptorLookupCache getPublicCache() {
+        return Jenkins.getInstance().getInjector().getInstance(DescriptorLookupCache.class);
+    }
+
+    @Initializer(after = InitMilestone.EXTENSIONS_AUGMENTED)  // Prevents potential leakage on reload
```
abayer: There seems to be a sporadic hang here - not honestly sure what the right approach is?

abayer: Fixed it. Was kind of stupid.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
```diff
@@ -0,0 +1,141 @@
...
+import org.jenkinsci.plugins.structs.SymbolLookup;
+import org.jenkinsci.plugins.structs.describable.DescribableModel;
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+@Extension
+public class DescriptorLookupCache {
```
abayer: I've now synchronized all the relevant methods - is that what you were looking for?

rsandell: yepp :)

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
```diff
@@ -0,0 +1,141 @@
...
+    private transient Map<String, StepDescriptor> stepMap;
+    private transient Map<String, DescribableModel<? extends Step>> modelMap;
+    private transient Map<String, Descriptor<? extends Describable>> describableMap;
+    private transient Map<String, DescribableModel<? extends Describable>> describableModelMap;
+
+    public static DescriptorLookupCache getPublicCache() {
+        return Jenkins.getInstance().getInjector().getInstance(DescriptorLookupCache.class);
```
rsandell: Just wondering why you are using `Jenkins.getInstance().getInjector()` and not something like `ExtensionList.lookup(...)`?

abayer: 'cos I honestly didn't know which one to use and was guessing? =)

abayer: Then we shall do that!

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
```diff
@@ -46,7 +46,7 @@
     private transient Map<String, DescribableModel<? extends Describable>> describableModelMap;
 
     public static DescriptorLookupCache getPublicCache() {
-        return Jenkins.getInstance().getInjector().getInstance(DescriptorLookupCache.class);
+        return Jenkins.getInstance().getExtensionList(DescriptorLookupCache.class).get(0);
```
rsandell: My suggestion was to use `ExtensionList.lookup(DescriptorLookupCache.class).get(...)` instead.

abayer: Yeaaaaah fixed that. =)

kzantow:
@rsandell you know, `Jenkins.getInstance()` is now `@Nonnull` ;)

rsandell:
@kzantow yes but `ExtensionList.lookup()` can still return an empty list in the same situations as `Jenkins.getInstance()` can throw IllegalStateException ;)

PR #88, opened 2017-01-10
Log feedback for skipped stages

abayer:
It's a bit confusing when you see the stage step in the logs for a
skipped stage but no steps executed. There should be log messages for
the skip cases. So, well, here they are.

cc @reviewbybees esp @batmat @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -33,6 +33,7 @@ import hudson.ExtensionList
 import hudson.model.Describable
 import hudson.model.Descriptor
 import org.apache.commons.codec.digest.DigestUtils
+import org.apache.commons.lang.reflect.FieldUtils
 import org.jenkinsci.plugins.pipeline.StageStatus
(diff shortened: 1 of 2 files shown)


batmat:
:bee: \o/ Merci !

rsandell:
But now you actually have steps in the stage ;)

batmat:
@rsandell for stage view right now, it's better since it will show an empty stage, but maybe this would cause issues to BO? Andrew says BO shows those stages disabled in a clearer way, but how does it identify those? 

cc @michaelneale @i386 ?

abayer:
@rsandell Yeah, that's a good point, actually - @vivek, will this cause problems in Blue Ocean visualization if skipped stages have the one `echo` invocation? I know there were problems when we had an `echo` at the end of a failed stage, but can't remember if that would be an issue here.

vivek:
@abayer Blueocean shows skipped stage as not built, if you have a step inside it, its going to show up thats kind of odd. cc: @i386, @michaelneale

rsandell:
If you could somehow get a hold of the active TaskListener and just do `getLogger().println(...)` I guess it wouldn't show up as a step invocation.

abayer:
@rsandell @vivek Got a `TaskListener` approach up now. =)

vivek:
@abayer thats better, thanks:)

PR #89, opened 2017-01-10
Testing that stages after a skipped stage run right

abayer:
Came up in conversation, decided it was worth having the test here
even though it works. =)

cc @reviewbybees esp @batmat @rsandell

### DIFF
--- a/pipeline-model-definition/src/test/resources/whenLaterStage.groovy
+++ b/pipeline-model-definition/src/test/resources/whenLaterStage.groovy
@@ -1,0 +1,54 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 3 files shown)


pipeline-model-definition/src/test/resources/whenLaterStage.groovy
```diff
@@ -0,0 +1,59 @@
...
+            when {
+                expression {
+                    echo "Should I run?"
+                    return false
+                }
+            }
+            steps {
+                script {
+                    echo "World"
+                    echo "Heal it"
```
rsandell: Funny how a song was once stuck in my head so it leaked into my test data, and now I need to pay it forward months later since it gets stuck in my head again for every code review :)

pipeline-model-definition/src/test/resources/whenLaterStage.groovy
```diff
@@ -0,0 +1,59 @@
...
+                }
+
+            }
+        }
+        stage("Three") {
+            when {
+                expression {
```
rsandell: Should perhaps also have a stage after that doesn't have a when condition at all, just to be even more safe.

abayer: Yeah, I originally had a stage without a `when` but the discussion ended up being about consecutive stages with `when` so I changed it. I'll add an additional stage.

pipeline-model-definition/src/test/resources/whenLaterStage.groovy
abayer: See my "banana" and "monkey" references all over the place. =)

rsandell:
:bee:

PR #90, opened 2017-01-11
[FIXED JENKINS-40984] Always evaluate/run all post conditions

abayer:
[JENKINS-40984](https://issues.jenkins-ci.org/browse/JENKINS-40984)

Even if there's an error in an earlier condition execution, continue
to the subsequent ones. Additionally, switch to doing a one-off check
at the beginning of the post section to see if *any* conditions are
satisfied at that time, and then when actually iterating through the
conditions, check for satisfaction at *that* time. That's so that, for
example, if the build is successful when it gets to evaluating post
conditions and then there's a failure in the execution of `always`,
the `success` block won't be executed, and the `failure` block will,
since the build status has changed.

cc @reviewbybees esp @rsandell @batmat

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -150,7 +150,7 @@ public class ModelInterpreter implements Serializable {
      *
      * @param c The closure to execute
      */
-    def setUpDelegate(Closure c) {
+    def delegateAndExecute(Closure c) {
         c.delegate = script
         c.resolveStrategy = Closure.DELEGATE_FIRST
         c.call()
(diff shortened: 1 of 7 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -425,27 +412,54 @@ public class ModelInterpreter implements Serializable {
...
-                    catchRequiredContextForNode(root.agent) {
-                        for (int i = 0; i < postBuildClosures.size(); i++) {
-                            setUpDelegate(postBuildClosures.get(i))
-                        }
+                String conditionName = orderedConditions.get(i)
+
+                Closure c = responder.closureForSatisfiedCondition(conditionName, script.getProperty("currentBuild"))
+                if (c != null) {
+                    catchRequiredContextForNode(agentContext) {
+                        setUpDelegate(c)
```
rsandell: OK, this looks weird, is `setUpDelegate` actually also executing the closure?

rsandell:
:bee: I think

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Yeah, let me actually rename that.

PR #91, opened 2017-01-12
[FIXED JENKINS-41012] Actually get when branch/env to work

abayer:
[JENKINS-41012](https://issues.jenkins-ci.org/browse/JENKINS-41012)

Stop relying on the step context, since that wasn't actually getting
provided, and instead pass in the needed arguments from the *Script
class. Also make sure we actually have tests for all the conditions
*passing*, not just failing. Oy, but I feel silly there.

Assuming this is good, I'll do an 0.8.2 release first thing Thursday.


pipeline-model-definition/src/test/resources/whenBranchFalse.groovy
```diff
@@ -24,6 +24,9 @@
 
 pipeline {
     agent any
+    environment {
+        BRANCH_NAME = "master"
```
rsandell: Really? Is the environment not populated with the "from scm" settings that we use in the tests?

pipeline-model-definition/src/test/resources/whenEnvTrue.groovy
```diff
@@ -0,0 +1,50 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
```
rsandell: 2017

pipeline-model-definition/src/test/resources/whenBranchFalse.groovy
abayer: It is not. Sigh.

pipeline-model-definition/src/test/resources/whenEnvTrue.groovy
abayer: Oy, copy paste!

pipeline-model-definition/src/test/resources/whenEnvFalse.groovy
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016, CloudBees, Inc.
+ * Copyright (c) 2017, CloudBees, Inc.
```
rsandell: Well... the file only changed name, so the year shouldn't really be updated. But that's just a really _anal nit_ :)

abayer: Well, fixed. =)

abayer: Actually, not this one and OH GOD STOP WITH THE COPYRIGHTS I WILL CRY. =)

rsandell: Sorry, I used to work for an enterprise that made it's living on copyrights and patents, so I've been thoroughly indoctrinated on the matter ;)

pipeline-model-definition/src/test/resources/whenBranchFalse.groovy
rsandell: Then perhaps we should be looking for some other way of figuring out what branch it is. But that could be a future enhancement.

kuhnroyal:
@abayer does this somehow allow to match multiple patterns? `feature/**|PR*` or something like this?

rsandell:
@kuhnroyal that would be #99

kuhnroyal:
@rsandell thanks

PR #92, opened 2017-01-12
Fixing errors using endpoint/CLI for conversion/validation

abayer:
We didn't have @Library in our classloader, so we'd get errors. That
was annoying. So here is a fix.

@kzantow - can you do a local build of this and verify it works right for you now?

cc @reviewbybees esp @kzantow @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
@@ -32,7 +32,7 @@ import com.github.fge.jsonschema.main.JsonSchema
 import com.github.fge.jsonschema.report.ProcessingReport
 import com.github.fge.jsonschema.tree.JsonTree
 import com.github.fge.jsonschema.tree.SimpleJsonTree
-import com.github.fge.jsonschema.util.JsonLoader
+import jenkins.model.Jenkins
 import net.sf.json.JSONObject
 import org.apache.commons.lang.reflect.FieldUtils
 import org.codehaus.groovy.control.CompilationFailedException
@@ -106,10 +106,16 @@ public class Converter {
         CompilationUnit cu = new CompilationUnit(
             makeCompilerConfiguration(),
             new CodeSource(src, new Certificate[0]),
-            new GroovyClassLoader());
+            getCompilationClassLoader());
         cu.addSource(src);
 
         return compilationUnitToPipelineDef(cu)
+    }
+
+    private static GroovyClassLoader getCompilationClassLoader() {
+        Jenkins j = Jenkins.instance
+        ClassLoader cl = j != null ? j.getPluginManager().uberClassLoader : new GroovyClassLoader()
+        return new GroovyClassLoader(cl)
     }
 
     /**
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
```diff
@@ -106,12 +106,18 @@ public class Converter {
...
             new CodeSource(src, new Certificate[0]),
-            new GroovyClassLoader());
+            getCompilationClassLoader());
         cu.addSource(src);
 
         return compilationUnitToPipelineDef(cu)
     }
 
+    private static GroovyClassLoader getCompilationClassLoader() {
+        Jenkins j = Jenkins.instance
```
kzantow: y not the `@Nonnull` `Jenkins.getInstance()` ??

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
```diff
@@ -200,7 +211,7 @@ public class Converter {
 
         String rawScript = execution.script
 
-        return scriptToPipelineDef(rawScript)
+        return scriptToPipelineDef(rawScript, CpsThread.current()?.getExecution()?.getShell()?.classLoader)
```
kzantow: Why not put this in the get `getCompilationClassLoader` and use that here?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
abayer: Durrr. Good point. `Jenkins.instance` is `Jenkins.getInstance()` (thanks, Groovy magic!), but there's no need for the null check.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
abayer: Smart man.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
```diff
@@ -106,23 +107,33 @@ public class Converter {
...
 
         return compilationUnitToPipelineDef(cu)
     }
 
+    private static GroovyClassLoader getCompilationClassLoader(GroovyClassLoader inputLoader = null) {
+        if (inputLoader != null) {
+            return inputLoader
+        } else {
+            Jenkins j = Jenkins.instance
+            ClassLoader cl = j != null ? j.getPluginManager().uberClassLoader : new GroovyClassLoader()
```
kzantow: `@Nonnull Jenkins.getInstance()` ?

abayer:
@kzantow All of them addressed in latest commit!

rsandell:
Injecting uberClassaLoader like that might override the sandbox.

abayer:
This isn't in the Pipeline code - it's all sandbox-free. =)

abayer:
Plus the only way we ever end up hitting *that* code path is if we're not even being called within a `CpsThread`.

jglick:
Recommend adding some coverage to `acceptance-test-harness`. Would suffice to have one end-to-end test that pulls in a library and a bunch of tricky features and proves everything works.

abayer:
@jglick Yeah, I need to revisit my ATH PR.

PR #86, opened 2017-01-08
[FIXED JENKINS-39684] Add registryUrl and registryCredentialsId

abayer:
[JENKINS-39684](https://issues.jenkins-ci.org/browse/JENKINS-39684)

Added to both docker and dockerfile agent types. Not sure how to test this, honestly.

This is on top of https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/78 due to its rewrite of `DeclarativeAgent` and friends.

cc @reviewbybees esp @rsandell @HRMPW @jglick

PR #93, opened 2017-01-12
[JENKINS-40866] Docker agent in stage can reuse global node

abayer:
[JENKINS-40866](https://issues.jenkins-ci.org/browse/JENKINS-40866)

Very much still work in progress right now. On top of #86 because we're already messing with `DockerPipeline` and `DockerPipelineFromDockerfile` there and I felt silly creating conflicts for something that'll land after that does anyway.

cc @reviewbybees esp @rsandell @HRMPW @michaelneale @i386


HRMPW:
If we wanted a bit of magic we could have something like:

`agent {
    docker {
      image 'ubuntu'
      label global
    }
}`

If we got to the point of nested we could have: 

`agent {
    docker {
      image 'ubuntu'
      label parent
    }
}`

abayer:
I would strongly prefer no magic - bare words like `any` and `none` are a pain to deal with in the interpreter - [this is the only reason they work](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy#L52-L64) and we'd have to do the same thing for `global` and `parent` etc.

abayer:
Plus that wouldn't be entirely accurate - yes, we'd be on the same `label`, but really, we're on the same `node` - a finer level of accuracy than just `label`. If I say `label parent`, then I'm not going to easily be able to tell whether I'm saying "run on the same node that I'm running on now" or "run on a node with the same label as I'm running on now".

HRMPW:
That's why I updated my comment. The point is to just stay in the current context. Not have to worry about the context of where the statement is used.

abayer:
Ehhh. Again, the biggest problem is the magic bareword, so it doesn't *really* matter what other arguments there are for it. =) But I still don't like the overloading of `label`. fwiw, you can still put `label` in next to `reuseNode` - it just gets ignored when appropriate.

HRMPW:
which gets ignored?

abayer:
`label` gets ignored if `reuseNode` is specified.

michaelneale:
My preference is that if possible, defaulting to reusing node happens, due to principle of least surprise. If this isn't possible when the top level is a docker container, then that is probably ok. 

A common practice (and this has been reflected in feedback) is to run stages in different containers with different tools, but across the one workspace, in a linear fashion.

abayer:
If I can figure out a reliable way of telling when we're already inside `docker.image.inside`, ok. Otherwise...not sure.

HRMPW:
would we need to? can't you fail the syntax if that scenario was attempted?

abayer:
Not easily, no.

abayer:
More detailed answer - with the current architecture of validation, there's no logical place to put validation that spans that far across the AST. It's also a far more specialized check than we do currently - reliant on knowing which agents are valid agents to reuse inside which other ones, for example. It could be done, but it'd be difficult to do generically, and I won't add non-generic things for extension points to the validator unless absolutely necessary.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
```diff
@@ -39,12 +41,39 @@ public class DockerPipelineFromDockerfileScript extends DeclarativeAgentScript<D
 
     @Override
     public Closure run(Closure body) {
-        String targetLabel = describable.label
-        if (targetLabel == null) {
-            targetLabel = script.dockerLabel()?.trim()
+        if (describable.reuseNode && script.getContext(FilePath.class) != null) {
```
rsandell: ... because even if `reuseNode` is always true this would only take effect if we already are on a node.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
```diff
@@ -40,12 +41,38 @@ public class DockerPipelineScript extends DeclarativeAgentScript<DockerPipeline>
 
     @Override
     public Closure run(Closure body) {
-        String targetLabel = describable.label
-        if (targetLabel == null) {
-            targetLabel = script.dockerLabel()?.trim()
+        if (describable.reuseNode && script.getContext(FilePath.class) != null) {
```
rsandell: These changes looks like a copy of the changes in `DockerPipelineFromDockerfileScript` can't we get some sharing going on like you did with the config?
The only difference seems to be in `runImage` so an `AbstractDeclarativeAgentScript` with all the common bits and an `abstract runImage` would be good.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipeline.java
```diff
@@ -38,12 +38,42 @@
     private String label;
     private String image;
     private String args = "";
+    private String registryUrl;
+    private String registryCredentialsId;
+    private boolean reuseNode;
```
abayer: We talked this out - I really don't want to default to `true` since that would be a change from existing behavior and could theoretically break existing pipelines.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
abayer: Yeah, good point.

abayer:
Updated to use a new `AbstractDockerPipelineScript` - let's see how it works!

michaelneale:
nice one!

PR #95, opened 2017-01-17
[FIXED JENKINS-41090, JENKINS-41089] Groovy serialization cleanup

abayer:
[JENKINS-41089](https://issues.jenkins-ci.org/browse/JENKINS-41089)
[JENKINS-41090](https://issues.jenkins-ci.org/browse/JENKINS-41090)

Get rid of superfluous newlines, make sure we have a newline after
'label "foo"', don't serialize unneeded parentheses for single unnamed
arguments for steps/functions.

cc @reviewbybees esp @rsandell @i386

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStep.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStep.java
@@ -100,13 +100,16 @@ public class ModelASTStep extends ModelASTElement {
     private String withOrWithoutParens(ModelASTArgumentList argList) {
         if (argList == null) {
             return name + "()";
-        } else if (!(this instanceof ModelASTTreeStep) &&
-                argList instanceof ModelASTSingleArgument &&
-                // Special-casing for list/map args since they still need parentheses.
-                !argList.toGroovy().startsWith("[")) {
-            return name + " " + argList.toGroovy();
         } else {
-            return name + "(" + argList.toGroovy() + ")";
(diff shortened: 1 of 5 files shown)


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStep.java
```diff
@@ -94,7 +94,20 @@ public String toGroovy() {
...
+        return withOrWithoutParens(argList);
+    }
+
+    private String withOrWithoutParens(ModelASTArgumentList argList) {
+        if (argList == null) {
+            return name + "()";
+        } else if (!(this instanceof ModelASTTreeStep) &&
+                argList instanceof ModelASTSingleArgument &&
+                // Special-casing for list/map args since they still need parentheses.
+                !argList.toGroovy().startsWith("[")) {
```
rsandell: _nit_ could potentially be better optimized by doing just one call to `toGroovy()`

abayer:
@rsandell We already do that - see https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTPipelineDef.java#L121-L156

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStep.java
abayer: Done

PR #94, opened 2017-01-12
[JENKINS-41050] Always do SCM checkout directly on agent, not in container

abayer:
[JENKINS-41050](https://issues.jenkins-ci.org/browse/JENKINS-41050)

cc @reviewbybees esp @michaelneale @rsandell @HRMPW

PR #96, opened 2017-01-20
Get rid of declarativeProps step

abayer:
On top of #94 after refactoring this change out of there in the first place. =)

The declarativeProps...well, why is it even a step? We're basically
just using it internally, we can whitelist it, we can avoid the log
spam of "[Pipeline] declarativeProps" showing up a bunch, we can avoid
cluttering the flow graph... Yeah.


rsandell:
ok, holding off reviewing untill #94 is merged and this is rebased.

abayer:
@rsandell Ready for review now.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeDockerUtils.java
```diff
@@ -0,0 +1,120 @@
...
+        return getLabel(null);
+    }
+
+    @Whitelisted
+    public static String getLabel(@Nullable String override) {
+        if (!StringUtils.isBlank(override)) {
+            return override;
+        } else {
+            for (DockerPropertiesProvider provider : DockerPropertiesProvider.all()) {
+                String label = provider.getLabel(currentRun());
```
rsandell: add `Run current = currentRun()` outside of for loop

abayer: Fixing!

PR #97, opened 2017-01-21
[JENKINS-41243] Slightly faster tagging of synthetic stages

abayer:
[JENKINS-41243](https://issues.jenkins-ci.org/browse/JENKINS-41243)

This approach yielded an average time from just before the stage call
to just before actually attaching the tag of 15.4ms over 5 runs, vs
40.8ms over 5 runs with the previous approach. I'm sure there are many
things wrong with this approach, though.

cc @reviewbybees esp @rsandell @svanoort @vivek @jglick

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -197,9 +197,12 @@ public class Utils {
         ModelASTStages stages = model.stages
 
         stages.removeSourceLocation()
-
-        r.addAction(new SyntheticStageGraphListener.GraphListenerAction())
-        r.addAction(new ExecutionModelAction(stages))
+        if (r.getAction(SyntheticStageGraphListener.GraphListenerAction.class) == null) {
+            r.addAction(new SyntheticStageGraphListener.GraphListenerAction())
+        }
+        if (r.getAction(ExecutionModelAction.class) == null) {
+            r.addAction(new ExecutionModelAction(stages))
+        }
     }
 
     static Predicate<FlowNode> endNodeForStage(final StepStartNode startNode) {
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -33,7 +33,6 @@ import hudson.ExtensionList
 import hudson.model.Describable
 import hudson.model.Descriptor
 import org.apache.commons.codec.digest.DigestUtils
-import org.apache.commons.lang.reflect.FieldUtils
 import org.jenkinsci.plugins.pipeline.StageStatus
 import org.jenkinsci.plugins.pipeline.StageTagsMetadata
 import org.jenkinsci.plugins.pipeline.SyntheticStage
@@ -61,8 +60,6 @@ import javax.annotation.Nullable
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
 import java.util.concurrent.TimeUnit
-import java.util.logging.Level
-import java.util.logging.Logger;
 
 // TODO: Prune like mad once we have step-in-groovy and don't need these static whitelisted wrapper methods.
 /**
@@ -193,7 +190,7 @@ public class Utils {
         }
     }
 
-    static void attachExecutionModel(CpsScript script) {
+    static void attachDeclarativeActions(CpsScript script) {
         WorkflowRun r = script.$build()
         ModelASTPipelineDef model = Converter.parseFromCpsScript(script)
 
@@ -201,13 +198,7 @@ public class Utils {
 
         stages.removeSourceLocation()
 
-        CpsThread thread = CpsThread.current()
-        CpsFlowExecution execution = thread?.execution
-
-        if (execution != null && !execution.complete) {
-            execution.addListener(new SyntheticStageGraphListener())
-        }
-
+        r.addAction(new SyntheticStageGraphListener.GraphListenerAction())
         r.addAction(new ExecutionModelAction(stages))
     }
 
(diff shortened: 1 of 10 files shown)


abayer:
@jglick Well, `StageStep` is final, so I couldn't do an extension of it, and I assumed that it'd be a Bad Thing to reimplement all of `StageStep` with some sugar for this.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -0,0 +1,70 @@
...
+                }
+                if (SyntheticStageNames.postStages().contains(node.getDisplayName())) {
+                    attachTag(node, SyntheticStage.getPost());
+                }
+            }
+        }
+    }
+
+    private void attachTag(FlowNode currentNode, String syntheticContext) {
+        TagsAction tagsAction = currentNode.getAction(TagsAction.class);
```
svanoort: Suggest using getPersistentAction instead (best practice, it's a bit faster and lighter).

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -0,0 +1,70 @@
...
+        if (tagsAction == null) {
+            tagsAction = new TagsAction();
+            tagsAction.addTag(SyntheticStage.TAG_NAME, syntheticContext);
+            currentNode.addAction(tagsAction);
+        } else if (tagsAction.getTagValue(SyntheticStage.TAG_NAME) == null) {
+            tagsAction.addTag(SyntheticStage.TAG_NAME, syntheticContext);
+            try {
+                currentNode.save();
+            } catch (IOException e) {
+                // TODO: should probably do something here? Would prefer to be calling FlowNode.persistSafe() but it's private
```
svanoort: 🐛  You need to handle it somehow -- at least log it.

abayer: Hence the `TODO`. =) I'll just ape `FlowNode.persistSafe()`, I guess.

abayer:
@svanoort Yeah, very much hasn't touched on the resume issue. I wanted to try to make sure it was even worth pursuing before going that far.

abayer:
@svanoort Made some changes - thoughts welcome/desired. =)

jglick:
> a Bad Thing to reimplement all of `StageStep`

You mean this?

```java
getContext().newBodyInvoker().withCallback(BodyExecutionCallback.wrap(getContext())).withDisplayName(step.name).start();
return false;
```

abayer:
@jglick Plus `StageAction`, plus getting Blue Ocean and Stage View to grok it, etc...

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -193,21 +190,15 @@ public class Utils {
...
         stages.removeSourceLocation()
 
-        CpsThread thread = CpsThread.current()
-        CpsFlowExecution execution = thread?.execution
-
-        if (execution != null && !execution.complete) {
-            execution.addListener(new SyntheticStageGraphListener())
-        }
-
+        r.addAction(new SyntheticStageGraphListener.GraphListenerAction())
```
svanoort: Check if action exists before adding it.

svanoort: 🐛

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.IOException;
+import java.util.logging.Logger;
+
+import static java.util.logging.Level.WARNING;
+
+public final class SyntheticStageGraphListener implements GraphListener {
+    private static final Logger LOGGER = Logger.getLogger(SyntheticStageGraphListener.class.getName());
+
+    @Override
+    public void onNewHead(FlowNode node) {
```
svanoort: @abayer Not a huge fan of using the displayName here -- can we do getPersistentAction(LabelAction.class) instead -- there are solid performance reasons why you don't want to do the displayName (much more complex).

abayer: Sure thing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -0,0 +1,106 @@
...
+                if (SyntheticStageNames.postStages().contains(node.getDisplayName())) {
+                    attachTag(node, SyntheticStage.getPost());
+                }
+            }
+        }
+    }
+
+    private void attachTag(FlowNode currentNode, String syntheticContext) {
+        TagsAction tagsAction = currentNode.getPersistentAction(TagsAction.class);
+        if (tagsAction == null) {
```
svanoort: I had an API that would do this in TagsAction at one point... removed it during review IIRC, but maybe worth reconsidering

abayer:
Merging once tests pass!

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+    }
+
+    public static class GraphListenerAction extends InvisibleAction implements RunAction2 {
+        @Override
+        public void onLoad(Run<?, ?> r) {
+            if (r != null && r instanceof WorkflowRun) {
+                WorkflowRun run = (WorkflowRun) r;
+                attachListener(run);
```
jglick: [JENKINS-43055](https://issues.jenkins-ci.org/browse/JENKINS-43055)

PR #98, opened 2017-01-23
[JENKINS-41334] Very very preliminary stage parallelism

abayer:
[JENKINS-41334](https://issues.jenkins-ci.org/browse/JENKINS-41334)

Note - VERY VERY preliminary. To the point where this doesn't actually
support parallel stages yet, it just has the infrastructure for it,
but I wanted to see what else might explode from these changes while
working on other things, so - WIP PR.


abayer:
Moving this to target 1.2 - there's enough discussion that's needed before we lock down design and enough interrelated work with Blue Ocean that I don't want to hold up 1.1 waiting for it.

abayer:
Closing for a newer, cleaner version.

PR #99, opened 2017-01-24
[JENKINS-41185] Initial work on when and/not/or

abayer:
[JENKINS-41185](https://issues.jenkins-ci.org/browse/JENKINS-41185)

Post-1.0 feature, this is just the first work. Still needs negative
testing to make sure it doesn't, y'know, fail horribly when there are
problems, but I'll pick this back up again post-1.0.

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTWhen.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTWhen.java
@@ -37,24 +37,24 @@ import java.util.List;
  */
 public class ModelASTWhen extends ModelASTElement {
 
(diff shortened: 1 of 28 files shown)


rsandell:
On a quick glance it looks to be an api breaking change in pipeline-model-api and we are not supposed to have those post 1.0

abayer:
@rsandell That's an interesting point. The guarantee we've made is syntax compatibility, not backend API compatibility - this change wouldn't break either `Jenkinsfile` syntax or JSON representation. I'll do a check to make sure nothing out there is depending on `pipeline-model-api`'s API, but I'm fairly sure there isn't at this point, so we should still be ok.

rsandell:
Well API compatibility is imho just a default strive in the Jenkins project and any deviation from that needs to have motivation. especially on a module whose sole purpose is to be an API :)

abayer:
Replacing with a new PR.

PR #100, opened 2017-01-24
[FIXED JENKINS-41391] Switch skipDefaultCheckout to default true

abayer:
[JENKINS-41391](https://issues.jenkins-ci.org/browse/JENKINS-41391)

That is - make skipDefaultCheckout() work the same as
skipDefaultCheckout(true).

cc @reviewbybees esp @bitwiseman @rsandell


pipeline-model-definition/src/test/resources/multipleProperties.groovy
```diff
@@ -36,7 +36,7 @@ pipeline {
     options {
         buildDiscarder(logRotator(numToKeepStr:'1'))
         disableConcurrentBuilds()
-        skipDefaultCheckout(true)
+        skipDefaultCheckout()
```
bitwiseman: Need a couple more tests, specifically ones where you call skip skipDefaultCheckout twice, once with one value then again with a different value.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/OptionsTest.java
```diff
@@ -74,6 +74,8 @@ public void multipleProperties() throws Exception {
 
         WorkflowJob p = b.getParent();
 
+        // We test for skipDefaultCheckout() in the Jenkinsfile itself by verifying that Jenkinsfile isn't in the workspace
```
bitwiseman: I don't understand this comment.

abayer: If you look at `src/test/resources/multipleProperties.groovy`, you can see that it includes a `sh "test -f Jenkinsfile"` call - i.e., if there's a `Jenkinsfile` in the workspace, the build will pass. Which isn't the desired result - it should fail, since we skipped the default checkout, and it does, as expected.

pipeline-model-definition/src/test/resources/multipleProperties.groovy
abayer: So...what that actually means is that we should have a validation check to make sure that the `options` are unique, 'cos god only knows what the expected behavior of `properties([...])` would be. So...I'll add that check.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -343,16 +342,21 @@ class ModelValidatorImpl implements ModelValidator {
...
+                            valid = false
+                        }
                     } else {
                         requiredParams.eachWithIndex { DescribableParameter entry, int i ->
                             def argVal = meth.args.get(i)
                             if (argVal instanceof ModelASTMethodCall) {
                                 valid = validateElement((ModelASTMethodCall) argVal)
                             } else {
                                 if (!validateParameterType((ModelASTValue) argVal, entry.erasedType)) {
-                                    errorCollector.error((ModelASTValue)argVal, Messages.ModelValidatorImpl_WrongParameterType(entry.name))
```
rsandell: I don't understand why this needed removing

abayer: Because it was redundant - we are already reporting errors in `validateParameterType`.

pipeline-model-definition/src/test/resources/multipleProperties.groovy
bitwiseman: That makes sense as long as the `skipDefaultCheckout()` step can only ever be called from inside the `options` block.   Could I call it in a library?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -409,7 +409,7 @@ class ModelParser implements Parser {
         def mc = matchMethodCall(st);
         if (mc == null) {
             if (st instanceof ExpressionStatement && st.expression instanceof MapExpression) {
-                errorCollector.error(thisProp, Messages.ModelParser_MapNotAllowed(Messages.Parser_Options()))
+                errorCollector.error(thisOpt, Messages.ModelParser_MapNotAllowed(Messages.Parser_Options()))
```
bitwiseman: Why didn't this throw when it was `thisProp`?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -343,16 +342,21 @@ class ModelValidatorImpl implements ModelValidator {
                     List<DescribableParameter> requiredParams = model.parameters.findAll { it.isRequired() }
 
                     if (requiredParams.size() != meth.args.size()) {
-                        errorCollector.error(meth, Messages.ModelValidatorImpl_WrongNumberOfStepParameters(meth.name, requiredParams.size(), meth.args.size()))
-                        valid = false
+                        // TODO: This is a specialized hack for allowing single-required-Boolean-parameter constructors
```
bitwiseman: What is the TODO saying we'll need to do eventually?  Whatever it is file a JIRA and reference it.  Otherwise, it is just a comment not a TODO.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/options/impl/SkipDefaultCheckout.java
```diff
@@ -30,16 +30,18 @@
 import org.jenkinsci.plugins.pipeline.modeldefinition.options.DeclarativeOptionDescriptor;
 import org.kohsuke.stapler.DataBoundConstructor;
 
+import javax.annotation.Nullable;
+
 public class SkipDefaultCheckout extends DeclarativeOption {
-    boolean skipDefaultCheckout;
+    private Boolean skipDefaultCheckout;
```
bitwiseman: Isn't `private` the default?  
I don't know the style guide for jenkinsci, but it didn't have `private` before.

bitwiseman:
@abayer - Only a couple questions.  Functionally, looks good.

pipeline-model-definition/src/test/resources/multipleProperties.groovy
abayer: Nope. It's utterly irrelevant outside of `options`.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: Because we never actually hit this particular path. Would make sense to expand the testing going forward, but not in the scope of this PR.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
abayer: I leave TODOs as notes for things that should be reevaluated periodically.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/options/impl/SkipDefaultCheckout.java
abayer: Without an explicit scope, it's package-scoped.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
bitwiseman: Capture this particular path (or set of paths) in a JIRA.  Otherwise, it will be forgotten.

abayer: [JENKINS-41448](https://issues.jenkins-ci.org/browse/JENKINS-41448)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
bitwiseman: For me TODOs in code are punch lines to jokes you'll tell a couple years from now. 😄 

Put another way, think of this from the perspective of not-you.  Not-you sees `TODO` but no clear statement of what is _to be done_.  What does that mean?   You know what you mean, but not-you does not. 

I see `TODO` I expect as statement of what is to be done.  
```
//TODO: revisit post 1.0
// This allows single-required-Boolean-parameter constructors
// to be called like "foo()", Groovy-style, passing null as the parameter value.
// <description of why this needs revisiting> 
```

bitwiseman:
:bee:

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
abayer: Reworded, removed `TODO`.

PR #101, opened 2017-01-25
Correct the synthetic "Post" stage label

rtyler:
Once upon a time the syntax was called "postBuild' but was renamed to "post"
and this label never got updated.


bitwiseman:
🐝

PR #102, opened 2017-01-26
Clarifying agent type name validation

abayer:
Better messaging when there is no agent type at all, and actually
calling out (and highlighting) the agent type when it's invalid, or is
missing required parameters.

@HRMPW - I'm leaning towards merging this and #101, seeing what else along these lines pops up by tomorrow morning and then cutting an 0.9.1 release. Thoughts?


HRMPW:
:bee: does this have any affect on the validation for any other `agent` syntax like forgetting to put in an `agent` at all.

abayer:
Nope, just covers the cases of `agent { }` and `agent { banana 'monkey' }`, which were already reporting as errors, just without any distinction between the two.

stephenc:
:bee:

PR #104, opened 2017-01-26
[FIXED JENKINS-41490, JENKINS-41491] Tree steps and validation!

abayer:
[JENKINS-41490](https://issues.jenkins-ci.org/browse/JENKINS-41490), [JENKINS-41491](https://issues.jenkins-ci.org/browse/JENKINS-41491)

So this started with fixing JENKINS-41490, so that the editor can
actually do nested tree steps (d'oh), but in the process, I discovered
a strange decision from waaaaaay back in the day to force validation
in certain cases to treat the step parameter type as a String, even
when it wasn't one. That...was bad. So, fixing both those things.

Note that these are both blockers - JENKINS-41490 is hosing the editor and JENKINS-41491 is just plain stupid. So long as this doesn't cause any test regressions (which it hasn't in my local testing), this is going in 0.9.1.

### DIFF
--- a/pipeline-model-api/src/main/resources/ast-schema.json
+++ b/pipeline-model-api/src/main/resources/ast-schema.json
@@ -228,7 +228,14 @@
           "type": "array",
           "minItems": 1,
           "items": {
-            "$ref": "#/definitions/step"
+            "anyOf": [
+              {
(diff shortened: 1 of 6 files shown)


stephenc:
:bee:

PR #105, opened 2017-01-27
[FIXED JENKINS-41518] Validate env var names for validity

abayer:
[JENKINS-41518](https://issues.jenkins-ci.org/browse/JENKINS-41518)

Specifically for validity as Java identifiers - otherwise everything
'splodes. This was never a problem from the Jenkinsfile side, since
there'd be compilation errors before getting to validation if you used
an invalid identifier, but it *is* a problem coming from JSON.

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -57,6 +57,7 @@ import org.jenkinsci.plugins.workflow.steps.StepDescriptor
 import org.jenkinsci.plugins.workflow.support.steps.StageStep
 
 import javax.annotation.Nullable
+import javax.lang.model.SourceVersion
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
 import java.util.concurrent.TimeUnit
@@ -387,5 +388,14 @@ public class Utils {
         wrapper.setClosure(c)
 
         return wrapper
+    }
+
+    public static boolean validEnvIdentifier(String i) {
+        if (!SourceVersion.isIdentifier(i)) {
+            return false
+        } else if (!i.matches("[a-zA-Z_]+[a-zA-Z0-9_]*")) {
+            return false
+        }
+        return true
     }
 }
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
@@ -35,6 +35,7 @@ import jenkins.model.Jenkins
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter
 import org.jenkinsci.plugins.pipeline.modeldefinition.DescriptorLookupCache
 import org.jenkinsci.plugins.pipeline.modeldefinition.Messages
+import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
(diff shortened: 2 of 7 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -126,6 +127,12 @@ class ModelValidatorImpl implements ModelValidator {
             errorCollector.error(env, Messages.ModelValidatorImpl_NoEnvVars())
             valid = false
         }
+        env.variables.each { k, v ->
+            if (!SourceVersion.isIdentifier(k.key)) {
```
rsandell: A valid java identifier would be `$myvar` but I'm unsure if it would be a valid env var?

http://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8

abayer: Hrrrrrm.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/Messages.properties
```diff
@@ -117,5 +117,5 @@ ModelValidatorImpl.MissingAgentParameter=Missing required parameter for agent ty
 ModelValidatorImpl.InvalidAgentParameter=Invalid config option "{0}" for agent type "{1}". Valid config options are {2}
 ModelValidatorImpl.UnknownWhenConditional=Unknown conditional {0}. Valid conditionals are: {1}
 ModelValidatorImpl.MultipleAgentParameters=Multiple parameters are required for agent type "{0}" - {1}. Please use a block instead.
-
+ModelValidatorImpl.InvalidIdentifierInEnv="{0}" is not a valid Java identifier and cannot be used for an environment variable.
```
rsandell: I would skip the mentioning of Java. Simply go with `is not a valid identifier`. No need to confuse or angry those who don't want to do the programming.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
abayer: Yup, that's correct. Needs to be addressed. Good catch!

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/Messages.properties
abayer: Will do.

abayer:
@rsandell Take your nits and shove them - I thought those both merited changes. =) Which are in now!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -388,4 +389,13 @@ public class Utils {
 
         return wrapper
     }
+
+    public static boolean validEnvIdentifier(String i) {
+        if (!SourceVersion.isIdentifier(i)) {
+            return false
+        } else if (!i.matches("[a-zA-Z_]+[a-zA-Z0-9_]*")) {
```
rsandell: I guess this doesn't happen often enough to merit a `Pattern.compile`.

abayer: Pretty much, yeah.

PR #106, opened 2017-02-06
[JENKINS-41645] Better validation for non-binary expressions in env

abayer:
[JENKINS-41645](https://issues.jenkins-ci.org/browse/JENKINS-41645)

We can't handle all invalid environment identifiers currently - things
like 123_ABC or something-with-hyphens or 'somethingInQuotes' all end
up failing out at initial Groovy compilation, before we get the AST to
work with. We do catch those when coming from JSON, which is
something. But until this, we didn't give useful errors for any
invalid environment identifier that actually gets through compilation
- i.e., 3D_GLASSES, 123ABC. Those aren't BinaryExpressions, so they
were skipped and we ended up with just the general error case
reporting. This commit changes that. =)

cc @reviewbybees esp @rsandell and @orrc

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
@@ -492,6 +492,14 @@ public class ValidatorTest extends AbstractModelDefTest {
                 .go();
     }
 
+    @Issue("JENKINS-41645")
+    @Test
+    public void invalidEnvironmentIdentifiers() throws Exception {
+        expectError("invalidEnvironmentIdentifiers")
+                .logContains(Messages.ModelParser_InvalidEnvironmentIdentifier("1FOO = \"BLARGH\""))
+                .go();
+    }
+
     @Issue("JENKINS-39799")
     @Test
     public void badPostContent() throws Exception {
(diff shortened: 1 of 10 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
```diff
@@ -484,6 +484,14 @@ public void notificationsSectionRemoved() throws Exception {
                 .go();
     }
 
+    @Issue("JENKINS-41645")
+    @Test
+    public void invalidEnvironmentIdentifiers() throws Exception {
+        expectError("invalidEnvironmentIdentifiers")
+                .logContains(Messages.ModelParser_InvalidEnvironmentIdentifier("1FOO = \"BLARGH\""))
```
rsandell: what about the other invalid formats? 'some-darn-key' for example.

abayer: Valid point. The other formats mentioned in the JIRA barf out before we get to validation, so I didn't think to add them. Lemme consider that.

abayer:
@rsandell Not sure what test cases you mean?

rsandell:
@abayer there is only one test case testing one invalid case, no edge cases or other combinations of bad names.

abayer:
@rsandell Ah, ok.

abayer:
@rsandell Added better tests.

PR #109, opened 2017-02-09
[FIXED JENKINS-41900] Expose doCheckout on DeclarativeAgent

abayer:
* JENKINS issue(s):
    * [JENKINS-41900](https://issues.jenkins-ci.org/browse/JENKINS-41900)
* Description:
    * Reduces dependency load for other plugins extending DeclarativeAgent. Relevant for [JENKINS-41758](https://issues.jenkins-ci.org/browse/JENKINS-41758)
* Documentation changes:
    * n/a - no user-facing impact.
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -73,7 +73,7 @@ public class ModelInterpreter implements Serializable {
                         // Stage execution and post-build actions run in try/catch blocks, so we still run post-build actions
                         // even if the build fails.
                         // We save the caught error, if any, for throwing at the end of the build.
-                        inDeclarativeAgent(root, root.agent) {
+                        inDeclarativeAgent(root, root, root.agent) {
                             withCredentialsBlock(root.getEnvCredentials()) {
                                 toolsBlock(root.agent, root.tools) {
                                     for (int i = 0; i < root.stages.getStages().size(); i++) {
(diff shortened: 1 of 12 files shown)


abayer:
ping @rsandell - ignore the yellow, that's my fault due to doing some dumb test runs on another master. =)

abayer:
@rsandell Very good point. Let me think about that - may want to actually tweak logic.

abayer:
@rsandell Check out https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/109/commits/1ba916c76118a838fd76793070cb6a60d6f199fe - I think it handles the running Pipeline case. Good catch!

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -284,17 +284,18 @@ public class ModelInterpreter implements Serializable {
      * Executes the given closure inside a declarative agent block, if appropriate.
      *
      * @param context Either a stage or root object, the context we're running in.
+     * @param root The root object for this pipeline
      * @param agent The agent context we're running in
      * @param body The closure to execute
      * @return The return of the resulting executed closure
      */
-    def inDeclarativeAgent(Object context, Agent agent, Closure body) {
+    def inDeclarativeAgent(Object context, Root root, Agent agent, Closure body) {
```
rsandell: What about this one?

abayer: That one shouldn't be a problem - anything within ModelInterpreter will already be loaded. I think.

rsandell: What we really need is some `@LocalData` tests with already running pipelines in them that are supposed to be resumed on startup and should succeed. So we don't feel uncomfortable in guessing we should be fine.

abayer: Agreed. Opened [JENKINS-42114](https://issues.jenkins-ci.org/browse/JENKINS-42114) for that.

PR #111, opened 2017-02-10
[FIXED JENKINS-41911] Shade away Jackson

abayer:
* JENKINS issue(s):
    * [JENKINS-41911](https://issues.jenkins-ci.org/browse/JENKINS-41911)
* Description:
    * This took some hoop-jumping, but now we should be guaranteed to use our own copy of Jackson at all times. Phew. Note that I added a test dependency on the `jackson2-api` plugin, basically to prove that this works, since the tests fail hard without the shade change but with that plugin.
* Documentation changes:
    * n/a - purely internal change, not at all user facing.
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @kzantow

### DIFF
--- a/pipeline-model-json-shaded/pom.xml
+++ b/pipeline-model-json-shaded/pom.xml
@@ -1,7 +1,7 @@
 <!--
   ~ The MIT License
   ~
-  ~ Copyright (c) 2016, CloudBees, Inc.
+  ~ Copyright (c) 2017, CloudBees, Inc.
   ~
   ~ Permission is hereby granted, free of charge, to any person obtaining a copy
   ~ of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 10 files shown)


michaelneale:
LGTM

rsandell:
So in the comments on the issue you mentioned that Guava was an issue as well? Don't we need to shade that as well for jackson to use?

abayer:
@rsandell Switched to a more specific shaded package. The Guava thing was that newer versions of `json-schema-validator` which depend on a compatible version of Jackson to the `jackson2-api` plugin also depend on excessively new Guava versions - the version of `json-schema-validator` here is fine.

abayer:
Waiting for an actual test run to finish - ci.jenkins.io has been hanging jobs lately. sigh.

abayer:
@rsandell @kzantow Need re-reviews on this due to more shading pain needed. Sigh.

pipeline-model-json-shaded/pom.xml
```diff
@@ -54,12 +54,34 @@
             <includes>
               <include>com.github.fge:*</include>
               <include>com.fasterxml*:*</include>
+              <include>joda-time:joda-time:*</include>
+              <include>com.google.guava:guava:*</include>
+              <include>com.googlecode.libphonenumber:*</include>
+              <include>org.mozilla:rhino:*</include>
```
rsandell: wow, that is probably a big one!

pipeline-model-json-shaded/pom.xml
```diff
@@ -0,0 +1,104 @@
+<!--
```
kzantow: Only nit is the module name now that it's shading a bunch of stuff.

pipeline-model-json-shaded/pom.xml
```diff
@@ -54,12 +54,34 @@
             <includes>
               <include>com.github.fge:*</include>
               <include>com.fasterxml*:*</include>
+              <include>joda-time:joda-time:*</include>
+              <include>com.google.guava:guava:*</include>
+              <include>com.googlecode.libphonenumber:*</include>
+              <include>org.mozilla:rhino:*</include>
+              <include>javax.mail:mailapi:*</include>
```
rsandell: has `javax.mail` ever changed?

abayer:
@rsandell I went for a "screw it, shade eeeeeerything" approach when I kept hitting more problems while iterating. Ah well.

pipeline-model-json-shaded/pom.xml
abayer: No, but better safe than sorry.

pipeline-model-json-shaded/pom.xml
abayer: So all the dependencies shaded in here are directly related to either `json-schema-validator` or `jackson-datatype-json-org` - we're not invoking any of the other shaded dependencies directly anywhere, just the JSON stuff.

pipeline-model-json-shaded/pom.xml
abayer: Yyyyup. But again, better safe than sorry. The freakin' `libphonenumber` ended up being a problem, so I just went for everything at that point.

PR #110, opened 2017-02-09
JENKINS-41748, JENKINS-41890 - Fix env var references and WORKSPACE

abayer:
* JENKINS issue(s):
    * [JENKINS-41748](https://issues.jenkins-ci.org/browse/JENKINS-41748)
    * [JENKINS-41890](https://issues.jenkins-ci.org/browse/JENKINS-41890)
* Description:
    * Fixes env var references in `environment` section so that referring to existing env vars (either from the overall build context or, when in a `stage`, in the top-level `environment` actually works right.
    * Switches ordering of evaluation of `environment` and `agent` at top-level so that we already have a `WORKSPACE` (if not `agent none`) before evaluating the `environment` - this is [JENKINS-41890](https://issues.jenkins-ci.org/browse/JENKINS-41890), which depended on the other change and didn't seem to merit a separate PR.
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/626
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @aheritier

PR #112, opened 2017-02-10
JENKINS-41759 read properties into environment

rsandell:
* JENKINS issue(s):
    * [JENKINS-41759](https://issues.jenkins-ci.org/browse/JENKINS-41759)
* Description:
    * Refactored the pluggable functions in environment directive to enable extensions and made credentials into such a _(special)_ extension. Then added the `fromPropertiesFile` function that utilizes readTrusted so it doesn't require a workspace.
* Documentation changes:
    * Yes, needs documentation but I don't know how
* Users/aliases to notify:
    * @reviewbybees 
    * @abayer

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
@@ -92,39 +92,6 @@ public class Stage implements NestedModel, Serializable {
         return this
     }
 
-    /**
-     * Helper method for translating the key/value pairs in the {@link Environment} into a list of "key=value" strings
-     * suitable for use with the withEnv step.
-     *
-     * @return a list of "key=value" strings.
-     */
-    List<String> getEnvVars(CpsScript script) {
-        List<String> e =  environment.findAll{k, v -> !(v instanceof DeclarativeEnvironmentContributor)}.collect { k, v ->
-            "${k}=${v}"
-        }
-        environment.each {k, v ->
-            if (v instanceof DeclarativeEnvironmentContributor && !(v instanceof DeclarativeEnvironmentContributor.MutedGenerator)) {
-                List<String> ee = v.generate(script, k)
-                if (ee != null) {
-                    e.addAll(ee)
-                }
-            }
-        }
-        return e
-    }
-
-    @Nonnull
-    Map<String, Credentials> getEnvCredentials() {
-        Map<String, Credentials> m = [:]
-        environment.each {k, v ->
-            if (v instanceof  Credentials) {
-                m["${k}"] = v;
-            }
-        }
-        return m
-    }
-
-
     @Override
     public void modelFromMap(Map<String,Object> m) {
         m.each { k, v ->
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
@@ -26,7 +26,10 @@ package org.jenkinsci.plugins.pipeline.modeldefinition.model
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
 import groovy.transform.EqualsAndHashCode
 import groovy.transform.ToString
+import org.jenkinsci.plugins.pipeline.modeldefinition.environment.DeclarativeEnvironmentContributor
+import org.jenkinsci.plugins.pipeline.modeldefinition.environment.impl.Credentials
 import org.jenkinsci.plugins.pipeline.modeldefinition.steps.CredentialWrapper
+import org.jenkinsci.plugins.workflow.cps.CpsScript
 
 import javax.annotation.Nonnull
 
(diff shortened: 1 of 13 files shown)


rsandell:
@batmat Is this something in the area of what you requested?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/environment/impl/TrustedProperties.java
```diff
@@ -0,0 +1,54 @@
...
+    @DataBoundConstructor
+    public TrustedProperties(String path) {
+        this.path = path;
+    }
+
+    public String getPath() {
+        return path;
+    }
+
+    @Extension @Symbol("fromPropertiesFile")
```
rsandell: not super liking the name, but `properties` is a "reserved word" in pipeline so cannot use that.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/PropertiesToMapTranslator.groovy
```diff
@@ -52,16 +57,18 @@ public class PropertiesToMapTranslator implements MethodMissingWrapper, Serializ
...
         }
 
-        if (resolveCredentials && s == "credentials") {
-            if (args.length == 1) {
-                String id = "${args[0]}"
-
-                RunWrapper currentBuild = script.getProperty("currentBuild")
+        if (resolveEnvironmentContributors) {
+            def retVal
+            if (argValue != null) {
```
rsandell: Here is the reason why we can't use `properties` as a special name in `environment` i.e. override what properties normally means. If we could find some other way of getting the `UninstantiatedDescribable` we could potentially override names of already declared steps here.

rsandell:
whaaat???, It worked on my machine!

abayer:
FYI, rebased on master.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/environment/impl/TrustedProperties.java
abayer: Yup, we had the same problem with `SCMTrigger` and `scm` as a symbol.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/PropertiesToMapTranslator.groovy
abayer: This is actually an interesting question in the context of #110, where we stop actually evaluating the `PropertiesToMapTranslator` entirely - with that, we may be able to switch to `@Symbol("properties")` after all.

bitwiseman:
Neat change.  

Is there any way in here to read propertes directly into the environment? 

```
// my.props
A=myA
B=myB

// Jenkinsfile
pipeline {
    environment {
        P = fromPropertiesFile('my.props')
        // what I want is
        // fromPropertiesFile('my.props')
    }
}
```

I'll end up with env vars `P_A` and `P_B`?  
What if I want `A` and  `B`?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/environment/impl/TrustedProperties.java
jtnord: Given it is trusted why not trustedProperties or readTrustedProperties,  sourceTrustedProps or load....  I think that it is trusted is important in this context.

rsandell:
@bitwiseman I did some preparations for `_ = fromPropertiesFile('my.props')` but I haven't tested it fully. Another idea me and @abayer was tossing around when discussing it was to add a similar directive to `options {}` and there it would not have a prefix at all.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
```diff
@@ -26,7 +26,10 @@ package org.jenkinsci.plugins.pipeline.modeldefinition.model
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
 import groovy.transform.EqualsAndHashCode
 import groovy.transform.ToString
+import org.jenkinsci.plugins.pipeline.modeldefinition.environment.DeclarativeEnvironmentContributor
```
rsandell: Unused imports, sometimes I really miss my checkstyle rules that I use in other plugins.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/environment/DeclarativeEnvironmentContributorDescriptor.java
```diff
@@ -0,0 +1,36 @@
...
+ * THE SOFTWARE.
+ *
+ */
+
+package org.jenkinsci.plugins.pipeline.modeldefinition.environment;
+
+import org.jenkinsci.plugins.pipeline.modeldefinition.withscript.WithScriptDescriptor;
+
+/**
+ * Created by rsandell on 2017-02-06.
```
rsandell: Needs better javadoc

rsandell:
Hmm I can't push to this branch anymore since @abayer did the rebase. New changes are on [this branch](https://github.com/jenkinsci/pipeline-model-definition-plugin/tree/JENKINS-41759-a) instead.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/environment/impl/TrustedProperties.java
HRMPW: Won't that just confuse users @jtnord ?  That seems like an implementation detail that users wouldn't need to know about and implies that there would be untrusted properties as well.

Is "from" necessary? `propertiesFile` would seem to be intuitive enough

jtnord: There already are untrusted properties..... 
And it is called readProperties

rsandell: Well in this case 'readProperties' is a step that I don't think can be called inside 'environment'

rsandell:
Replaced by #116

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/PropertiesToMapTranslator.groovy
rsandell: Yepp, I fixed it in the other PR, there is now a util method that does about the same thing as `script."${name}"(args)` would do but you can search for a more specific `Describable` descendant. :-D

PR #113, opened 2017-02-11
[FIXED JENKINS-41950] Properly report errors outside stages

abayer:
* JENKINS issue(s):
    * [JENKINS-41950](https://issues.jenkins-ci.org/browse/JENKINS-41950)
* Description:
    * We hadn't been reporting errors in the pre-stage blocks as `firstError`, so they were getting swallowed, particularly if the `post` that gets run if there are errors pre-stage itself had errors (i.e., due to running `deleteDir()` outside a node context. So now let's catch said pre-stage errors, report them properly, etc.
* Documentation changes:
    * n/a - internal fix.
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AgentTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AgentTest.java
@@ -141,6 +141,19 @@ public class AgentTest extends AbstractModelDefTest {
                 .go();
     }
 
+    @Issue("JENKINS-41950")
+    @Test
+    public void nonExistentDockerImage() throws Exception {
+        assumeDocker();
+        // Bind mounting /var on OS X doesn't work at the moment
+        onAllowedOS(PossibleOS.LINUX);
+
+        expect(Result.FAILURE, "nonExistentDockerImage")
+                .logContains("ERROR: script returned exit code 1")
+                .go();
+    }
+
+
     @Test
     public void fromDockerfile() throws Exception {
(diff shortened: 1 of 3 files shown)


pipeline-model-definition/src/test/resources/nonExistentDockerImage.groovy
```diff
@@ -0,0 +1,48 @@
...
+        stage("foo") {
+            steps {
+                sh 'cat /usr/local/apache2/conf/extra/httpd-userdir.conf'
+                sh 'echo "The answer is 42"'
+            }
+        }
+    }
+    post {
+        always {
+            deleteDir()
```
rsandell: Should leave an output here to check that the post section got executed

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AgentTest.java
```diff
@@ -141,6 +141,19 @@ public void agentAnyInStage() throws Exception {
...
 
+    @Issue("JENKINS-41950")
+    @Test
+    public void nonExistentDockerImage() throws Exception {
+        assumeDocker();
+        // Bind mounting /var on OS X doesn't work at the moment
+        onAllowedOS(PossibleOS.LINUX);
+
+        expect(Result.FAILURE, "nonExistentDockerImage")
+                .logContains("ERROR: script returned exit code 1")
```
rsandell: Should check that post ran as well.

abayer:
ping @rsandell

PR #115, opened 2017-02-14
[FIXED JENKINS-42039] Add option for skipping stages after unstable

abayer:
* JENKINS issue(s):
    * [JENKINS-42039](https://issues.jenkins-ci.org/browse/JENKINS-42039)
* Description:
    * Adds a `DeclarativeOption` for skipping stages once the build status has gone to `UNSTABLE`. Right now, this also sets subsequent stages with a new tag saying "this was skipped due to the build being unstable now" but that may need to switch to just use the same "skipped for failure" one used for post-failure skipped stages if Blue Ocean wants to wait to add visualization.
* Documentation changes:
    * To be done - will be needed.
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @jtnord 
    * @michaelneale @vivek @kzantow @i386 for commentary from the Blue Ocean perspective.


michaelneale:
OMG yes! Wonderful feature wise (I am sure the code is totes legit, but will let others weigh in on that). 

In an ideal world I would love skipping to be the default and to opt out of it, as it seems to surprise pretty much everyone that uses pipeline ever (especially given that post steps can catch the test results, as one example). 

on that note - do post steps still run as appropriate? If so - MASSIVE LGTM from me.

bitwiseman:
Do we already skip stages after failed?

michaelneale:
@bitwiseman failed - yes. Unstable - no.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/options/impl/SkipStagesAfterUnstable.java
```diff
@@ -0,0 +1,51 @@
...
+import hudson.Extension;
+import org.jenkinsci.Symbol;
+import org.jenkinsci.plugins.pipeline.modeldefinition.options.DeclarativeOption;
+import org.jenkinsci.plugins.pipeline.modeldefinition.options.DeclarativeOptionDescriptor;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nullable;
+
+public class SkipStagesAfterUnstable extends DeclarativeOption {
+    private Boolean skipStagesAfterUnstable;
```
rsandell: A bit redundant naming isn't it?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/options/impl/SkipStagesAfterUnstable.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @DataBoundConstructor
+    public SkipStagesAfterUnstable(@Nullable Boolean s) {
+        this.skipStagesAfterUnstable = s;
+    }
+
+    public boolean isSkipStagesAfterUnstable() {
```
rsandell: simply `is()`, `get()` or `should()`? I'm not sure what the bean standard says in this case :)

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -135,6 +135,25 @@ public void allStagesExist() throws Exception {
...
+        FlowExecution execution = b.getExecution();
+        List<FlowNode> heads = execution.getCurrentHeads();
+        DepthFirstScanner scanner = new DepthFirstScanner();
+        FlowNode startFoo = scanner.findFirstMatch(heads, null, Utils.isStageWithOptionalName("foo"));
+        assertNotNull(startFoo);
+        assertTrue(startFoo instanceof StepStartNode);
+        FlowNode endFoo = scanner.findFirstMatch(heads, null, Utils.endNodeForStage((StepStartNode)startFoo));
+        assertNotNull(endFoo);
+        assertEquals(GenericStatus.UNSTABLE, StatusAndTiming.computeChunkStatus(b, null, startFoo, endFoo, null));
+    }
```
rsandell: A tests that makes sure the default also works would be good.

abayer: Good point.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/options/impl/SkipStagesAfterUnstable.java
abayer: Yeaaaaah. Lemme think on this a bit.

abayer:
@rsandell Updated with a simpler `SkipStagesAfterUnstable` class and a default case test.

PR #116, opened 2017-02-15
[JENKINS-41759] read properties into environment

rsandell:
* JENKINS issue(s):
    * [JENKINS-41759](https://issues.jenkins-ci.org/browse/JENKINS-41759)
* Description:
    * Refactored the pluggable functions in environment directive to enable extensions and made credentials into such a _(special)_ extension. Also changed how Describables are "instantiated" so we can override what the scripted DSL does and search for more specific descriptor/describable extensions. Then added the `properties` function that utilizes readTrusted so it doesn't require a workspace.
* Documentation changes:
    * Yes, needs documentation but I don't know how
* Users/aliases to notify:
    * @reviewbybees 
    * @abayer


abayer:
@rsandell Needs a merge/rebase. Docs are in https://github.com/jenkins-infra/jenkins.io, specifically in https://github.com/jenkins-infra/jenkins.io/blob/master/content/doc/book/pipeline/syntax.adoc.

Need to discuss the naming etc over on #112 here as well - paging in @jtnord @HRMPW and @bitwiseman from that thread.

bitwiseman:
@rsandell 
> Then added the properties function that utilizes readTrusted so it doesn't require a workspace.

But if there is a workspace and the file is in the workspace it will be read right? 

Docs need to be part of a change like this.  If you need any help understanding how to do them, I'm happy to chat.  See #110 and how abayer linked to a doc change.

bitwiseman:
@rsandell @abayer 
Could you give us a new starting point for talking about "naming etc".  What's the question?

rsandell:
@bitwiseman 

> But if there is a workspace and the file is in the workspace it will be read right?

No, I don't think so. It uses `readTrusted` for two reasons; not needing a workspace and making sure a non trusted contributor can't try to do any injection style attacks on the pipeline.

rsandell:
@bitwiseman I guess what @abayer is asking is: "Is `properties('some/gile.properties')` a good name, or should we have something more expressive like `readPropertiesFromFile`?"

abayer:
Well, `properties` isn't an option, since that's a `Step` and I'm deeply wary of overlapping with `Step` names - I'm not even sure we could do that for a function.

rsandell:
@abayer the ability of overriding step names is exactly what 8609c0f is about!

abayer:
Well, I'm dumb and that's kinda brilliant. That said, I'm still not hugely keen on the name `properties` - I'd ideally like it to be  learned that the argument is a file path.

bitwiseman:
@rsandell 
> It uses readTrusted for two reasons; not needing a workspace and making sure a non trusted contributor can't try to do any injection style attacks on the pipeline.

Scenario: 
1. Generate properties during the build.
2. Load those properties as environment variables. 

This is a very common scenario and `readTrusted` won't support it. Only being able to read checked in properties files is not enough. 

It also, if we're talking about injection style attacks, does requiring the file to be checked in solve the problem?  If I create a PR with a changed properties file, I can still get my properties loaded by checking them in.

HRMPW:
@bitwiseman I also asked this morning about using libraryResource to load properties from shared libraries.  If we default to readTrusted but supported other strings or maps it would work to do one of the following:

`readPropertiesFromFile(readProperties("foo"))`  using `pipeline-utility-steps`

`readPropertiesFromFile(libraryResource("foo"))`  using shared libraries

I don't think defaulting to readTrusted is bad as long as we support other alternatives.

rsandell:
@HRMPW commit 5d78778 is a PoC for enabling loading library resources as well and it would be possible to extend to what @bitwiseman wants as well; enable reading from the workspace.

rsandell:
Let's say we also want to introduce other file types (xml, yaml, json etc.) how should we handle those?

We could have the file type be the function name and from where to load it be an argument, or we could have from where to load it be the function name and the file type be an argument, or inferred from the file extension.

eg:
```groovy
environment {
  P = properties data: '/some/path.properties', mode: 'ws'
  Y = yaml data: 'some/path.yaml', mode: 'resources'
  X = xml '/some/scm/path.xml' //read from scm by default
}
```
or
```groovy
environment {
  P = workspace '/some/path.properties'
  Y = scm 'some/path.yaml'
  X = resource 'some/path.data', type: 'xml'
  M = data [OPT1: 'something', OPT2: 'somethingelse'] //could be a map returned from an arbitrary function
}
```

rsandell:
What I'm saying there is that I personally don't like having both from where and the file type in the function name since the combinatorics would be horrifying _imho_. E.g. `propertiesFromScm`, `propertiesFromResource`, `propertiesFromWorkspace`, `yamlFromScm`, `yamlFromResource` etc.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/environment/DeclarativeEnvironmentContributorScript.java
```diff
@@ -0,0 +1,44 @@
...
+ * Script base for {@link DeclarativeEnvironmentContributor}s.
+ */
+public abstract class DeclarativeEnvironmentContributorScript<C extends DeclarativeEnvironmentContributor<C>> extends WithScriptScript<C> {
+
+
+    public DeclarativeEnvironmentContributorScript(CpsScript s, C d) {
+        super(s, d);
+    }
+
+    public abstract List<String> generate(String key);
```
jtnord: what does this do?  How would I expect to be able to implement one of these from the associated Javadoc :-)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/environment/impl/TrustedPropertiesScript.groovy
```diff
@@ -0,0 +1,92 @@
...
+
+        if (describable.data == null) {
+            return Collections.emptyList()
+        } else if (describable.data instanceof Map) {
+            return generateEnvStrings(prefix, (Map)describable.data)
+        } else {
+            String sData = "${describable.data}"
+            String content = null
+            if (sData.contains("=") && sData.contains("\n")) {
+                //It is the contents of a Properties file (have to have at least end with a new line)
```
jtnord: I guess MAC style line ends are just dead these days :)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/environment/impl/TrustedPropertiesScript.groovy
```diff
@@ -0,0 +1,92 @@
...
+                content = sData
+            } else {
+                //treat it as a path
+                if (sData.startsWith("resources://")) { //TODO should be an extension
```
jtnord: NIT: I dislike opaque non registered URLS. 
Why not take a parameter such as `resolver: "scm"` or  `resolver: "workspace"` it is cleaner and `scm://` is already used by Maven with a completely different syntax that what you are proposing to use.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/UtilsTest.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.*;
+
+/**
+ * Tests for {@link Utils}.
+ */
+public class UtilsTest {
```
jtnord: groovy where Java would be just as good 😭

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/environment/impl/TrustedPropertiesTest.java
```diff
@@ -0,0 +1,129 @@
...
+public class TrustedPropertiesTest extends AbstractModelDefTest {
+
+    @Before
+    public void writeMarkers() throws Exception {
+        sampleRepo.write("marker.properties", "NAME=Bobby\nNUM=1\n");
+        sampleRepo.write("stage/marker.properties", "NAME=Andrew\nNUM=0\n");
+        sampleRepo.git("init");
+        sampleRepo.git("add", "marker.properties");
+        sampleRepo.git("add", "stage/marker.properties");
+        sampleRepo.git("commit", "--message=Markers");
```
jtnord: replace the above 3 lines with one line `git commit -a -m Markers"  or `git commit marker.properties stage/marker.properties -m "Markers"`

jtnord:
> We could have the file type be the function name and from where to load it be an argument, or we could have from where to load it be the function name and the file type be an argument, or inferred from the file extension.

You can have both be arguments...

`properties file: "/path/to/file.ext", resolver: "workspace", type: "yaml"`

then just default resolver to workspace (readTrusted) and type to "properties"
I am not sure how you intend to map properties from JSON or YAML which are hierarchical....

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/environment/impl/TrustedPropertiesScript.groovy
abayer: Agreed. I prefer multiple parameters to embedding multiple things in a single parameter.

abayer:
Got some work on explicit type rather than URI (and moving the extension point into `pipeline-model-extensions`) at https://github.com/abayer/pipeline-model-definition-plugin/tree/jenkins-41759-abayer.

abayer:
I'm now experimenting with seeing if we can do something like https://gist.github.com/abayer/35030ab269f7472766a88857ddb4b802 - not there yet and may not be able to pull it off due to the fun fun fun of runtime translation.

abayer:
That work in progress/prototyping is up at https://github.com/abayer/pipeline-model-definition-plugin/tree/jenkins-41759-nested, fwiw. I'm stuck figuring out the runtime model/translation currently, but do have the AST/parsing/validation/JSON/etc done.

abayer:
@rsandell Could we get an RFC/design doc for this, like https://github.com/abayer/pipeline-model-definition-plugin/blob/jenkins-41118/rfc/JENKINS-41118-custom-workspaces.md, added to the PR?

abayer:
Ow, ow, ow. Yeah, I spent way too long trying to experiment with alternate syntaxes...we really need an RFC before moving any further with this.

abayer:
Moving this to 1.2 and unblocking #125, since I'm just not happy with this currently and don't have an answer as to what it should be doing instead.

jglick:
The main problem with `*.properties` is that few people (even Java programmers) can remember the escaping rules in the syntax. Java 9 finally treats them as UTF-8 rather than ISO-8859-1 by default, but there are some unexpected metacharacters. JSON and to some extent YAML are probably more familiar.

I would *hope* that Windows path names are rarely going to be committed in such files, but…

bitwiseman:
@rsandell @abayer 
I've linked this PR to the JIRA for tracking.
The last update to this was over 18 months ago.  Unless we are making further progress on this we should close the PR.

jglick:
I still stand by https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/116#issuecomment-285179870. If you are going to do anything, do YAML.

bitwiseman:
I'm going to close this PR and move the branch to https://github.com/jenkinsci/pipeline-model-definition-plugin/tree/issue/JENKINS-41759-a .  If we decided to do this work at some point in the future it will be saved there.

PR #119, opened 2017-02-21
[FIXED JENKINS-42226] Treat null/empty branch comparison as false

abayer:
* JENKINS issue(s):
    * [JENKINS-42226](https://issues.jenkins-ci.org/browse/JENKINS-42226)
* Description:
    * We should be null-checking `compare` in all cases as well as `actualBranch`, and if either is null/empty, we should return false. This change does that.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @imod


imod:
👍

PR #121, opened 2017-02-21
[FIXED JENKINS-42168] Add validateDeclarativePipeline(path) step

abayer:
* JENKINS issue(s):
    * [JENKINS-42168](https://issues.jenkins-ci.org/browse/JENKINS-42168)
* Description:
    * Given a path to a file containing a Declarative Pipeline to validate, this will read that file from the workspace, validate its contents, echo out relevant errors to the build log, and return true if the contents validated, or false other wise.
* Documentation changes:
    * TBD?
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @jtnord 
    * @batmat

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/ValidateDeclarativePipelineStep.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/ValidateDeclarativePipelineStep.java
@@ -1,0 +1,124 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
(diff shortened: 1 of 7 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/ValidateDeclarativePipelineStep.java
```diff
@@ -0,0 +1,124 @@
...
+        @StepContextParameter
+        private transient FilePath cwd;
+
+        @StepContextParameter
+        private transient TaskListener listener;
+
+        @Override
+        public Boolean run() throws Exception {
+            if (StringUtils.isEmpty(step.getPath())) {
+                listener.getLogger().println("No Declarative Pipeline file specified.");
```
jtnord: the descriptor says "Validate a string containing a Declarative Pipeline" yet we are using files here?

Which one is it `Files` or `Strings`?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/ValidateDeclarativePipelineStep.java
```diff
@@ -0,0 +1,124 @@
...
+        public DescriptorImpl() {
+            super(ValidateDeclarativePipelineStepExecution.class);
+        }
+
+        @Override public String getFunctionName() {
+            return "validateDeclarativePipeline";
+        }
+
+        @Override public String getDisplayName() {
+            return "Validate a string containing a Declarative Pipeline";
```
jtnord: :bug: You are validating a file is a declarative pipeline not a String.

abayer:
@jtnord Whoops, an earlier iteration validated a string and I forgot to change that. Fixed!

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/ValidateDeclarativePipelineStep.java
```diff
@@ -0,0 +1,124 @@
...
+
+        @StepContextParameter
+        private transient FilePath cwd;
+
+        @StepContextParameter
+        private transient TaskListener listener;
+
+        @Override
+        public Boolean run() throws Exception {
+            if (StringUtils.isEmpty(step.getPath())) {
```
rsandell: Since you have `@Inject(optional=true)` I'm guessing that `step` can be null and hence you are risking an NPE here.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/ValidateDeclarativePipelineStep.java
```diff
@@ -0,0 +1,124 @@
...
+                listener.getLogger().println("No Declarative Pipeline file specified.");
+                return false;
+            } else {
+                FilePath f = cwd.child(step.getPath());
```
rsandell: Dito on the NPE unless the above `if` statement fixes it.

abayer:
@rsandell So yeah, I wasn't sure what the right return should be. I decided to go with returning true/false but logging errors. Given that @jtnord requested this in the first place, his input would be valuable. =)

rsandell:
@abayer the logging of the errors, would that be the innocuous looking `e.toString()`?

rsandell:
Ah yes I see that now from the unit tests.

Another alternative _(at a later date)_ could be to add the results as a `TestResultAction` :)

PR #123, opened 2017-02-23
[JENKINS-41118] Allow specifying top level ws(...) block

abayer:
* JENKINS issue(s):
    * [JENKINS-41118](https://issues.jenkins-ci.org/browse/JENKINS-41118)
* Description:
    * Enables custom workspaces in Declarative as an `agent` option.
* RFC document:
    * https://github.com/abayer/pipeline-model-definition-plugin/blob/jenkins-41118/rfc/JENKINS-41118-custom-workspaces.md
* Documentation changes:
    * Doc update will be needed for the new `customWorkspace` option for `label`, `docker` and `dockerfile`.
    * [x] Doc PR up - [jenkins.io PR #715](https://github.com/jenkins-infra/jenkins.io/pull/715)
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
@@ -30,6 +30,7 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.agent.DeclarativeAgent
 import org.jenkinsci.plugins.pipeline.modeldefinition.agent.DeclarativeAgentDescriptor
 import org.jenkinsci.plugins.pipeline.modeldefinition.agent.impl.None
 import org.jenkinsci.plugins.pipeline.modeldefinition.options.impl.SkipDefaultCheckout
+import org.jenkinsci.plugins.structs.SymbolLookup
 import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable
 
 import javax.annotation.CheckForNull
@@ -56,12 +57,10 @@ public class Agent extends MappedClosure<Object,Agent> implements Serializable {
      * @return The instantiated declarative agent or null if not found.
      */
     public DeclarativeAgent getDeclarativeAgent(@CheckForNull Root root, Object context) {
-        DeclarativeAgentDescriptor foundDescriptor = DeclarativeAgentDescriptor.all().find { d ->
-            getMap().containsKey(d.getName())
-        }
-
-        if (foundDescriptor != null) {
-            def val = getMap().get(foundDescriptor.getName())
+        String foundSymbol = findSymbol()
+        if (foundSymbol != null) {
+            DeclarativeAgentDescriptor foundDescriptor = DeclarativeAgentDescriptor.byName(foundSymbol)
+            def val = getMap().get(foundSymbol)
             def argMap = [:]
             if (val instanceof Map) {
                 argMap.putAll(val)
@@ -70,6 +69,7 @@ public class Agent extends MappedClosure<Object,Agent> implements Serializable {
             }
 
             DeclarativeAgent a = DeclarativeAgentDescriptor.instanceForDescriptor(foundDescriptor, argMap)
+
             boolean doCheckout = false
             if (context instanceof Root) {
                 a.setInStage(false)
@@ -88,6 +88,23 @@ public class Agent extends MappedClosure<Object,Agent> implements Serializable {
         } else {
             return null
         }
+    }
+
+    /**
+     * Needed to handle the combination of describable ordinals *and* Descriptor lookup.
+     * @return The first symbol (in descriptor-ordinal-order searching) found in the map.
+     */
+    private String findSymbol() {
+        String sym = null
+        DeclarativeAgentDescriptor.all().each { d ->
+            SymbolLookup.getSymbolValue(d)?.each { s ->
+                if (getMap().containsKey(s) && sym == null) {
+                    sym = s
+                }
+            }
+        }
+
+        return sym
     }
 
     public boolean hasAgent() {
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
@@ -41,22 +41,34 @@ public class LabelScript extends DeclarativeAgentScript<Label> {
         return {
             try {
                 script.node(describable?.label) {
-                    if (describable.isDoCheckout() && describable.hasScmContext(script)) {
-                        if (!describable.inStage) {
-                            script.stage(SyntheticStageNames.checkout()) {
-                                script.checkout script.scm
-                            }
-                        } else {
-                            // No stage when we're in a nested stage already
-                            script.checkout script.scm
+                    if (describable.customWorkspace != null && describable.customWorkspace != "") {
+                        script.ws(describable.customWorkspace) {
+                            checkoutAndRun(body).call()
                         }
+                    } else {
+                        checkoutAndRun(body).call()
                     }
-                    body.call()
                 }
             } catch (Exception e) {
                 script.getProperty("currentBuild").result = Result.FAILURE
                 throw e
             }
+        }
+    }
+
+    private Closure checkoutAndRun(Closure body) {
+        return {
+            if (describable.isDoCheckout() && describable.hasScmContext(script)) {
+                if (!describable.inStage) {
+                    script.stage(SyntheticStageNames.checkout()) {
+                        script.checkout script.scm
+                    }
+                } else {
+                    // No stage when we're in a nested stage already
+                    script.checkout script.scm
+                }
+            }
+            body.call()
         }
     }
 }
--- a/pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
+++ b/pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
@@ -23,9 +23,11 @@
  */
 
 pipeline {
-    agent any
-    options {
-        workspaceDir("some-sub-dir")
+    agent {
+        label {
+            label ""
+            customWorkspace "some-sub-dir"
+        }
     }
     stages {
         stage("foo") {
--- a/pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
+++ b/pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
@@ -24,7 +24,7 @@
 
 pipeline {
     agent {
-        label {
+        node {
             label ""
             customWorkspace "some-sub-dir"
         }
--- a/pipeline-model-definition/src/test/resources/inRelativeCustomWorkspace.groovy
+++ b/pipeline-model-definition/src/test/resources/inRelativeCustomWorkspace.groovy
@@ -1,0 +1,42 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    agent {
+        label {
+            label ""
+            customWorkspace "./relative/custom/workspace/../../custom2/workspace2/./../workspace3"
+        }
+    }
+    stages {
+        stage("foo") {
+            steps {
+                echo "Workspace dir is ${pwd()}"
+            }
+        }
+    }
+}
+
+
+
--- a/pipeline-model-definition/src/test/resources/inRelativeCustomWorkspace.groovy
+++ b/pipeline-model-definition/src/test/resources/inRelativeCustomWorkspace.groovy
@@ -24,6 +24,7 @@
 
 pipeline {
     agent {
+        // This is still deliberately 'label' here for testing label->node conversion in ModelParserTest#labelWithOptionsBecomesNode
         label {
             label ""
             customWorkspace "./relative/custom/workspace/../../custom2/workspace2/./../workspace3"
--- a/rfc/JENKINS-41118-custom-workspaces.md
+++ b/rfc/JENKINS-41118-custom-workspaces.md
@@ -1,0 +1,72 @@
+## RFC: [JENKINS-41118](https://issues.jenkins-ci.org/browse/JENKINS-41118) - custom workspaces in Declarative
+
+* Author: Andrew Bayer
+* Created Date: February 23, 2017
+* Approved Date: n/a
+* JIRA: [JENKINS-41118](https://issues.jenkins-ci.org/browse/JENKINS-41118)
+* Status: Draft
+
+### Problem Statement
+
+As described in the ticket, there are multiple reasons why one 
+may want to use a hardcoded workspace directory under
+`$JENKINS_HOME/workspace` for a Pipeline, such as reusing the
+same checkout of a very large repository, or guaranteeing a
+certain full directory layout to make Golang happy, etc. This
+may be needed both at the top level and per-stage, so that
+individual stages could have a custom workspace when they had
+their own agent configuration as well.
+
+### Runtime Implementation
+
+The actual runtime implementation is very clear - if a custom
+workspace is specified, then immediately after entering the
+relevant `node` block and before doing a `checkout scm`, we
+would enter a `ws('some-dir')` block. 
+
+### Syntax
+
+#### Option Name
+
+Regardless of where the option ends up living in the syntax, I
+propose that it be named `customWorkspace`. This maps to the
+Freestyle "Custom Workspace" option, which is analogous to the
+behavior of the `ws(...)` step (though not exactly, since
+"Custom Workspace" is an absolute path while `ws(...)` is
+relative to the workspace root on the agent).
+
+#### Where the Option is Specified
+The easiest place to add this option so that it would be
+available for use anywhere an `agent` could be specified would
+be in the `agent` itself, as an additional option on relevant
+`DeclarativeAgent` types. That would currently mean `label`,
+`docker`, and `dockerfile` - `any` is a special alias for
+`label ''` and can't be modified, and this would not be
+relevant for `none`. 
+
+For `docker` and `dockerfile`, adding an additional option is 
+trivial, given that they already have additional options for 
+`args` etc. But for `label`, this would be the first additional 
+option and so would create the first case of specifying `label` 
+with a block. However, the block form of `label` actually works 
+currently - `label 'something'` is in fact just a shortcut for 
+`label { label 'something' }` behind the scenes.
+
+#### Examples
+
+```groovy
+agent {
+    label {
+        label "some-label" // This could be changed to something other than "label"
+        customWorkspace "some/path"
+    }
+}
+
+agent {
+    docker {
+        image "some-image"
+        args "-v /tmp:/tmp -p 80:80"
+        customWorkspace "some/path"
+    }
+}
+```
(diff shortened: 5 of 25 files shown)


abayer:
Argh, just realized this isn't gonna work right - it's doing the `ws(...)` step *after* SCM checkout. Grr.

abayer:
Updated to move to `customWorkspace` as an option for `label`, `docker` and `dockerfile` agents.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
```diff
@@ -41,22 +41,34 @@ public class LabelScript extends DeclarativeAgentScript<Label> {
...
                 script.node(describable?.label) {
-                    if (describable.isDoCheckout() && describable.hasScmContext(script)) {
-                        if (!describable.inStage) {
-                            script.stage(SyntheticStageNames.checkout()) {
-                                script.checkout script.scm
-                            }
-                        } else {
-                            // No stage when we're in a nested stage already
-                            script.checkout script.scm
+                    if (describable.customWorkspace != null && describable.customWorkspace != "") {
```
rsandell: This is Groovy; you should be able to shorten it to `if (describable.customWorkspace)`

abayer: For no good reason, I'm always nervous about doing that. =)

abayer:
Rewrote this a bit - I encapsulated the checkout logic so that other Declarative agents can reuse it without having to copy-paste/reinvent the wheel.

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
```diff
@@ -0,0 +1,42 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    agent {
+        label {
+            label ""
```
bitwiseman: `label { label '' }`?  Not loving that progression.

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
```diff
@@ -0,0 +1,42 @@
...
+            customWorkspace "some-sub-dir"
```
bitwiseman: Some questions:
Name: Do we need to call this "custom"?  
Does the workspace have to be relative or could it be absolute? 
If absolute allowed, we need a test for that. 
Need tests showing how this works with `docker` as well. 

It seems like this would make more sense as child of agent, is there a reason why it needs to be inside the children?

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
abayer: Me neither, but what else would I call it inside the `label` block? Suggestions welcome. =)

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
abayer: Since this directly leads to the `ws('some-path')` step, which is basically analogous to the freestyle world "Custom Workspace" functionality, I felt like keeping the name was useful.

Again, since this is `ws('some-path')` on the backend, it's relative to the workspace root on the agent - i.e., the immediate parent of a normal autoassigned workspace directory.

`agent` can only contain Declarative agents - adding options to it as well would require a fair amount of changes.

bitwiseman:
@abayer @rsandell  
Is there a process in place for design review for Declarative?
I am really liking the new syntax. There are great new features and changes going in (like this one). 
I hate arriving late in the process and asking pointed questions about an otherwise finished product.

abayer:
@bitwiseman There is not - @HRMPW should help us create one. =)

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
HRMPW: `name`  is probably the easiest alternative

HRMPW:
Would this work with a `docker` agent or other agent?

abayer:
@HRMPW Yes, this does work within `docker` or `dockerfile`, since they live on top of `label` - i.e., the custom workspace directory will be on the actual node that Docker is running on, which is then mapped into the container as normal.

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
bitwiseman: Label isn't a block now, but it turned into one by this PR?

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
bitwiseman: @abayer - "agent can only contain Declarative agents" - ah, I get it now.

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
abayer: Technically, you could do `label { label "something" }` now and it'd work. `label "something"` is a shortcut, a la `docker "someimage"` or `dockerfile true`, it's just that there's been no reason to do anything *but* the shortcut 'til now.

As I said on the JIRA, I thought hard about whether to make this a top-level option or what, and concluded that `agent` was the right place for it. Another possible approach could be to have it as a top-level option and add per-stage options and have it there as well. I think that could be done, but it adds more clutter to `stage`.

bitwiseman: @abayer @HRMPW 
But it isn't a `name`, that implies "mycomputer".  Label is a...  filter? query? value? specifier? 

Alternatives: 
```
agent {
    label ('value') { 
        workspacePath 'some-dir'
    }
}
```

Or a new top level name:
```
agent {
    customAgent {  // Bad name here, but you get the idea 
        label 'my-filter'
        workspacePath 'some-dir'
    }
}
```

abayer: A new top level name would be problematic because then `label "something"` wouldn't work any more. Sadly. And the `label('value') { ... }` syntax is a non-starter because I hatesssssss it in `stage` and won't let it crop up anywhere else. =)

If we really want to avoid the `label` block, then top-level + per-stage option seems the right way to go. I'm pretty sure I can get it in the right places.

HRMPW: Since `label` already works inside `docker` adding another block specifier would get confusing. I would rather stick with `label` inside `label` than having breaking changes.

abayer:
I've added an RFC doc (at https://github.com/abayer/pipeline-model-definition-plugin/blob/jenkins-41118/rfc/JENKINS-41118-custom-workspaces.md as part of this PR) as a trial run for reviewing design docs before finalizing code. @rtyler - if we continue down this road, could we get a GitHub group set up for Declarative design review notifications?

HRMPW:
I would rather keep this in the `agent` block than as a standalone block. Putting `customWorkspace "foo" ` inside an existing block is much cleaner than creating a new block just for `customWorkspace`

rfc/JENKINS-41118-custom-workspaces.md
```diff
@@ -0,0 +1,81 @@
...
+relevant `node` block and before doing a `checkout scm`, we
+would enter a `ws('some-dir')` block. 
+
+### Syntax
+
+#### Option Name
+
+Regardless of where the option ends up living in the syntax, I
+propose that it be named `customWorkspace`. This maps to the
+Freestyle "Custom Workspace" option, which is analogous to the
```
bitwiseman: The Freestyle UI can say a lot of things.  
on this level we know it's it "custom" because we're setting it.  
Most settings in Declarative are customizations of defaults. 
I'd prefer `workspace`.

rfc/JENKINS-41118-custom-workspaces.md
```diff
@@ -0,0 +1,81 @@
+## RFC: [JENKINS-41118](https://issues.jenkins-ci.org/browse/JENKINS-41118) - custom workspaces in Declarative
+
+* Author: Andrew Bayer
```
bitwiseman: This document is great!  Very detailed.

abayer:
@HRMPW But if we're going to eventually have other per-stage configuration options in the future...

rfc/JENKINS-41118-custom-workspaces.md
```diff
@@ -0,0 +1,81 @@
...
+per-stage option. This would require adding an `options` 
+section for `stage`, since we don't have anywhere for options 
+to be configured per-stage currently. 
+
+#### Examples
+
+```groovy
+agent {
+    label {
+        label "some-label" // This could be changed to something other than "label"
```
bitwiseman: @HRMPW @abayer 
Since this is design discussion, I'll move it down here.  

@abayer  
> A new top level name would be problematic because then label "something" wouldn't work any more. 

Is there no way to overload it?  Make `label` in agent map to the new name? 
I thought groovy was pretty good at this.

Equivalent: 
```
agent {
    label 'mymachine'
}
```

```
agent {
    node {
        label 'mymachine'
    }
}
```

rfc/JENKINS-41118-custom-workspaces.md
abayer: I'm not a huge fan of `workspace` - I really think it's worthwhile to make it clear that this is specifying a non-standard workspace. I honestly didn't really understand the `ws(...)` step until working on this PR, so I do think more clarity than just `workspace` is appropriate.

rfc/JENKINS-41118-custom-workspaces.md
abayer: Perhaps it'd be possible - validation/instantiation is based on the block name mapping to the `@Symbol` on the `DeclarativeAgentDescriptor` for the agent type. Currently we just look at the first value from the `@Symbol` but I suppose we could theoretically look at multiple values if they exist...however, then, both `label` and `node` would work in both block and non-block contexts - there's no viable way I can think of to support one in one context and the other in the other context without a lot of special-casing I really don't want to do.

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
bitwiseman: @abayer 
Even if absolute path doesn't move the workspace to the absolute path, I'd still like to see tests that starts with `/`, `./`, `../`, `../../`.  That this uses `ws` inside is an implementation detail that the tests shouldn't depend on.

abayer: Huh, on further digging, I think `FilePath` may actually end up handling absolute paths. Which sounds kind of awful. I'll look into it and add appropriate tests.

abayer: @bitwiseman One question - testing for `../` and friends is actually a bit hairy, since I don't know what the directory the test will be running in is. So it's hard for me to be able to match a string without knowing what the full path would be ahead of time. I can match the last leaf, for sure, but at least off the top of my head, I can't see an easy way to test for what the parent directory is without knowing what the parent directory should be ahead of time.

rfc/JENKINS-41118-custom-workspaces.md
bitwiseman: @abayer 
Well, I think we all agree `label-label` is pretty bad.  I know you hate `label ('name') { }`, but you have to admit it looks way better than `label { label 'name` }`.


I don't want to get into yak shaving (I've already done enough of that by asking about RFC), but this won't be the last time we run into a problem like this.   We're going to need a way to make changes like this in a non-breaking way.   Filed https://issues.jenkins-ci.org/browse/JENKINS-42294. 

Sticking

rfc/JENKINS-41118-custom-workspaces.md
bitwiseman: @abayer 
Fair point.  You're right.  Thanks!

rfc/JENKINS-41118-custom-workspaces.md
abayer: It may look better, but I *loathe* it. I still really wish we'd gone with `stage { name 'foo'; steps { ... } }` instead. =)

pipeline-model-definition/src/test/resources/inCustomWorkspace.groovy
bitwiseman: @abayer Absolute path should hand the cases outside the default workspace.  How about `./relative/custom/workspace/../../custom2/workspace2/./../workspace3`?

abayer: Ok, will do.

rfc/JENKINS-41118-custom-workspaces.md
bitwiseman: @abayer - So I'm back to this.

```
agent {
    customAgent {  // Bad name here, but you get the idea 
        label 'my-filter'
        workspacePath 'some-dir'
    }
}
```

This doesn't conflict with `node.  
The `agent { customAgent {} }` and `agent { label '' }` would be separate classes, with their own `@Symbol` each.  

What you were saying were problems with `@Symbol` conflicts seems like they would allow this:
```
agent {
    label {
        label {
            label 'test'
        }
    }
}
```

abayer: That...could be done? I'd like thoughts from @michaelneale @kzantow @i386 @hrmpw from the UX/editor perspective.

kzantow: Trying to catch up... this is about having a custom workspace as part of the agent block? And supporting other agent types? Or something else? The first two should be pretty trivial, the agent configuration is primarily driven by metadata...

abayer: So this PR overall is about adding custom workspace support to agents, yes, but the immediate question is whether we should replace/duplicate the `label` agent with a new one that behaves identically but has a different name so as to avoid needing to do `agent { label { label 'foo'; customWorkspace 'bar' } }`.

That'd probably mean that there'd be two agent types that were functionally identical but had different names - we'd need to keep `label` regardless, so `customAgent` (or whatever we'd call it) would be a clone of `label` but named differently for better UX in the Jenkinsfile.

kzantow: seems tangential to agent, to me... could you put it at the same level as the agent config? e.g.:

`agent { label 'asdf'; workspace 'someplace' }` ?
`agent { docker { image 'httpd'; dockerArgs 'whatever' } workspace 'someplace' }`

bitwiseman: Basically, I think we all seem to agreed that this is a good thing:

```
agent {
    <agentType> {
        label 'something'
        customWorkspace 'some/path'
    }
}
```

The problem is that we want to continue to allow the following when `customWorkspace` is not present:

```
agent {
    label 'something'
}
```

but when `customWorkspace` is present, label-label is confusing and odd:
```
agent {
    label {
        label 'something'
        customWorkspace 'some/path'
    }
}
```

kzantow: Assuming it's technically possible, how would that be better than:

```
agent {
  label 'windows'
  customWorkspace 'some/path'
}
```
or
```
agent {
  docker 'httpd...'
  customWorkspace 'some/path'
}
```
or, worst case:
```
agent {
  docker {
    image 'httpd...'
    args 'stuff'
  }
  customWorkspace 'some/path'
}
```

kzantow: workspace is tangential to the agent - you can add a workspace step in your pipeline and change the workspace later on, if you so choose... this is just convenience

abayer: While it's probably theoretically possible to put something that isn't an `agent` in `agent`, I really don't want to go down that route - that's a significant amount of increased code complexity. 

My _first_ choice would honestly be to create an `options` directive for `stage` like we have for `pipeline` and have `customWorkspace` live in those `options` directives - i.e., pull it out of `agent` completely. If that's not viable, then making it a field for `agent`s is the next best option, with making the `agent` directive a mixed container for both `agent` types and configuration for `agent`s far behind.

abayer: @kzantow - it's tangential to `agent` except that we do the `checkout scm` in the `agent` processing, so we need to be able to specify a custom workspace such that the `agent` processing can actually pick it up and call `ws(...) {}` before doing the `checkout`.

kzantow: fair point. from what i can tell, any way this lands should be easy enough to support in the editor, pending a design for it

abayer: If we've got the duplicate `customAgent`, the editor will end up displaying both `label` and `customAgent` as possible agent types, right?

kzantow: yes. but you would need that to differentiate between how it generates a jenkinsfile, since presumably:
```
label {
  label 'blah'
  custom...
}
```

would still be valid.

kzantow: I mean, we can add custom editors for this stuff, but it seems like just specifying this as an overall agent option as i've described would be the easiest for both end users and solve the issue of adding random types just so we don't have `label` twice...

abayer: As I said, I'd strongly prefer having it in `options` (and adding `options` to `stage`) over making `agent` contain things other than agent types.

kzantow: i'm fine with that. it just moves the workspace a little further from the agent config. but it seems more logical than adding it to agent

michaelneale: I am not a user of custom workspace or `ws` myself - but to me it does look like it is in the area of of agent configuration from a UX point of view. Ie "where my stuff runs". However I could ssee the argument that agent config is separate to ws config and it is probably unusual enough to use workspaces like this having it as an option is also fine.

rsandell: Well I could argue that the times when I want to use a custom workspace would be

1. When I need a new workspace on the agent I'm on that is completely different from the one that I get by default, e.g. something that doesn't work by using the `dir` step.
2. I have a workspace on most of my agents where I download common stuff used by different pipelines that I want to share between them.
3. The default workspace path is too long or contains characters that some of my tools can't handle.

For 1 and 2 I feel it would then be strange to have to put the directive in the `agent` section because I actually want to reuse the agent I'm currently on and it would seem like with that syntax that I am declaring to use a new agent.
For 3 it would seem more natural that I specify the custom workspace together with the agent.

michaelneale: fair point @rsandell 

So the ws step won't cater to this because it is too late by then as the agent has already checked thigs out into the default right? (just trying to undersand the motivation of this RFE)

abayer:
@bitwiseman @HRMPW @kzantow Ok, so we've now got `node` as an alias for `label` as an `agent` type. What's more, it'll automatically switch to `node` when it sees `label` in a block context (i.e., `label { label 'foo'; customWorkspace 'some/path' }` but not `label 'foo'`) and translate that to `node { label 'foo'; customWorkspace 'some/path' }` when serializing out to JSON or Groovy. Thoughts?

abayer:
Oh, and I found a nasty bug in `GString` conversion to JSON and fixed that with https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/123/commits/9227b842a0f532b2fe5faf488e8bff1df5d5cb5e.

pipeline-model-definition/src/test/resources/inRelativeCustomWorkspace.groovy
```diff
@@ -0,0 +1,42 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    agent {
+        label {
```
bitwiseman: Should this be `node` now?

bitwiseman: I think you should have separate  commented test for the `label { label '' }` behaviors.

abayer: It shoulda been `node`, yeah, but lemme through a comment in there and a matching JSON to make sure the conversion works.

bitwiseman:
Design looks great to me.  I'm approving this.  I'd like to see at least one test for the label-label behavior with a comment or reference to the RFCs with explanation there.  

:bee:

abayer:
@bitwiseman Added the `label`->`node` test. `label`->`label` staying that way without the additional arguments is already in place with literally everything else that uses `label`. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
```diff
@@ -90,6 +90,23 @@ public class Agent extends MappedClosure<Object,Agent> implements Serializable {
         }
     }
 
+    /**
+     * Needed to handle the combination of describable ordinals *and* Descriptor lookup.
+     * @return The first symbol (in descriptor-ordinal-order searching) found in the map.
+     */
+    private String findSymbol() {
+        String sym = null
+        DeclarativeAgentDescriptor.all().each { d ->
```
rsandell: Well in this case I think a standard java for each loop found be more efficient since you then could break the loop when you have found the thing instead of doing the full combinatorics till the end.

rfc/JENKINS-41118-custom-workspaces.md
rsandell: @michaelneale The checkout could still be valid in a previous stage or on a global level. But to make my scenarios work I would need to specify `reuseNode` as well as `customWorkspace` to get what I need, and you would only know to set that if you are deeply familiar with how this plugin works I think.

jeffchan:
Just curious, will this work for the `post` section as well?

rsandell:
@jeffchan Yes, the global agent includes the global post sections (unless an error happened before we started running on the agent) And each stage's agent covers the post stage as well.

PR #125, opened 2017-02-23
[FIXED JENKINS-41185] Add support for and/not/or of when conditions

abayer:
* JENKINS issue(s):
    * [JENKINS-41185](https://issues.jenkins-ci.org/browse/JENKINS-41185)
* Description:
    * Allows combinations and more complicated conditions. Not sure what additional testing may be needed.
* Documentation changes:
    * To be done - definitely needed, though.
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @bitwiseman


rsandell:
I'm not liking the syntax so much. It is going into the territory of ant/jelly with all these layers of unnatural closures ending up in a big tree with potentially endless nesting, it doesn't seem very "declarative" to me.

Could we experiment with more expression types?
For example how would the following work?
```
when {
  branch "master" or branch "release-*"
}

**Comment by abayer** [comment]
@rsandell Good question. Worth experimenting for sure.

**Comment by abayer** [comment]
...though I should confess that I'm actually weirdly a fan of the nesting. =)

**Comment by rsandell** [comment]
But you promised!!! :'(

**Comment by abayer** [comment]
It's my Lisp background kicking in! I want everything to be an s-expression, dangit.

File: pipeline-model-definition/src/test/resources/whenAnd.groovy [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+    }
+    stages {
+        stage("One") {
+            steps {
+                echo "Hello"
+            }
+        }
+        stage("Two") {
+            when {
+                and {
```
abayer: Yup, will add those tests. And yes, `and` and `or` can have 1..n children of any type (`expression`, `branch`, `environment`, `and`, `or`, `not`).

abayer: Tests added.

abayer: oops, need error cases!

abayer: And, as it turned out, needed error *checking* too. So now we're properly restricting/requiring child count for `and`, `or` and `not`.

bitwiseman:
It would make more sense as it stands to say `all` and `any`.  
```
when {
    all {
        any {
            branch 'master'
            branch 'develop'
        }
        expression { true }
        expression { a == b }    
    }
}
```

abayer:
@kzantow - would love your thoughts here from an editor perspective.

kzantow:
@abayer doable, I've built these conditional-tree UIs before, they take some time to get right from a usability perspective. Maybe we could start simple and just let the user pick the top-level all/any and add conditions. I also like `all` and `any` for this better than and/or.

abayer:
So! This doesn't work with `any` and `all` - specifically because of `any`. `all` blows up due to `@Symbol` fun, but @rsandell's workaround on #116 for `properties` can be used here too. But `any`? Nope. That's ending up calling `public static boolean org.codehaus.groovy.runtime.DefaultGroovyMethods.any(java.lang.Object,groovy.lang.Closure)` whether I like it or not. 

So - @HRMPW, @rsandell, @bitwiseman, @kzantow - how about `anyOf` and `allOf`? Namespace collisions are a thing, so we need _something_ other than `any`/`all`.

kzantow:
`anyOf`/`allOf` seem ok to me. Or maybe `all`/`or` or `all`/`one` or `all`/`either`. I'll stay out of it for the time being, up to you guys...

HRMPW:
I'm not a big fan of it but don't have a better suggestion right now. Feels like a variable name instead of a part of a language.

abayer:
ping @bitwiseman - would like your signoff on the new names.

bitwiseman:
@abayer  - are you sure there's no way to work around the `any`?  That is super annoying, there must be a way to nuke that, but it is probably more pain that it is worth.  Might be worth asking on the Groovy boards.   

I'd almost rather name them `whenAny` and `whenAll'.

Could we then map the current `when` to `whenAll`?   Another non-breaking syntax change, but it would allow some one to do:

```
stage('A') {
    whenAny {
        expression { ... }
        expression { ... }
    }
}
```

Instead of:

```
stage('A') {
    when {
        whenAny {
            expression { ... }
            expression { ... }
        }
    }
}
```


Which would make for a non-breaking syntax change

abayer:
100% sure. `any { ... }` ends up appearing to the Groovy compiler as `any(Closure)` and then things go bad, before I can actually do anything about it. Very annoying, but yeah. `agent any` works because it's not a method call.

I admittedly quite like the `Of` suffix, so I'm gonna take some persuading. =)

abayer:
Actually, it's even weirder than that:
```
java.lang.UnsupportedOperationException: Calling public static boolean org.codehaus.groovy.runtime.DefaultGroovyMethods.any(java.lang.Object,groovy.lang.Closure) on a CPS-transformed closure is not yet supported (JENKINS-26481); encapsulate in a @NonCPS method, or use Java-style loops
 at org.jenkinsci.plugins.workflow.cps.GroovyClassLoaderWhitelist.checkJenkins26481(GroovyClassLoaderWhitelist.java:90)
 at org.jenkinsci.plugins.workflow.cps.GroovyClassLoaderWhitelist.permitsStaticMethod(GroovyClassLoaderWhitelist.java:60)
 at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onMethodCall(SandboxInterceptor.java:92)
 at org.kohsuke.groovy.sandbox.impl.Checker$1.call(Checker.java:149)
 at org.kohsuke.groovy.sandbox.impl.Checker.checkedCall(Checker.java:146)
 at com.cloudbees.groovy.cps.sandbox.SandboxInvoker.methodCall(SandboxInvoker.java:16)
 at WorkflowScript.run(WorkflowScript:38)
 at org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator.resolveClosure(/home/jenkins/workspace/el-definition-plugin_PR-125-XTDTA2CVTQRERUFBHS4JGKZXFSFRJH5D33LNRD3KVR72QRIFN7MA/pipeline-model-definition/target/classes/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy:214)
 at org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator.methodMissing(/home/jenkins/workspace/el-definition-plugin_PR-125-XTDTA2CVTQRERUFBHS4JGKZXFSFRJH5D33LNRD3KVR72QRIFN7MA/pipeline-model-definition/target/classes/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy:149)
 at WorkflowScript.run(WorkflowScript:37)
 at org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator.resolveClosure(/home/jenkins/workspace/el-definition-plugin_PR-125-XTDTA2CVTQRERUFBHS4JGKZXFSFRJH5D33LNRD3KVR72QRIFN7MA/pipeline-model-definition/target/classes/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy:214)
 at org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator.methodMissing(/home/jenkins/workspace/el-definition-plugin_PR-125-XTDTA2CVTQRERUFBHS4JGKZXFSFRJH5D33LNRD3KVR72QRIFN7MA/pipeline-model-definition/target/classes/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy:128)
 at WorkflowScript.run(WorkflowScript:36)
 at org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator.resolveClosure(/home/jenkins/workspace/el-definition-plugin_PR-125-XTDTA2CVTQRERUFBHS4JGKZXFSFRJH5D33LNRD3KVR72QRIFN7MA/pipeline-model-definition/target/classes/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy:214)
 at org.jenkinsci.plugins.pipeline.modeldefinition.ClosureModelTranslator.methodMissing(/home/jenkins/workspace/el-definition-plugin_PR-125-XTDTA2CVTQRERUFBHS4JGKZXFSFRJH5D33LNRD3KVR72QRIFN7MA/pipeline-model-definition/target/classes/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy:166)
 at WorkflowScript.run(WorkflowScript:30)
 at org.jenkinsci.plugins.pipeline.modeldefinition.ModelInterpreter.call(/home/jenkins/workspace/el-definition-plugin_PR-125-XTDTA2CVTQRERUFBHS4JGKZXFSFRJH5D33LNRD3KVR72QRIFN7MA/pipeline-model-definition/target/classes/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy:58)
 at WorkflowScript.run(WorkflowScript:25)
```
That's the actual stacktrace. From what I can tell, it *might* be possible to cheat this by adding a `any` method to `StageConditionalTranslator`, but that's some truly nasty hardcoding that I really, really, really don't want to keep doing - `agent any` and `agent none` were the last times I was willing to do something like that. I'm definitely not willing to do hardcoding for one implementation of an extension point, 'cos then someone might add another implementation that collides with another built-in method, needing another hardcoding, etc...

Oh, and the error at the top of the stacktrace is amusing - if it weren't for the special logic in `workflow-cps` to proactively barf on use of Groovy's `.each`, `.any` and friends due to serialization issues, I legitimately don't know how this would have behaved at *all*.

abayer:
oh, and 👎 to the `when` aliasing. That would require having the AST/schema/etc processing in `pipeline-model-api` know about a specific implementation of `DeclarativeStageConditional`, which would be complicated and a bad idea.

bitwiseman:
Groovy is so groovy We could overload `and/or`, but not `any/all`. 

Alternatives are: 
* `all/some`
* `allOf/anyOf`
* `all/anyOf`
* `whenAll/whenAny`
* `all/atleastOne`
* `all/oneOrMore
* `all/none` (with `not none` being the same as `any` 😱 )

The list above is in the order of my preference, but I'm okay with any of them... or `anyOf` them.  
 


Given the choices, I'll take `anyOf/allOf`.  Most of us don't love it, but none of us hate it.

abayer:
I'm gonna go with `anyOf`/`allOf`, since that's what we've got currently and I like it better than `all`/`some` or any of the other options. =)

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTWhenCondition.java
```diff
@@ -0,0 +1,178 @@
...
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author Andrew Bayer
+ */
+@ToString(includeSuper = true, includeSuperProperties = true)
```
rsandell: Does this work on Java objects?
Besides you already have an implemented `toString()` below.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -402,4 +407,75 @@ public class Utils {
         }
         return true
     }
+
+    public static boolean whenConditionDescriptorFound(String name) {
```
rsandell: missing javadoc

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/DeclarativeStageConditionalDescriptor.java
```diff
@@ -43,6 +43,14 @@
         extends WithScriptDescriptor<S> {
 
     /**
+     * How many nested conditions are allowed. -1 for unlimited, 0 for none, anything greater than 0 for
+     * requiring exactly that many nested conditions.
+     */
+    public int allowedNested() {
```
rsandell: Would prefer this followed the bean standard

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -193,32 +193,53 @@ class ModelValidatorImpl implements ModelValidator {
...
 
-            Descriptor desc = DeclarativeStageConditionalDescriptor.byName(condition.name)
+            DeclarativeStageConditionalDescriptor desc = DeclarativeStageConditionalDescriptor.byName(condition.name)
 
-            return validateStep(condition, model, desc)
+            if (desc.allowedNested() != 0) {
+                if (condition.args != null) {
+                    errorCollector.error(condition, Messages.ModelValidatorImpl_NestedWhenNoArgs(condition.name))
+                    valid = false
+                } else if (desc.allowedNested() == -1) {
```
rsandell: Risk of boundary bug, be safe and check `< 0` instead.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -411,13 +411,14 @@ public class ModelInterpreter implements Serializable {
         if (when == null) {
             return true
         } else {
+            boolean ret = true
             for (int i = 0; i < when.conditions.size(); i++) {
                 DeclarativeStageConditional c = when.conditions.get(i)
                 if (!c.getScript(script).evaluate()) {
-                    return false
+                    ret = false
```
rsandell: This seems to be an unneeded change? Why would you need to iterate through them all?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/StageConditionalTranslator.groovy
```diff
@@ -62,18 +66,24 @@ public class StageConditionalTranslator implements MethodMissingWrapper, Seriali
...
 
         if (Utils.instanceOfWrapper(Closure.class, argVal)) {
-            argVal = createStepsBlock(argVal)
-        }
-
-        DeclarativeStageConditionalDescriptor descriptor = DeclarativeStageConditionalDescriptor.byName(s)
-        if (descriptor == null) {
-            throw new NoSuchMethodError(Messages.ModelValidator_ModelASTWhen_unknown(s, DeclarativeStageConditionalDescriptor.allNames().join(", ")))
-        }
+            if (Utils.nestedWhenCondition(s)) {
```
rsandell: Why the double, no triple, lookup?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/StageConditionalTranslator.groovy
```diff
@@ -62,18 +66,24 @@ public class StageConditionalTranslator implements MethodMissingWrapper, Seriali
...
+                def st = new StageConditionalTranslator(script)
+                resolveClosure(argVal, st)
+                List<DeclarativeStageConditional> conds = st.toWhen()?.conditions
+                if (Utils.takesWhenConditionList(s)) {
```
rsandell: And again

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/StageConditionalTranslator.groovy
```diff
@@ -62,18 +66,24 @@ public class StageConditionalTranslator implements MethodMissingWrapper, Seriali
...
+                } else {
+                    argVal = null
+                }
+            } else {
+                argVal = createStepsBlock(argVal)
+            }
 
-        if (argVal != null) {
-            retVal = script."${s}"(argVal)
+            retVal = Utils.getDescribable(s, DeclarativeStageConditional.class, argVal)
```
rsandell: ...and again

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AllOfConditionalScript.groovy
```diff
@@ -0,0 +1,51 @@
...
+    public AllOfConditionalScript(CpsScript s, AllOfConditional c) {
+        super(s, c)
+    }
+
+    @Override
+    public boolean evaluate() {
+        List<DeclarativeStageConditional<? extends DeclarativeStageConditional>> nested = describable.nested
+        for (int i = 0; i < nested.size(); i++) {
+            DeclarativeStageConditional n = nested.get(i)
+            DeclarativeStageConditionalScript s = (DeclarativeStageConditionalScript)n?.getScript(script)
```
rsandell: Why force all conditionals to implement a script?! seems very tedious and unnecessary. Would be simpler to just have an `evaluate` method in the conditional that by default does `getScript(script).evaluate()`.

pipeline-model-definition/src/test/resources/whenNestedCombinations.groovy
```diff
@@ -0,0 +1,75 @@
...
+            steps {
+                echo "Third stage meets condition"
+            }
+        }
+        stage("Four") {
+            when {
+                anyOf {
+                    allOf {
+                        not {
+                            expression { false }
```
rsandell: must it be so? Since `not` only allows one conditional I would hope that you can write `not expression { false }`. Some of the parsing above seems to indicate it being so.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/DeclarativeStageConditionalDescriptor.java
rsandell: e.g. `getAllowedNestedCount()` or similar.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTWhenCondition.java
abayer: Good question! Probably not, so I'm cleaning it up in all the AST classes.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/DeclarativeStageConditionalDescriptor.java
abayer: Works for me.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
abayer: Fixed.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Good point.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/StageConditionalTranslator.groovy
abayer: ...'cos I didn't think that one through. =)

abayer: Wait, check that, found a reason - `DescribableStageConditionalDescriptor` isn't serializable, and as far as I can tell, most (if not all) descriptors aren't serializable so it'd be a bit weird to make this one so. As a result, we can't actually have an instance of a `DescribableStageConditionalDescriptor` in CPS code. Lemme think about caching instead.

pipeline-model-definition/src/test/resources/whenNestedCombinations.groovy
abayer: Fairly sure it wouldn't work currently, but I could be wrong. Will test.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AllOfConditionalScript.groovy
abayer: Let's see how this works...

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/StageConditionalTranslator.groovy
abayer: ok, caching now.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AllOfConditionalScript.groovy
abayer: Argh. Due to CPS fun, we do need to do it all in scripts. When I changed `NotConditional` and friends to not run their own scripts but instead call `evaluate(script)` on their nested scripts, it only calls the first nested condition and resolves the parent conditional as true. I assume this is due to the flow control by exception whackiness. Sigh. Reverting back to everything needing a script.

rsandell: Hmm, that's not the behavior I experienced when I used a similar approach on credentials and properties.

abayer: I think it's the multiple calls that cause the problem, but I could be wrong.

pipeline-model-definition/src/test/resources/whenNestedCombinations.groovy
abayer: Ok, so I'm not gonna support that - why? Because at the very least I can't come up with a way to know that we can serialize to `not expression { false }`, since the AST model doesn't know anything about the potential `when` conditions and which ones can take one child vs multiple children... I want the consistency in round-tripping more than I want a shortcut.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AllOfConditional.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class AllOfConditional extends DeclarativeStageConditional<AllOfConditional> {
+    private final List<DeclarativeStageConditional<? extends DeclarativeStageConditional>> nested;
+
+    @DataBoundConstructor
+    public AllOfConditional(List<DeclarativeStageConditional<? extends DeclarativeStageConditional>> nested) {
+        this.nested = nested;
+    }
+
+    public List<DeclarativeStageConditional<? extends DeclarativeStageConditional>> getNested() {
```
rsandell: Why call them `children` in the model and then `nested` here?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AllOfConditional.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditional;
+import org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditionalDescriptor;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import java.util.List;
+
+/**
+ * Match all of a list of stage conditions
+ */
+public class AllOfConditional extends DeclarativeStageConditional<AllOfConditional> {
```
rsandell: Looks like `AllOfConditional`, `AnyOfConditional` and `NotConditional` could just inherit from the same abstract base class to avoid declaring the _"children"_ multiple times.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTWhenContent.java
```diff
@@ -0,0 +1,41 @@
...
+package org.jenkinsci.plugins.pipeline.modeldefinition.ast;
+
+import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator;
+
+/**
+ *
+ * @author Andrew Bayer
+ */
+public interface ModelASTWhenContent {
+    String toGroovy();
```
abayer: Yeah, I can do that.

pipeline-model-api/src/main/resources/ast-schema.json
```diff
@@ -348,7 +376,7 @@
                 "$ref": "#/definitions/step"
```
abayer: That...is weird legacy syntax. I should undo it.

pipeline-model-definition/src/test/resources/whenNestedCombinations.groovy
abayer: I don't think it's worth the pain in terms of more complicated parsing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AllOfConditional.java
abayer: Because oops. Fixing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AllOfConditional.java
abayer: Fair point.

abayer: `NotConditional`, since it only has one child, didn't make sense, but the other two did, so - done.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTWhen.java
```diff
@@ -37,60 +37,50 @@
...
     }
 
     @Override
     public Object toJSON() {
-        final JSONArray a = new JSONArray();
-        for (ModelASTWhenContent c : conditions) {
-            a.add(c.toJSON());
-        }
-        return new JSONObject().accumulate("conditions", a);
+        return new JSONObject().accumulate("condition", condition.toJSON());
```
rsandell: I the editor prepared for this schema change?

abayer: The editor doesn't yet have `when` support in the first place, I believe - @kzantow, can you verify?

abayer: Got verification from Keith. woot.

leonhartX:
please, when you guys make changes like this, at least update the doc!
the v1.1 (with this PR) just broke all my jenkinsfile like this (quote from wiki)
```
      // Conditional execution of this stage - only run this stage if the when condition is true.
      when {
        // One and only one condition is allowed.

        // Only run if the branch matches this Ant-style pattern
        branch "master"

        // Only run if the environment contains this given variable name with this given value
        environment name: "FOO", value: "notBar"

        // Only run if this Scripted Pipeline expression doesn't return false or null
        expression {
          echo "You can run any Pipeline steps in here"
          return "foo" == "bar"
        }
      }
```

abayer:
Argh, I thought we were blocking multiple `when` conditions already. Crap. Sorry.

leonhartX:
@abayer hi, this may not related to this PR, but after fix the when condition, I got this error:
```
java.lang.NoClassDefFoundError: Unable to load class org.jenkinsci.plugins.pipeline.modeldefinition.model.Environment due to missing dependency java/util/function/Function
```

all plugin is up-to-date, do you have any idea about this?
thanks.

abayer:
Craaaaap. I must have inadvertently built it with Java 8. I'll get a rebuilt 1.1.1 out the door right now, should be in the update center in a couple hours.

abayer:
@leonhartX So 1.1.1 is out the door and should solve the `java.util.function.Function` problem. I've opened https://issues.jenkins-ci.org/browse/JENKINS-42762 for reverting to the previous allowing multiple {{when}} conditions as well.

PR #128, opened 2017-03-01
[JENKINS-38110] Add a libraries section

abayer:
* JENKINS issue(s):
    * [JENKINS-38110](https://issues.jenkins-ci.org/browse/JENKINS-38110)
* Description:
    * Initial work on adding a `libraries` section to Declarative. Current syntax looks like:
```groovy
libraries {
    lib('foo@1.2.3')
    lib('bar')
}
```
Yes, I'd prefer not to have that pointless `lib` there, but the alternative is something like `libraries 'foo@1.2.3', 'bar'` and as we may remember from `agent` back in the day, I do not care for that syntax at _all_. Plus, we may end up wanting to include the ability to specify a `LibraryRetriever` (see [LibraryRetriever.java in workflow-cps-global-lib](https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRetriever.java) for details), in which case going to `lib(identifier:'foo@1.2.3', retriever:[$class:...])` is a much better option than anything else I could find.
* Documentation changes:
    * Oh my yes will there need to be doc changes. But not 'til this actually is much closer to done.
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @bitwiseman 
    * @HRMPW 
    * @kzantow @michaelneale @i386 (particularly for editor/BO concerns)

### DIFF
--- a/pipeline-model-api/src/main/resources/ast-schema.json
+++ b/pipeline-model-api/src/main/resources/ast-schema.json
@@ -156,6 +156,14 @@
         }
       ]
     },
+    "libraries": {
+      "description": "One or more shared library identifiers to load",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "minItems": 1
+    },
     "options": {
       "description": "One or more options (including job properties, wrappers, and options specific to Declarative Pipelines)",
       "type": "object",
@@ -440,6 +448,9 @@
         },
         "parameters": {
           "$ref": "#/definitions/parameters"
+        },
+        "libraries": {
+          "$ref": "#/definitions/libraries"
         }
       },
       "required": [
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTranslator.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTranslator.groovy
@@ -29,7 +29,7 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.model.Libraries
 import org.jenkinsci.plugins.workflow.cps.CpsScript
 
 /**
- * Translates a closure containing a sequence of "library('string')" calls into an instance of {@link Libraries}.
+ * Translates a closure containing a sequence of "lib('string')" calls into an instance of {@link Libraries}.
  *
  * @author Andrew Bayer
  */
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTranslator.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTranslator.groovy
@@ -1,0 +1,52 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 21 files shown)


abayer:
Note that this is very preliminary at this point - like, it doesn't
actually call the library step yet, since we don't yet have a release
of workflow-cps-global-lib-plugin with that included to depend on. We
also don't have parsing tests or any of that jazz. But that's ok.

kzantow:
Could the `LibraryRetriever` have some special metadata that could list available libraries in some cases? I guess we'd just allow a random string for now or am I missing something?

HRMPW:
@kzantow `LibraryRetriever` lets you load a library from an arbitrary source.  `lib` itself could possibly know the list of available libraries based on the configuration in the folder or global but I don't think you can pre-populate `LibraryRetriever`

abayer:
Yeah, the `LibraryRetriever` could specify things like credentials, an alternate repo (since it'd be GitHub by default), etc... I'm also trying to figure out what to do about validation of the library itself - it looks like I'd need the job to be able to validate the identifier string if folder context needs to be taken into account.

abayer:
@HRMPW [LibraryStep](https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/14) allows an optional `LibraryRetriever` argument. Assuming we do add support for it, the syntax would probably be `lib("foo@1.2.3")` or `lib("foo@1.2.3", modernSCM([$class: 'GitSCMSource', credentialsId: '', excludes: '', id: 'id', ignoreOnPushNotifications: true, includes: '*', remote: 'https://nowhere.net/']))`. Ideally, we'll get `@Symbol`s and proper `DataBoundConstructor` setups on the `SCMSource` implementations to make that syntax smoother, though.

kzantow:
Sorry, was conflating things... my goal for the editor is - as much as possible - to give the user choices for things to minimize errors, so as much info as we can pull from jenkins, the better... e.g. which `LibraryRetriever` types are present, possibly what they resolve, etc., and any known entries for the `lib` itself.

abayer:
@kzantow Yup, that we can do to at least some extent - leveraging [LibraryStep's auto-complete logic](https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/14/files#diff-55519320b71a7601a2be2194202b0ef5R119)

jglick:
> we don't yet have a release of workflow-cps-global-lib-plugin with that included to depend on

Try: `2.7-20170301.201004-1`

abayer:
Thanks, @jglick  - done.

abayer:
Moving out of work-in-progress - decided to leave out `LibraryRetriever` for the initial release and wait to see if there's demand.

rsandell:
If we go with adding folder context to the validation we should also revisit the validation of the credentials function in environment so we can fail there as well fast if the user has entered an non existent credential id.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTranslator.groovy
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+
+package org.jenkinsci.plugins.pipeline.modeldefinition
+
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.Libraries
+import org.jenkinsci.plugins.workflow.cps.CpsScript
+
+/**
+ * Translates a closure containing a sequence of "library('string')" calls into an instance of {@link Libraries}.
```
rsandell: Wasn't it `lib('string')`?

pom.xml
```diff
@@ -148,7 +148,7 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-cps-global-lib</artifactId>
-        <version>2.5</version>
+        <version>2.7-20170301.201004-1</version> <!-- TODO: Update to 2.7 once released -->
```
jglick: Released.

pipeline-model-api/src/main/resources/ast-schema.json
```diff
@@ -156,6 +156,14 @@
         }
       ]
     },
+    "libraries": {
+      "description": "One or more shared library identifiers to load",
+      "type": "array",
+      "items": {
+        "type": "string"
```
abayer: By supporting both the current array-of-strings and a new format for `LibraryRetriever` - `anyOf` will let us do that easily enough.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTranslator.groovy
abayer: Fixed!

pom.xml
abayer: Updated.

PR #130, opened 2017-03-03
[FIXED JENKINS-42473] Only do parsing if source is WorkflowScript

abayer:
* JENKINS issue(s):
    * [JENKINS-42473](https://issues.jenkins-ci.org/browse/JENKINS-42473)
* Description:
    * Deals with cases where there are multiple SourceUnits in the CompilationUnit. This shows up with more recent git plugin versions.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @imod

### DIFF
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -119,7 +119,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>git</artifactId>
-      <optional>true</optional>
+      <scope>test</scope>
       <exclusions>
         <exclusion>
           <groupId>org.apache.httpcomponents</groupId>
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -119,7 +119,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>git</artifactId>
-      <scope>test</scope>
+      <optional>true</optional>
       <exclusions>
         <exclusion>
           <groupId>org.apache.httpcomponents</groupId>
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
@@ -26,8 +26,7 @@ package org.jenkinsci.plugins.pipeline.modeldefinition.parser
 import com.cloudbees.groovy.cps.NonCPS
 import com.github.fge.jsonschema.util.JsonLoader
 import org.jenkinsci.plugins.pipeline.modeldefinition.shaded.com.fasterxml.jackson.databind.JsonNode
-import org.jenkinsci.plugins.pipeline.modeldefinition.shaded.com.fasterxml.jackson.databind.ObjectMapper
-import org.jenkinsci.plugins.pipeline.modeldefinition.shaded.com.fasterxml.jackson.datatype.jsonorg.JsonOrgModule
+import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 import com.github.fge.jsonschema.exceptions.ProcessingException
 import com.github.fge.jsonschema.main.JsonSchema
 import com.github.fge.jsonschema.report.ProcessingReport
@@ -35,7 +34,6 @@ import com.github.fge.jsonschema.tree.JsonTree
 import com.github.fge.jsonschema.tree.SimpleJsonTree
 import jenkins.model.Jenkins
 import net.sf.json.JSONObject
-import org.apache.commons.lang.reflect.FieldUtils
 import org.codehaus.groovy.control.CompilationFailedException
 import org.codehaus.groovy.control.CompilationUnit
 import org.codehaus.groovy.control.CompilerConfiguration
@@ -45,9 +43,9 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.ASTSchema
 import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTPipelineDef
 import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStep
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution
-import org.jenkinsci.plugins.workflow.cps.CpsScript
 import org.jenkinsci.plugins.workflow.cps.CpsThread
 import org.jenkinsci.plugins.workflow.cps.GroovyShellDecorator
+import org.jenkinsci.plugins.workflow.job.WorkflowRun
 
 import java.security.CodeSource
 import java.security.cert.Certificate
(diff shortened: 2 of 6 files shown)


pipeline-model-definition/pom.xml
```diff
@@ -120,7 +120,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>git</artifactId>
-      <scope>test</scope>
+      <optional>true</optional>
```
jglick: I hope this was an accident?

abayer: Nope, sadly. The test wouldn't fail with the git plugin just as a test dependency.

jglick: Uh, why?

abayer: I don't know? My guess is there's something else in the dependency tree that's getting a newer version when git is a non-test dependency.

jglick:
```sh
mvn dependency:tree > /tmp/tree-opt.txt
perl -p -i -e 's{<optional>true</optional>}{<scope>test</scope>}' pom.xml
mvn dependency:tree > /tmp/tree-test.txt
diff /tmp/tree-*.txt
```

abayer: Already tried that - only difference was scope. I'm checking the actual hpi files used in the tests now...

abayer: Ok, now my head is exploding. `target/test-classes/test-dependencies` is identical whether I'm running `mvn clean test -Dtest="BasicModelDefTest#folderLibraryParsing"` with `<optional>true</optional>` or `<scope>test</scope>`, but the test is consistently failing with `<optional>true</optional>` (with the fix in `Converter.groovy` commented out) and passing with `<scope>test</scope>`.

I am very, very mystified.

abayer: For the record, here's the `dependency:tree` diff:

```
49,56c49,56
< [INFO] |  +- org.jenkins-ci.plugins:git-client:jar:1.19.7:compile
< [INFO] |  |  +- org.eclipse.jgit:org.eclipse.jgit:jar:3.7.1.201504261725-r:compile
< [INFO] |  |  |  +- com.jcraft:jsch:jar:0.1.50:compile
< [INFO] |  |  |  \- com.googlecode.javaewah:JavaEWAH:jar:0.7.9:compile
< [INFO] |  |  +- org.eclipse.jgit:org.eclipse.jgit.http.server:jar:3.7.1.201504261725-r:compile
< [INFO] |  |  +- org.eclipse.jgit:org.eclipse.jgit.java7:jar:3.7.1.201504261725-r:compile
< [INFO] |  |  \- org.apache.httpcomponents:httpclient:jar:4.4.1:compile
< [INFO] |  |     \- org.apache.httpcomponents:httpcore:jar:4.4.1:compile
---
> [INFO] |  +- org.jenkins-ci.plugins:git-client:jar:1.19.7:test
> [INFO] |  |  +- org.eclipse.jgit:org.eclipse.jgit:jar:3.7.1.201504261725-r:test
> [INFO] |  |  |  +- com.jcraft:jsch:jar:0.1.50:test
> [INFO] |  |  |  \- com.googlecode.javaewah:JavaEWAH:jar:0.7.9:test
> [INFO] |  |  +- org.eclipse.jgit:org.eclipse.jgit.http.server:jar:3.7.1.201504261725-r:test
> [INFO] |  |  +- org.eclipse.jgit:org.eclipse.jgit.java7:jar:3.7.1.201504261725-r:test
> [INFO] |  |  \- org.apache.httpcomponents:httpclient:jar:4.4.1:test
> [INFO] |  |     \- org.apache.httpcomponents:httpcore:jar:4.4.1:test
58,63c58,63
< [INFO] +- org.jenkins-ci.plugins:git:jar:3.0.1:compile
< [INFO] |  +- joda-time:joda-time:jar:2.9.5:compile
< [INFO] |  +- com.infradna.tool:bridge-method-annotation:jar:1.14:compile
< [INFO] |  +- org.jenkins-ci.plugins:ssh-credentials:jar:1.12:compile
< [INFO] |  \- org.jenkins-ci.plugins:matrix-project:jar:1.7.1:compile
< [INFO] |     \- org.jenkins-ci.plugins:junit:jar:1.2:compile
---
> [INFO] +- org.jenkins-ci.plugins:git:jar:3.0.1:test
> [INFO] |  +- joda-time:joda-time:jar:2.9.5:test
> [INFO] |  +- com.infradna.tool:bridge-method-annotation:jar:1.14:provided
> [INFO] |  +- org.jenkins-ci.plugins:ssh-credentials:jar:1.12:test
> [INFO] |  \- org.jenkins-ci.plugins:matrix-project:jar:1.7.1:test
> [INFO] |     \- org.jenkins-ci.plugins:junit:jar:1.2:test
184c184
< [INFO] |  +- commons-codec:commons-codec:jar:1.8:compile
---
> [INFO] |  +- commons-codec:commons-codec:jar:1.8:provided
```

abayer:
Welp, I can't get it to be consistent, but at least I've gotten it back to the old git plugin version and scope. Again, with the fix commented out, it fails a fairly large percentage of the time, but not *every* time and I have no idea why. Ah well.

abayer:
fwiw, the root cause seems to be that *sometimes* the library files are in the `CompilationUnit`s `sources` (due to being in the `GroovyClassLoader`'s list of URLs) but not always. I'm suspecting it's something about the timing of the libs directory being populated in the workspace but am not sure. Also fwiw, the root cause is not, in my opinion, a bug - it's purely a bug on the Declarative side, where we had been working under the unsafe assumption that there'd never be any other `SourceUnit`s getting processed when we do a conversion outside of compilation time. Hence the fix.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
```diff
@@ -194,13 +196,13 @@ public class Converter {
...
-     * Converts the raw script from a {@link CpsScript} into {@link ModelASTPipelineDef}
+     * Converts the raw script from a {@link WorkflowRun} into {@link ModelASTPipelineDef}
      *
-     * @param cpsScript The {@link CpsScript} to pull from.
+     * @param run The {@link WorkflowRun} to pull from.
      * @return A parsed and validated {@link ModelASTPipelineDef}
      */
-    public static ModelASTPipelineDef parseFromCpsScript(CpsScript cpsScript) {
-        CpsFlowExecution execution = (CpsFlowExecution) FieldUtils.readField(cpsScript, "execution", true)
+    public static ModelASTPipelineDef parseFromWorkflowRun(WorkflowRun run) {
```
rsandell: Still makes my nose wrinkle when  public method changes name.

jglick:
> I'm suspecting it's something about the timing of the libs directory being populated in the workspace but am not sure.

Better track that down.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
```diff
@@ -106,7 +104,7 @@ public class Converter {
             makeCompilerConfiguration(),
             new CodeSource(src, new Certificate[0]),
             getCompilationClassLoader());
-        cu.addSource(src);
+        cu.addSource(src)
```
jglick: Why?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
jglick: If anyone is seriously treating this as an API, they deserve to have their plugin broken.

Really I would like to introduce a system like OSGi has (NetBeans modules, Java 9 modules, …) whereby we can just declare _all_ types in the plugin to be off limits unless and until explicitly exported to the world.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/Converter.groovy
abayer: Oops, was messing around with that at one point and didn't replace the semicolon.

PR #131, opened 2017-03-06
[FIXED JENKINS-42498] Dehydrate GStrings in the runtime model

abayer:
* JENKINS issue(s):
    * [JENKINS-42498](https://issues.jenkins-ci.org/browse/JENKINS-42498)
* Description:
    * This prevents non-serializable errors when there are attempts to serialize via `XStream`, which rejects `GStringImpl`. There's still a question as to whether `GStringImpl` should be rejected by `hudson.remoting.ClassFilter.DEFAULT`, but that's a separate issue. We fix the problem by "dehydrating" `GString`s into their `strings` and `values`, which we can then use to "rehydrate" them into `GStringImpl`s at the right time.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @jglick (would like your thoughts)

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/SerializableGString.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/SerializableGString.groovy
@@ -1,0 +1,62 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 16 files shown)


jglick:
> There's still a question as to whether `GStringImpl` should be rejected by `hudson.remoting.ClassFilter.DEFAULT`, but that's a separate issue.

Have you filed a `remoting` PR to allow it?

jglick:
This whole PR seems crazy. Set a breakpoint

    at org.jenkinsci.plugins.workflow.support.pickles.XStreamPickle.<init>(XStreamPickle.java:46)

and figure out what `Describable` is being stored that has such an instance serial form. This is your problem: `Describable` should be reserved for things configured in a Jenkins form. It should not normally crop up in program state. Certainly no object created _by_ the script.

abayer:
So I know where it's happening - `ExpressionConditional`, which contains a `StepsBlock`, which itself contains the `Closure` to actually evaluate at runtime. I'm using `Describable` for `DeclarativeStageConditional` (and a bunch of other things) so that we get `UninstantiatedDescribable` support, etc for them, and I'm storing the `Closure` there because there's no other way to take a `Closure` specified in the `Jenkinsfile` and execute it later, at least not that I can figure out. I'll experiment.

rsandell:
You don't seem to have the actual script in the ticket that caused the issue?

PR #132, opened 2017-03-06
[FIXED JENKINS-42498] Stop evaluating when conditions at runtime

abayer:
* JENKINS issue(s):
    * [JENKINS-42498](https://issues.jenkins-ci.org/browse/JENKINS-42498)
* Description:
    * Pull the when conditions in by translating them from the AST model, which does require some evaluation of strings here and there, but avoids having any `Closure`s in the `Describable`s. Note that this is on top of #125 - relevant commits start at 76a4350. If this PR looks more appropriate than #131, I'll close #131.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @jglick


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ClosureModelTranslator.groovy
```diff
@@ -46,6 +46,8 @@ import static org.jenkinsci.plugins.pipeline.modeldefinition.Utils.createStepsBl
  * @author Andrew Bayer
  */
 public class ClosureModelTranslator implements MethodMissingWrapper, Serializable {
+    private static final List<Class> UNTRANSLATED_CLASSES = [StageConditionals.class]
```
abayer: Making this a `List` to check against since we're doing this again for `Environment` over in #110, and I imagine there'll be more in the future as I gradually kill off the translators.

jglick:
Tried to load the [effective diff](https://github.com/abayer/pipeline-model-definition-plugin/compare/jenkins-41185...jenkins-42498-mk2) but it is showing commits from #125, which it should not have been; any idea?

jglick:
Maybe just get upstream reviewed & merged, then `git merge master` here and GitHub will show the incremental diff automatically.

abayer:
@jglick [effective diff](https://github.com/abayer/pipeline-model-definition-plugin/compare/jenkins-41185...jenkins-42498-mk2) fixed now, but yeah, I want to get #125 merged. I would appreciate your thoughts in the meantime, though.

jglick:
I am not sure I really understand what it is I am looking at, but it seems that for something like

```groovy
environment {
  TAG = "$JOB_NAME.$BUILD_NUMBER"
}
```

previously you would have been passing around a `Closure` and `call`ing it, and now you are passing around a `String` and `evaluate`ing it?

In an actual programming language this would be a huge red flag—you are throwing away proper handling of lexical scope, and generally setting the clock back to 1961. But in Declarative I guess it is fine because by the restrictive syntactic rules imposed by the verifier you would not be able to use anything bound into a closure anyway.

abayer:
@jglick Yeah, it's not my first choice, but in this particular case, either we're storing a `Closure` in a `Describable` and hitting the serialization problem, or we're storing a `String` and `evaluate`ing it. I think the tradeoffs come down in favor of the `String`/`evaluate` option, but I definitely want your input.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -200,19 +211,96 @@ public class Utils {
...
+
+        return c
+    }
+
+    /**
+     * Takes a string and makes sure it starts/ends with double quotes so that it can be evaluated correctly.
+     *
+     * @param s The original string
+     * @return Either the original string, if it already starts/ends with double quotes, or the original string
+     * prepended/appended with double quotes.
```
jglick: Uh, does this not need to do some kind of escaping?

abayer: Hrm. Honestly not sure. Lemme think on it.

abayer: So actually, I don't think it does. We're just taking the string from the original Jenkinsfile source, in effect.

PR #133, opened 2017-03-09
[JENKINS-42640, JENKINS-42551] Fix some JSON-specific validation problems

abayer:
* JENKINS issue(s):
    * [JENKINS-42640](https://issues.jenkins-ci.org/browse/JENKINS-42640)
    * [JENKINS-42551](https://issues.jenkins-ci.org/browse/JENKINS-42551)
* Description:
    * re JENKINS-42640: In validating the options contents, we didn't actually look for a StepDescriptor, just a symbol-driven Descriptor. Which won't work for steps! So we never actually validated the contents of wrappers in options. Besides that, we needed to reject converting to int from String/Boolean - that gets through ScriptByteCodeAdapter.castToType, but not runtime.
    * re JENKINS-42551: So now the JSON string "hello\\'" will cause a parse-time error. I also tweaked JSON validation endpoint to also convert to Jenkinsfile and validate *that*, so as to be sure we've actually got JSON that can be a valid Jenkinsfile, even if it has some kind of gotcha in it like the one fixed here.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @kzantow


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/JSONParser.groovy
```diff
@@ -419,19 +420,26 @@ class JSONParser implements Parser {
...
-                return ModelASTValue.fromConstant(o.node.get("value").textValue(), o)
+                val = ModelASTValue.fromConstant(o.node.get("value").textValue(), o)
             }
         } else {
-            return ModelASTValue.fromGString(o.node.get("value").textValue(), o)
+            val = ModelASTValue.fromGString(o.node.get("value").textValue(), o)
         }
+
+        String valGroovy = val?.toGroovy()
+        if (valGroovy != null && valGroovy != StringEscapeUtils.escapeJava(valGroovy)) {
```
abayer: It's a fun little trick - it's basically saying "is this a safe Java string?" by making sure it is equal to an escaped version of itself. If it's not equal to its escaped equivalent, then it's not a safe Java string. Tada!

kzantow: Ah, I was confusing this as a fix for something else, carry on :)

abayer: Well, turns out that didn't actually work the way I thought it did anyway. New commit(s) incoming.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
```diff
@@ -112,11 +112,15 @@ public synchronized StepDescriptor lookupStepDescriptor(String n) {
         return describableMap.get(n);
     }
 
-    public synchronized Descriptor<? extends Describable> lookupStepOrFunction(String name) {
+    public synchronized Descriptor<? extends Describable> lookupStepFirstThenFunction(String name) {
```
rsandell: we should put `@Restricted(NoExternalUse.class)` on things so that I can feel more calm when you change method signatures like this. I believe you can put in on classes as well now.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
```diff
@@ -112,11 +112,15 @@ public synchronized StepDescriptor lookupStepDescriptor(String n) {
...
         return lookupStepDescriptor(name) != null ? lookupStepDescriptor(name) : lookupFunction(name);
     }
 
-    public synchronized DescribableModel<? extends Describable> modelForStepOrFunction(String name) {
+    public synchronized Descriptor<? extends Describable> lookupFunctionFirstThenStep(String name) {
```
rsandell: Just realized something not exactly related to this, but we probably need to invalidate the cache when a new plugin is installed or someone configures a new global library.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -515,26 +511,26 @@ class ModelValidatorImpl implements ModelValidator {
...
 
     private boolean validateParameterType(ModelASTValue v, Class erasedType, ModelASTKey k = null) {
         if (v.isLiteral()) {
             try {
-                // Converting from boolean or int to string at runtime doesn't work, but does pass castToType. So.
-                if (erasedType.equals(String.class)
-                    && (v.value instanceof Integer || v.value instanceof Boolean)) {
+                // Converting amongst boolean, string and int at runtime doesn't work, but does pass castToType. So.
+                if ((erasedType.equals(String.class) && (v.value instanceof Integer || v.value instanceof Boolean)) ||
+                    (erasedType.equals(int.class) && (v.value instanceof String || v.value instanceof Boolean))) {
```
rsandell: Isn't `(erasedType.equals(int.class)` just an unneeded autoboxing?

rsandell: e.g. `(erasedType.equals(Integer.class)` would be the same but without the autoboxing?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
kzantow: MEGA 👍  to that, if we're not doing so already

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
abayer: So I honestly don't know, but yeah, looks like `Integer.class` works.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
abayer: Let's open a separate JIRA for that?

rsandell: :+1:

rsandell:
:bee:

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
abayer: Actually, just pushed a draft change adding expiration and fun stuffs.

abayer:
So the caching didn't work right, and it *does* need to be `int.class`, so...rolling back both those changes. =)

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DescriptorLookupCache.java
```diff
@@ -63,56 +64,71 @@ public DescriptorLookupCache() {
...
     public synchronized DescribableModel<? extends Step> modelForStep(String n) {
-        if (!modelMap.containsKey(n)) {
-            final StepDescriptor descriptor = lookupStepDescriptor(n);
-            Class<? extends Step> c = (descriptor == null ? null : descriptor.clazz);
-            modelMap.put(n, c != null ? new DescribableModel<>(c) : null);
-        }
-
-
-        return modelMap.get(n);
+        return modelMap.asMap().get(n);
```
rsandell: Are you sure this is what you want?
from the javadoc of `LoadingCache.asMap()`: **Note that although the view is modifiable, no method on the returned map will ever cause entries to be automatically loaded.**

abayer: Yeah, I buggered that up. I'll revisit in a later PR.

rsandell:
:bee:

PR #136, opened 2017-03-13
[FIXED JENKINS-42693] Add additionalBuildArgs dockerfile parameter

abayer:
* JENKINS issue(s):
    * [JENKINS-42693](https://issues.jenkins-ci.org/browse/JENKINS-42693)
* Description:
    * Allows specifying like
```groovy
dockerfile {
  additionalBuildArgs "--build-arg someArg=someValue"
}
```
which will get added to the "docker build ..." command line call.
* Documentation changes:
    * [Doc PR](https://github.com/jenkins-infra/jenkins.io/pull/751)
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
@@ -77,7 +77,8 @@ public class DockerPipelineFromDockerfileScript extends AbstractDockerPipelineSc
             try {
                 def hash = Utils.stringToSHA1(script.readFile("${describable.getDockerfilePath()}"))
                 def imgName = "${hash}"
-                script.sh "docker build -t ${imgName} -f \"${describable.getDockerfilePath()}\" \"${describable.getActualDir()}\""
+                def additionalBuildArgs = describable.getAdditionalBuildArgs() ? " ${describable.additionalBuildArgs}" : ""
+                script.sh "docker build -t ${imgName}${additionalBuildArgs} -f \"${describable.getDockerfilePath()}\" \"${describable.getActualDir()}\""
                 script.dockerFingerprintFrom dockerfile: describable.dockerfilePath, image: imgName, toolName: script.env.DOCKER_TOOL_NAME
                 return script.getProperty("docker").image(imgName)
             } catch (FileNotFoundException f) {
--- a/pipeline-model-definition/src/test/resources/additionalDockerBuildArgs.groovy
+++ b/pipeline-model-definition/src/test/resources/additionalDockerBuildArgs.groovy
@@ -1,0 +1,43 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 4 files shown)


pipeline-model-definition/src/test/resources/additionalDockerBuildArgs.groovy
```diff
@@ -0,0 +1,43 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    agent {
+        dockerfile {
+            args "-v /tmp:/tmp -p 8000:8000"
+            additionalBuildArgs "--build-arg someArg=thisOtherArg"
```
bitwiseman: Do you have a test for injection attack? (if someone uses `;` or `&&` or the like can the execute arbitrary shell code?   Not sure if it matters but might be worth considering.

abayer: Good question. Already a risk with `args` as well. Lemme add something for that.

abayer: Actually, that's a pain in the butt for something that really isn't a vulnerability - i.e., yes, someone could do a shell injection to the `docker build` command, but they could also do it by...a `sh` step. That is, we've already long since given access to run arbitrary shell code on agents. =)

bitwiseman: Agreed. As long as there isn't some automatic sudo that occurs for the docker stuff, I'm okay then.

abayer: There isn't. So now to fix the test. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
```diff
@@ -77,7 +77,8 @@ public class DockerPipelineFromDockerfileScript extends AbstractDockerPipelineSc
             try {
                 def hash = Utils.stringToSHA1(script.readFile("${describable.getDockerfilePath()}"))
                 def imgName = "${hash}"
-                script.sh "docker build -t ${imgName} -f \"${describable.getDockerfilePath()}\" \"${describable.getActualDir()}\""
+                def additionalBuildArgs = describable.getAdditionalBuildArgs() ? " ${describable.additionalBuildArgs}" : ""
+                script.sh "docker build -t ${imgName}${additionalBuildArgs} -f \"${describable.getDockerfilePath()}\" \"${describable.getActualDir()}\""
```
rsandell: btw is this really using the same docker executable logic as docker-pipeline-plugin? I thought it used some logic to check for tool installer first, this also does not seem to care about the `withDockerRegistry` stuff.

abayer: Good point. We had to move away from `docker.build(...)` due to it calling `dockerFingerprintFrom` against just the `Dockerfile` without taking into account working directory, etc, but I think we might be able to work around that. Worth raising a ticket.

PR #137, opened 2017-03-13
Remove pipeline-model-declarative-agent completely

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * We tombstoned the plugin in the 1.1 release, so now let's get rid of it completely.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees
    * @rsandell


abayer:
@rsandell When do you think would be a good point?

rsandell:
@abayer when we go 2.0 :)

abayer:
@rsandell Honest question - why wait? There are as of yet no external dependencies on `pipeline-model-declarative-agent`, and `pipeline-model-definition` depends on `pipeline-model-extensions`. Any upgrade of `pipeline-model-definition` will work fine.

rsandell:
Because if they do manual uppgrades for example, i.e. only upgrade specific plugins and not all, then if they uppgrade only pipeline-model-definition and not `pipeline-model-declarative-agent` then the old 1.0 version of the plugin will still be there and break loading of `pipeline-model-extensions` or whatever plugin has those classes now. If we say we depend on the tombstone version of `pipeline-model-declarative-agent` Jenkins will download the tombstone version when PMD gets upgraded and all is well.

abayer:
Ah-ha! Good point. I'll tweak this to depend on `pipeline-model-declarative-agent` 1.1.1 with an exclusion of the transitive dependency on `pipeline-model-extensions`. Thanks for walking me through that. =)

abayer:
@rsandell Ok, fixed - `pipeline-model-definition` now depends on `pipeline-model-declarative-agent:1.1.1` explicitly, so even if you upgrade straight from `1.0.2` -> `1.1.2` or later, you get the empty `1.1.1` release of `pipeline-model-declarative-agent`. It's goofy, but hey, it solves the problem!

PR #138, opened 2017-03-14
[FIXED JENKINS-42762] Allow multiple conditions in when directly

abayer:
* JENKINS issue(s):
    * [JENKINS-42762](https://issues.jenkins-ci.org/browse/JENKINS-42762)
* Description:
    * Reverts changes made in JENKINS-41185 fix. Does not affect the core of the changes in JENKINS-41185 - nested `when` conditions still work exactly the same, we're just back to allowing multiple top-level `when` conditions, where if any of them fail/return false, the `when` is not satisfied. This came to light when [a user reported the regression](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/125#issuecomment-286322236) so I consider it a priority.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @kzantow due to `ast-schema.json` reversion.

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -121,7 +121,7 @@ class ModelParser implements Parser {
         }
 
         if (pst==null) {
-            // Check if there's a 'pipeline' step somewhere children within the other statements and error out if that's the case.
+            // Check if there's a 'pipeline' step somewhere nexted within the other statements and error out if that's the case.
             src.statementBlock.statements.each { checkForNestedPipelineStep(it) }
             return null; // no 'pipeline', so this doesn't apply
         }
@@ -420,11 +420,8 @@ class ModelParser implements Parser {
         def stepsBlock = matchBlockStatement(statement)
         BlockStatement block = asBlock(stepsBlock.body.code)
         ModelASTWhen w = new ModelASTWhen(statement)
-        if (block.statements.size() != 1) {
-            errorCollector.error(w, Messages.ModelParser_WrongWhenCount())
-            return w
-        } else {
-            w.condition = parseWhenContent(block.statements.first())
+        block.statements.each { s ->
+            w.conditions.add(parseWhenContent(s))
         }
 
         return w
--- a/pipeline-model-definition/src/test/resources/json/whenEnvFalse.json
+++ b/pipeline-model-definition/src/test/resources/json/whenEnvFalse.json
@@ -31,25 +31,25 @@
           }]
         }]
       }],
-      "when": {"condition": {
+      "when": {"conditions": [      {
         "name": "environment",
-        "arguments": [
+        "arguments":         [
           {
             "key": "name",
-            "value": {
+            "value":             {
               "isLiteral": true,
               "value": "FOO"
             }
           },
           {
             "key": "value",
-            "value": {
+            "value":             {
               "isLiteral": true,
               "value": "SOME_OTHER_VALUE"
             }
           }
         ]
-      }}
+      }]}
     }
   ],
   "environment": [  {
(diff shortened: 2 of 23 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -121,7 +121,7 @@ class ModelParser implements Parser {
         }
 
         if (pst==null) {
-            // Check if there's a 'pipeline' step somewhere children within the other statements and error out if that's the case.
```
bitwiseman: nested

abayer: dangit.

abayer:
...yes. You did. =)

pipeline-model-definition/src/test/resources/json/whenEnvFalse.json
```diff
@@ -31,25 +31,25 @@
...
-            "value": {
+            "value":             {
               "isLiteral": true,
               "value": "FOO"
             }
           },
           {
             "key": "value",
-            "value": {
+            "value":             {
```
bitwiseman: What's with all the extra spaces?

abayer: I copy-pasted the JSON from our internal `ModelASTElement.toJSON().toString(2)` method, and sometimes that comes out weird in re: spacing.

IdanAdar:
Could someone please provide a declarative pipeline example with multiple `when`s?

rsandell:
@IdanAdar https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/test/resources/whenNestedCombinations.groovy

IdanAdar:
Thanks @rsandell. Would this then be a valid use?

```
stage ('....') {
    // We're in a PR, lets do something
    when {
        expression {
            (BRANCH_NAME != "develop") || (BRANCH_NAME != "master")
        }
    }

    steps {
        ...
    }
}
```

rsandell:
@IdanAdar yes it is valid, but would trigger for `master` since that is != `develop` and `BRANCH_NAME != "develop"` is evaluated first.
Better expressed would be:

```
 when {
    allOf {
        not { branch "develop" }
        not { branch "master" }
    }
}
```

abayer:
Or
```groovy
when {
    not {
        anyOf {
            branch "develop"
            branch "master"
        }
    }
}
```
because hey. =)

IdanAdar:
So likewise on the positive? 

```
when {
    branch "develop"
}

steps {
    ...
}
```

No need for `expression`?

IdanAdar:
@abayer @rsandell,

This succeeds: 

```
stage ("...") {
         environment {
            JENKINSBOT = credentials('...')
         }
         
         when {
            not { 
               anyOf {
                  branch "develop"
                  branch "master"
               }
            }
         }
         
         steps {
            // sh that uses $JENKINSBOT_PSW
         }
}
```

But this fails... I guess because of the `expression` bit:

```
stage ("...") {
         environment {
            JENKINSBOT = credentials('...')
         }
         
         when {
            allOf {
               expression { return env.shouldBuild != "false"  }
               not { branch "develop" }
               not { branch "master" }
            }
         }
         
         steps {
            // sh using $JENKINSBOT_PSW
         }
}
```

`env.shouldBuild` is defined like this:

```
pipeline {
    parameters {
        booleanParam(defaultValue: true, description: 'Execute pipeline?', name: 'shouldBuild')
   }
   ...
```

And updated like this:

```
result = sh (script: "git log -1 | grep '.*\\[ci skip\\].*'", returnStatus: true) 
if ((result == 0) && (BRANCH_NAME == "develop" || BRANCH_NAME == "master")) {
    echo ("This build should be skipped. Aborting.")
    env.shouldBuild = "false"
}
```

Background: I am doing this in order to skip all stages in some situations, because there is no ci skip functionality that works in declarative pipeline/multibranch pipeline.

PR #139, opened 2017-03-14
[FIXED JENKINS-42771] Allow + binary expressions in env values

abayer:
* JENKINS issue(s):
    * [JENKINS-42771](https://issues.jenkins-ci.org/browse/JENKINS-42771)
* Description:
    * This worked previously because we didn't actually pay attention to what was really in the value for an environment variable in the AST model, only at runtime. And the runtime parsing worked differently, so it did fine with `FOO = "A" + "B" + "C"`. Well, now so does the AST model. Yay!
* Documentation changes:
    * n/a, I think? @bitwiseman, I don't think we should be explicitly documenting that this works, especially since `FOO = "A" + "B" + "C"` will actually roundtrip through the editor to be `FOO = "ABC"`, since we concatenate the strings at parse time...
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -244,10 +244,22 @@ class ModelParser implements Parser {
                                 r.variables[key] = parseArgument(exp.rightExpression, true)
                                 return
                             } else if (exp.rightExpression instanceof MethodCallExpression) {
-                                r.variables[key] = parseInternalFunctionCall((MethodCallExpression)exp.rightExpression)
+                                r.variables[key] = parseInternalFunctionCall((MethodCallExpression) exp.rightExpression)
                                 return
+                            } else if (exp.rightExpression instanceof BinaryExpression) {
+                                if (((BinaryExpression)exp.rightExpression).operation.type  == Types.PLUS) {
+                                    // This is to support JENKINS-42771, allowing `FOO = "b" + "a" + "r"` sorts of syntax.
+                                    String envValue = envValueForStringConcat((BinaryExpression) exp.rightExpression)
+                                    if (envValue != null) {
+                                        r.variables[key] = ModelASTValue.fromConstant(envValue, exp.rightExpression)
+                                    }
+                                    return
+                                } else {
+                                    errorCollector.error(new ModelASTKey(exp.rightExpression), Messages.ModelParser_InvalidEnvironmentOperation())
+                                    return
+                                }
                             } else {
-                                errorCollector.error(key, Messages.ModelParser_InvalidEnvironmentValue())
+                                errorCollector.error(new ModelASTKey(exp.rightExpression), Messages.ModelParser_InvalidEnvironmentValue())
                                 return
                             }
                         }
@@ -269,6 +281,71 @@ class ModelParser implements Parser {
             }
         }
         return r;
+    }
+
+    /**
+     * This works functionally equivalent to the string used in {@link #parseArgument}, but without returning a
+     * {@link ModelASTValue}, returning just a string instead, and without handling {@link MapExpression} or
+     * {@link VariableExpression}, which is exclusively used for the `agent any` and `agent none` shortcuts regardless.
+     *
+     * @param e A non-null expression. If the expression is anything but a constant or a GString, an error will be thrown.
+     * @return The string value for that expression - in the case of {@link GStringExpression}, we explicitly remove the
+     * quotes from around it so that it's suitable for string concatenation. Returns null if an error was encountered.
+     */
+    @CheckForNull
+    private String envStringFromArbitraryExpression(@Nonnull Expression e) {
+        StringBuilder builder = new StringBuilder()
+        if (e instanceof ConstantExpression) {
+            builder.append(e.value)
+        } else if (e instanceof GStringExpression) {
+            String rawSrc = getSourceText(e)
+            builder.append(rawSrc.substring(1, rawSrc.length() - 1))
+        } else {
+            errorCollector.error(new ModelASTKey(e), Messages.ModelParser_InvalidEnvironmentConcatValue())
+            return null
+        }
+        return builder.toString()
+    }
+
+    /**
+     * Traverses a {@link BinaryExpression} known to be a {@link Types#PLUS}, to concatenate its various subexpressions
+     * together as string values.
+     * @param exp A non-null binary expression
+     * @return The concatenated string equivalent of that binary expression, assuming no errors were encountered on the
+     * various subexpressions, in which case it will return null.
+     */
+    @CheckForNull
+    private String envValueForStringConcat(@Nonnull BinaryExpression exp) {
+        StringBuilder builder = new StringBuilder()
+
+        if (exp.leftExpression instanceof BinaryExpression) {
+            if (((BinaryExpression)exp.leftExpression).operation.type  == Types.PLUS) {
+                String nestedString = envValueForStringConcat((BinaryExpression) exp.leftExpression)
+                if (nestedString == null) {
+                    return null
+                } else {
+                    builder.append(nestedString)
+                }
+            } else {
+                errorCollector.error(new ModelASTKey(exp), Messages.ModelParser_InvalidEnvironmentOperation())
+                return
+            }
+        } else {
+            String leftExpString = envStringFromArbitraryExpression(exp.leftExpression)
+            if (leftExpString == null) {
+                return null
+            } else {
+                builder.append(leftExpString)
+            }
+        }
+        String thisString = envStringFromArbitraryExpression(exp.rightExpression)
+        if (thisString == null) {
+            return null
+        } else {
+            builder.append(thisString)
+        }
+
+        return builder.toString()
     }
 
     public @Nonnull ModelASTLibraries parseLibraries(Statement stmt) {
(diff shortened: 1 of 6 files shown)


pipeline-model-definition/src/test/resources/errors/invalidMultiExpressionEnvironment.groovy
```diff
@@ -0,0 +1,50 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    environment {
+        FOO =
+            "B" -
```
bitwiseman: What does this minus even mean?

abayer: Doesn't matter! I'm just verifying that you'll get a specific syntax error if you use anything other than a single plus here. =)

pipeline-model-definition/src/test/resources/multiExpressionEnvironment.groovy
```diff
@@ -0,0 +1,49 @@
...
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    environment {
+        FOO =
+            "B" +
+                "${'A'}" +
+                "R"
+        _UNDERSCORE = "VALID"
```
bitwiseman: Could this also be `_UNDERSCORE = FOO + "VALID"?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -244,10 +244,22 @@ class ModelParser implements Parser {
                                 r.variables[key] = parseArgument(exp.rightExpression, true)
                                 return
                             } else if (exp.rightExpression instanceof MethodCallExpression) {
-                                r.variables[key] = parseInternalFunctionCall((MethodCallExpression)exp.rightExpression)
+                                r.variables[key] = parseInternalFunctionCall((MethodCallExpression) exp.rightExpression)
                                 return
+                            } else if (exp.rightExpression instanceof BinaryExpression) {
+                                if (((BinaryExpression)exp.rightExpression).operation.type  == Types.PLUS) {
+                                    // This is to support JENKINS-42771, allowing `FOO = "b" + "a" + "r"` sorts of syntax.
+                                    String envValue = envValueForStringConcat((BinaryExpression) exp.rightExpression)
```
rsandell: so what about other datatypes than string? e.g: `SOMETHING = 1 + 5` or `SOMETHING = 4 + "Bobby"` ?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -271,6 +283,71 @@ class ModelParser implements Parser {
...
+     * @return The string value for that expression - in the case of {@link GStringExpression}, we explicitly remove the
+     * quotes from around it so that it's suitable for string concatenation. Returns null if an error was encountered.
+     */
+    @CheckForNull
+    private String envStringFromArbitraryExpression(@Nonnull Expression e) {
+        StringBuilder builder = new StringBuilder()
+        if (e instanceof ConstantExpression) {
+            builder.append(e.value)
+        } else if (e instanceof GStringExpression) {
+            String rawSrc = getSourceText(e)
```
rsandell: So does this mean that GStrings gets evaluated at compile time now? That I believe needs to be documented since it is a different behavior to how groovy normally works.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: The latter will work - turns into `"4Bobby"`. The former...well, it'll be `"15"` currently. Need to think about it a bit - probably make sure at least one of the elements is a `String` or `GString`.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: Nope! `FOO = "A" + "${return 'B'}"` will end up equivalent to `FOO = "A${return 'B'}"`. The actual `GString` will still get evaluated at the same time no matter which way it's specified.

pipeline-model-definition/src/test/resources/multiExpressionEnvironment.groovy
bitwiseman: @abayer 
Would `_UNDERSCORE = FOO + "VALID"` also be allowed?  If so, test should include it.

liquid1982:
I ran into this issue too, and I have this block in my pipeline:

```
  environment {
    PROJECTS = env.PROJECTS
    PLATFORM_BRANCH = env.gitlabSourceBranch ? env.gitlabSourceBranch : env.PLATFORM_BRANCH
    UPDATE_DEPS = env.gitlabSourceBranch ? '1' : env.UPDATE_DEPS
  }
```

how am I suppose to modify my code to be compliant with the new changes?

abayer:
@liquid1982 Oh, you're easy - just change to `PLATFORM_BRANCH = "${env.gitlabSourceBranch ? env.gitlabSourceBranch : env.PLATFORM_BRANCH}"` etc and you're all good.

Senorsen:
I have this problem too, and my code is like this:
```
    environment {
        DB_SUFFIX = "ci_" + new Random().nextInt()
    }
```
After upgrade to plugin version 1.1.1, it won't work any more, and the error message shows up:
```

org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
WorkflowScript: 7: Environment variable values must either be strings or function calls. @ line 7, column 9.
           DB_SUFFIX = "ci_" + new Random().nextInt()
           ^

WorkflowScript: 6: No variables specified for environment @ line 6, column 5.
       environment {
       ^

2 errors

	at org.codehaus.groovy.control.ErrorCollector.failIfErrors(ErrorCollector.java:310)
	at org.codehaus.groovy.control.CompilationUnit.applyToPrimaryClassNodes(CompilationUnit.java:1085)
	at org.codehaus.groovy.control.CompilationUnit.doPhaseOperation(CompilationUnit.java:603)
	at org.codehaus.groovy.control.CompilationUnit.processPhaseOperations(CompilationUnit.java:581)
	at org.codehaus.groovy.control.CompilationUnit.compile(CompilationUnit.java:558)
	at groovy.lang.GroovyClassLoader.doParseClass(GroovyClassLoader.java:298)
	at groovy.lang.GroovyClassLoader.parseClass(GroovyClassLoader.java:268)
	at groovy.lang.GroovyShell.parseClass(GroovyShell.java:688)
	at groovy.lang.GroovyShell.parse(GroovyShell.java:700)
	at org.jenkinsci.plugins.workflow.cps.CpsGroovyShell.reparse(CpsGroovyShell.java:67)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.parseScript(CpsFlowExecution.java:430)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.start(CpsFlowExecution.java:393)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.run(WorkflowRun.java:238)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:405)
Finished: FAILURE
```

However, I tried clone this git repo (https://github.com/jenkinsci/pipeline-model-definition-plugin) and merge this pr (GH-139) locally, compile it and transfer all the *.hpi plugins to my Jenkins and restart - but no luck, still cannot work, and the same test error occurred.

I would like to know whether this pr can fix my case, or some of my steps were wrong, thanks!

abayer:
@Senorsen Have you tried 
```groovy
   environment {
        DB_SUFFIX = "ci_${new Random().nextInt()}"
    }
```

bitwiseman:
@abayer  - From the responses, obviously the docs will need to say some thing about this.  😄  

Is there any way to update the error message to provide a suggestion as to the fix?

Senorsen:
@abayer The snippet you provided works on version 1.1.1, thanks!

abayer:
@bitwiseman It now will say `Environment variable values to be concatenated together must be strings.` if you try to concatenate anything but strings.

bitwiseman:
@abayer I'm concerned that just saying "string" won't be sufficient.  

It should at least say "must be string literals" so people know that can't put bare variables or numbers in there.  

It might be even better to say something like "must be string literals or GString templates".  Sort of point people toward likely success.

abayer:
@bitwiseman How about "must be in single or double quotes"? Doesn't make people need to know what a GString literal is, etc.

bitwiseman:
@abayer - Are triple-quoted (`'''` or '"""`) strings not supported? 
If so, I think that's a bug - but a separate one from this. :smile:  

That aside - I can't think of a better message at this time.  We can adjust it later based on feedback, right?

abayer:
Triple-quoted strings are supported. And yes, we can adjust any time in the future.

PR #140, opened 2017-03-16
[FIXED JENKINS-42753] No more closure translation!

abayer:
* JENKINS issue(s):
* Description:
    * This enables/fixes a *lot* of things, here specifically allowing parameters to be referenced in `agent { label ... }`, though it sadly does not allow use of something from `environment` at the top level in an `agent` at the top level (or same for `environment`/`agent` at the same `stage` level) due to JENKINS-41890 taking priority - i.e., we evaluate `environment` after we evaluate `agent`. May revisit that later. ANYWAY.

- [x] Implementation works, all existing tests pass.
- [x] Ok, technically, all but one - `BasicModelTest#libraryObjectInScript` is blowing up over import issues. Needs more investigation, may block this whole thing, we'll see!
- [ ] Find more complicated usage patterns, such as those at [JENKINS-42777](https://issues.jenkins-ci.org/browse/JENKINS-42777), and make sure *those* still work too.
- [ ] No, seriously, TEST THE CRAP OUT OF THIS. I'm willing to bet there at least a couple regressions I haven't found due to not having coverage of those particular syntactic quirks, and I am not releasing this until we're 99% confident we've caught every possible regression.

* Documentation changes:
    * n/a - this is a purely internal change. Any changes in user-facing behavior would be regressions we need to find before we release anyway.
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell


abayer:
Incorporated [JENKINS-42829](https://issues.jenkins-ci.org/browse/JENKINS-42829) as well - notes from that commit:

Now that we're script.evaluate(...)ing everything, we need to be sure
that anything defined elsewhere in the Jenkinsfile outside the
pipeline block is available in that evaluated scope. This...is the
most effective way I can come up with to do so. It has some risks -
something with side effects could be problematic, but that's not a use
case we're keen on supporting anyway.

abayer:
This is a dead attempt, closing.

jglick:
Superseded by #174 perhaps?

PR #141, opened 2017-03-17
[FIXED JENKINS-42858] Process credentials before rest of environment

abayer:
* JENKINS issue(s):
    * [JENKINS-42858](https://issues.jenkins-ci.org/browse/JENKINS-42858)
* Description:
    * I don't think there's a reasonable use case for an environment variable being in the credentials ID, so there's really no reason *not* to do credentials first, is there? This allows credentials to be referenced by other environment variables.
    * Needs https://github.com/jenkinsci/credentials-binding-plugin/pull/34 to be merged/released before we can actually move forward with this, due to `withCredentials` currently requiring a workspace.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -71,9 +71,9 @@ public class ModelInterpreter implements Serializable {
 
                 // Entire build, including notifications, runs in the agent.
                 inDeclarativeAgent(root, root, root.agent) {
-                    withEnvBlock(root.getEnvVars(script)) {
-                        inWrappers(root.options) {
-                            withCredentialsBlock(root.getEnvCredentials()) {
+                    withCredentialsBlock(root.getEnvCredentials()) {
+                        withEnvBlock(root.getEnvVars(script)) {
+                            inWrappers(root.options) {
                                 toolsBlock(root.agent, root.tools) {
                                     for (int i = 0; i < root.stages.getStages().size(); i++) {
                                         Stage thisStage = root.stages.getStages().get(i)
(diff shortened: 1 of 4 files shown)


rsandell:
Since we switched to expanding the environment to after the agent has been allocated there is no reason any more. It used to be needed because the `withCredentials` step requires a workspace.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -91,19 +91,19 @@ public class ModelInterpreter implements Serializable {
...
                                                     // and so that we don't go into a per-stage agent if the when condition
                                                     // isn't satisfied.
-                                                    withEnvBlock(thisStage.getEnvVars(root, script)) {
-                                                        if (evaluateWhen(thisStage.when)) {
-                                                            inDeclarativeAgent(thisStage, root, thisStage.agent) {
-                                                                withCredentialsBlock(thisStage.getEnvCredentials()) {
+                                                    withCredentialsBlock(thisStage.getEnvCredentials()) {
+                                                        withEnvBlock(thisStage.getEnvVars(root, script)) {
+                                                            if (evaluateWhen(thisStage.when)) {
+                                                                inDeclarativeAgent(thisStage, root, thisStage.agent) {
```
rsandell: :bug: `withCredentials` step requires a workspace so this will only work when there is a global agent allocated.

abayer: So it won't work with `agent none` even without this change?

abayer: Yup, I see it now. I've opened https://issues.jenkins-ci.org/browse/JENKINS-42999 (with a PR at https://github.com/jenkinsci/credentials-binding-plugin/pull/34) to get `withCredentials` to stop requiring a workspace unless one or more of the `MultiBinding`s passed to it requires a workspace itself.

bitwiseman:
@abayer 
> I don't think there's a reasonable use case for an environment variable 

No, I can honestly say this not a safe assumption.  People will have different credentials depending on possibly branch or other env var - `"my-creds-for-${EXECUTION_ENVIRONMENT_NAME}"`. 

This  would a place where env var magical ordering is not as good as simple sequential evaluation.

abayer:
@bitwiseman Ugggggh. Well, that's a pain.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: That's in now, so yay.

abayer:
Closed in favor the omnibus at #147

PR #142, opened 2017-03-22
[FIXED JENKINS-41456] Properly validate publishHTML

abayer:
* JENKINS issue(s):
    * [JENKINS-41456](https://issues.jenkins-ci.org/browse/JENKINS-41456)
* Description:
    * Specifically, when we see that there's a sole required parameter, the only provided argument by the user is a map, at least one of those map keys doesn't have a corresponding parameter in the describable, and the sole required parameter's type is a describable itself, validate the argument map against that parameter type describable instead of against the parent describable. So for publishHTML, that means realizing `target` is an `HtmlPublisherTarget` and validating the map against that.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @bitwiseman

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
@@ -278,11 +278,13 @@ class ModelValidatorImpl implements ModelValidator {
                     soleDescribableMap = true
                     valid = validateDescribable(element, model.soleRequiredParameter.name, argList,
                         new DescribableModel<>(model.soleRequiredParameter.erasedType))
+                    // Note - this return is to break out of the .each loop only
                     return
                 }
 
                 if (!isValidStepParameter(model, k.key, k)) {
                     valid = false
+                    // Note - this return is to break out of the .each loop only
                     return
                 }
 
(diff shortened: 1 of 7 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -235,35 +235,59 @@ class ModelValidatorImpl implements ModelValidator {
...
+                // Check if there is a sole required parameter and it's describable
+                if (model.getParameter(k.key) == null &&
+                    model?.soleRequiredParameter != null &&
+                    Describable.class.isAssignableFrom(model.soleRequiredParameter.erasedType)) {
+                    // Check if the argument list validates as that describable. If it does, note that so
+                    // we can proceed.
+                    soleDescribableMap = true
+                    valid = validateDescribable(element, model.soleRequiredParameter.name, argList,
+                        new DescribableModel<>(model.soleRequiredParameter.erasedType))
+                    return
```
rsandell: `return valid` would have been a bit clearer and not have me sitting here wondering how the flow is for as long as I did :)

abayer: Except that `return` is actually "break out of the each loop" so yeah...more confusing! I'll add comments for clarity.

PR #143, opened 2017-03-22
[FIXED JENKINS-43035] Automatically include imports in when expr

abayer:
* JENKINS issue(s):
    * [JENKINS-43035](https://issues.jenkins-ci.org/browse/JENKINS-43035)
* Description:
    * Take imports in the Jenkinsfile and automatically include them in the code blocks for when expressions, so that they are actually available when we evaluate the when expression.
    * That said, I'm not sure I want to do it this way - this isn't round-trippable, etc. So thoughts/comments are welcome.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -30,6 +30,7 @@ import hudson.model.Describable
 import hudson.model.Descriptor
 import jenkins.model.Jenkins
 import org.codehaus.groovy.ast.ASTNode
+import org.codehaus.groovy.ast.ImportNode
 import org.codehaus.groovy.ast.ModuleNode
 import org.codehaus.groovy.ast.expr.*
 import org.codehaus.groovy.ast.stmt.BlockStatement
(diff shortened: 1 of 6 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -128,6 +129,17 @@ class ModelParser implements Parser {
...
 
+        List<ImportNode> additionalImports = []
+        [src.imports, src.staticImports, src.starImports, src.staticStarImports].each { imports ->
+            imports.each { ImportNode i ->
+                additionalImports.add(i)
+            }
+        }
+
+        if (!additionalImports.isEmpty()) {
+            r.setAdditionalImports(additionalImports.collect { it.text })
```
rsandell: Why not just skip the extra iteration and add `i.text` directly to `additionalImports`?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -643,6 +643,30 @@ public void libraryObjectInScript() throws Exception {
                 .go();
     }
 
+    @Issue("JENKINS-43035")
+    @Test
+    public void libraryObjectInWhenExpression() throws Exception {
+        otherRepo.init();
+        otherRepo.write("src/org/foo/Zot.groovy", "package org.foo;\n" +
```
rsandell: should probably test star and static imports as well in one script, to make sure they are added correctly.

rsandell:
Could we perhaps add a section to the schema for additional imports? The editor wouldn't need to do anything with it just keep it around and send it back on the roundtrip.

abayer:
Replaced this with #144, which is totally superior, yo.

PR #144, opened 2017-03-23
[FIXED JENKINS-43035] Add imports option for libraries

abayer:
* JENKINS issue(s):
    * [JENKINS-43035](https://issues.jenkins-ci.org/browse/JENKINS-43035)
* Description:
    * Note that this supersedes #143.
    * Also continues to support `libraries { lib("foo@master")}` for backwards compatible reasons, and adds existing `import foo` and `@Library` statements in the `Jenkinsfile` to the AST, thereby solving the original bug *and* roundtripping them off to the editor JSON!
    * Note - this is not backwards-compatible in the JSON form, but @kzantow assures me that the editor is not yet doing anything with libraries, so we still have that opportunity now anyway, and this format is going to be preferable in the long term anyway.
    * Still needs warnings added for usage of `@Library` or `import ...` directly in `Jenkinsfile` so that people are good and use the directives.
    * This allows you to specify libraries and imports as 
```groovy
libraries {
  "foo@master"
  "bar@master"
}
imports {
  "com.foo.Something"
  "static com.foo.Bar.BAZ"
  "com.foo.pkg.*"
  "static com.foo.Boo.*"
}
```
* Documentation changes:
    * @bitwiseman - it looks like we don't cover library usage in Declarative in [the docs](https://jenkins.io/doc/book/pipeline/shared-libraries/) - am I missing something?
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @kzantow 
    * @bitwiseman


abayer:
So endpoint validation/conversion is currently busted when you've got imports, since we can't actually load the library at that time. Thinking on it.

pipeline-model-definition/src/test/resources/libraryDirectiveObjectInWhenExpression.groovy
```diff
@@ -0,0 +1,55 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    agent any
+    libraries {
+        lib(library: 'zot-stuff@master', imports: [
```
bitwiseman: Can we use the full word `library` as the directive? 

Can the `library` parameter be `specifier` instead? (Ideally, I'd say have a required  `name` parameter and a separate optional parameter for the `version` ("master"), but I can also see keeping consistency with the @Library annotation format.) 

Also, what about `import`s not from a library?  Where do they go and what do they look like?  Just like normal `import` statements?

bitwiseman:
@abayer - You closed #143, but you don't have the tests from there.  Should this change also address that scenario are you coming back to it later?

pipeline-model-definition/src/test/resources/libraryDirectiveObjectInWhenExpression.groovy
abayer: Yup, we can do whatever naming you want there, though yeah, `specifier`, not separate `name` and `version`.

Honestly? I'm blowing off other imports. I'm not sure I see a good reason to support round-tripping them.

abayer:
So if I can't find a way to deal with the import resolution during endpoint validation/conversion, I may just punt on this entire JIRA and say "wrap your shared library classes in shared library global variables, which don't need importing". We shall see.

bitwiseman:
@abayer - If this handles the issue presented in #143, then if you pull the tests from that over and add a libraries section to them, they should work right? 

> wrap your shared library classes in shared library global variables 

Adding even more overhead? Ouch.

bitwiseman:
@abayer  re Declarative usage of libraries - it needs some expansion talking about the limitations.  But there shouldn't be a lot of differences from Scripted.  What do you think needs to be added?

pipeline-model-definition/src/test/resources/libraryDirectiveObjectInWhenExpression.groovy
bitwiseman: If you're going to blow off imports (not round trip them), that's another thing that needs to get a warning output.  They are supported now, not round tripping means deprecating them.  

could we handle non-library imports with a library directive without a specification, like this:
```
library(imports: [ 'org.apache.commons'])
```

abayer: Yup, it would require a warning.

abayer:
@bitwiseman We already have equivalent tests. I'm not sure what you're talking about - they're named a bit differently and obviously don't test *exactly* the same thing, since this is a different approach, but they're there.

bitwiseman:
@abayer - equivalent test: doh.  I see them now. 🙈 Sorry.

abayer:
FYI, I seem to have changed my mind again and decided to split out `imports` into its own directive after all, and add logic to auto-slurp literal `import foo.Bar` statements from the `Jenkinsfile` into the `imports` directive for translation-to-editor purposes, etc. Aaaaaand once I'm sure this all works, I'll add warnings for `@Library` and `import ...` usage, pointing to using the `libraries` and `imports` directives instead.

OH! And this is the syntax for 'em -
```groovy
libraries {
  "foo@master"
  "bar@master"
}
imports {
  "com.foo.Something"
  "static com.foo.Bar.BAZ"
  "com.foo.pkg.*"
  "static com.foo.Boo.*"
}
```
Yup, I figured out how to not need methods any more. I think. I gotta make sure this all works and I've got backwards compatibility hooked in, but...yeah. Progress?

bitwiseman:
I like it.  😄

rsandell:
Not having "real" imports will make future gdsl functionality a bit harder.

abayer:
@rsandell A user could add the `import ...` statements to the `Jenkinsfile` as well.

bitwiseman:
@abayer  Would those `import` statements round-trip or get stripped by the editor?

abayer:
@bitwiseman They would get roundtripped, but they'd move from `import ...` into `imports { ... }`. I don't see it very likely that someone's going to be editing the same `Jenkinsfile` through the editor *and* with GDSL completion in IntelliJ, but we'll preserve the data needed for runtime and the editor in either case.

jglick:
Seems like an increasingly crazy direction.

abayer:
That is not untrue. The question is where the line is between "excessively crazy" and "kinda bats but needed to deliver requirements".

abayer:
Deferring this 'til 1.2 at the earliest.

abayer:
Dead, closing.

PR #145, opened 2017-03-24
[FIXED JENKINS-42772] Don't fail out for single + required missing param

abayer:
* JENKINS issue(s):
    * [JENKINS-42772](https://issues.jenkins-ci.org/browse/JENKINS-42772)
* Description:
    * More specifically, don't fail validation if there's only a single parameter, it's required, and there are no arguments specified by the user. This may end up causing some surprise errors at runtime, but it appears that most everything is assuming they need to do null checks on just-one-and-it's-required parameters for steps, so let's give this a shot.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @stephendonner since this is your ticket I marked as not a bug, I thought you'd be pleased. =)


rsandell:
This looks like a workaround for bad design in another plugin, we should fix the other plugin instead. `@DataboundConstructor` parameters are supposed to be required, there's no real other way we can test for that requirement unless we start inspect it looking for things like `@CheckForNull` et.al.
If the plugin author says it is required we should also require it.

rochdev:
@rsandell In general I would agree, but the behaviour should also be consistent between declarative and scripted pipeline, and this is the current behaviour of the scripted pipeline. Also, a lot of plugins unfortunately rely on this for optional parameters.

rsandell:
But declarative is intended to be better than scripted in how it helps the user create pipelines and avoid pitfalls like not providing the correct parameters to a step etc. How can we do that if we become lenient on that front? Confusion and Anarchy :)

rochdev:
I know at a minimum [milestone](https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Milestone+Step+Plugin) and [ansiColor](https://wiki.jenkins-ci.org/display/JENKINS/AnsiColor+Plugin) are affected by this. Maybe it would be feasible to fix at least all plugins officially maintained by Jenkins.

abayer:
So I may be wrong, but I believe that without the parameter on a `@DataBoundConstructor`, you can't do `ansiColor("xterm")` - i.e., an unnamed parameter that is not also a mandatory parameter. Assuming I'm correct about that, then changing to not have the `@DataBoundConstructor` parameter would require users changing to `ansiColor(colorMapName: "xterm")`, which would be bad, since that would break all existing usage and look crappy. =)

jglick:
Right, it is legal for a step to declare a single constructor parameter but also treat it as optional at runtime. Maybe we would have been better served by an annotation like `@DefaultParameter` in `structs`, but that ship has sailed.

abayer:
@rsandell Good point. Lemme think on that.

abayer:
@rsandell Finally picking this up again. Re-review? =)

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/OptionsTest.java
```diff
@@ -42,7 +42,10 @@
 import org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty;
 import org.junit.Test;
 import org.jvnet.hudson.test.Issue;
+<<<<<<< HEAD
+=======
```
oleg-nenashev: 🐛 Diff

abayer: Oh c'mon me. Oy.

PR #147, opened 2017-03-27
[ENVIRONMENT OMNIBUS] Clean up environment issues stemming from 1.1 changes

abayer:
* JENKINS issue(s):
    * [JENKINS-43137](https://issues.jenkins-ci.org/browse/JENKINS-43137)
    * [JENKINS-43143](https://issues.jenkins-ci.org/browse/JENKINS-43143)
    * [JENKINS-43177](https://issues.jenkins-ci.org/browse/JENKINS-43177)
    * [JENKINS-43013](https://issues.jenkins-ci.org/browse/JENKINS-43013)
    * [JENKINS-42858](https://issues.jenkins-ci.org/browse/JENKINS-42858)
* Description:
    * Environment variables with triple-quoted strings and ones with double-quotes inside them did not evaluate correctly. This fixes that.
    * Make sure we process `params` into the environment, for normal environment variables and credentials.
    * Use our own variable resolution logic rather than that from `EnvVars` to handle more complex expressions.
    * The new round-robin variable resolution also deals with ordering issues!
    * Round-robin process credentials when we're resolving environment variables as well - we'll still end up re-processing credentials later in `withCredentials`, but this way, we can address [JENKINS-42858](https://issues.jenkins-ci.org/browse/JENKINS-42858) and let credential IDs refer to environment variables *and* environment variables refer to credentials env vars (within reason).
    * In general, simplify (ha!) a bunch of environment logic, rationalize it, make it consistent.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -324,25 +324,37 @@ public class Utils {
         return toEval
     }
 
+    static String trimQuotes(String s) {
+        if ((s.startsWith('"') && s.endsWith('"')) ||
+            (s.startsWith("'") && s.endsWith("'"))) {
+            return trimQuotes(s.substring(1, s.length() - 1))
+        } else {
+            return s
+        }
+    }
+
     static Environment environmentFromAST(WorkflowRun r, ModelASTEnvironment inEnv) {
         if (inEnv != null) {
             Environment env = new Environment()
 
-            Map<String, Object> inMap = [:]
+            Map<String, Environment.EnvValue> inMap = [:]
+            Map<String, CredentialWrapper> credMap = [:]
             inEnv.variables.each { k, v ->
                 if (v instanceof ModelASTInternalFunctionCall) {
                     ModelASTInternalFunctionCall func = (ModelASTInternalFunctionCall)v
                     // TODO: JENKINS-41759 - look up the right method and dispatch accordingly, with the right # of args
                     String credId = func.args.first().value
                     CredentialsBindingHandler handler = CredentialsBindingHandler.forId(credId, r)
-                    inMap.put(k.key, new CredentialWrapper(credId, handler.getWithCredentialsParameters(credId)))
+                    credMap.put(k.key, new CredentialWrapper(credId, handler.getWithCredentialsParameters(credId)))
                 } else {
-                    inMap.put(k.key, ((ModelASTValue)v).value)
+                    ModelASTValue val = (ModelASTValue)v
+                    Environment.EnvValue envVal = new Environment.EnvValue(isLiteral: val.isLiteral(), value: val.value)
+                    inMap.put(k.key, envVal)
                 }
             }
 
-            env.modelFromMap(inMap)
-
+            env.setValueMap(inMap)
+            env.setCredsMap(credMap)
             return env
         } else {
             return null
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -53,6 +53,7 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.model.Root
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.Stage
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.StepsBlock
 import org.jenkinsci.plugins.pipeline.modeldefinition.parser.Converter
+import org.jenkinsci.plugins.pipeline.modeldefinition.steps.CredentialWrapper
 import org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditional
 import org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditionalDescriptor
 import org.jenkinsci.plugins.structs.SymbolLookup
@@ -335,7 +336,25 @@ public class Utils {
         }
         return credsTuples
     }
-    
+
+    /**
+     * This exists for pre-1.1.2 -> later upgrades of running builds only.
+     *
+     * @param environment The environment to pull credentials from
+     * @return A non-null but possibly empty map of strings to {@link CredentialWrapper}s
+     */
+    @Nonnull
+    static Map<String, CredentialWrapper> getLegacyEnvCredentials(@Nonnull Environment environment) {
+        Map<String, CredentialWrapper> m = [:]
+        environment.each {k, v ->
+            if (v instanceof  CredentialWrapper) {
+                m["${k}"] = v;
+            }
+        }
+        return m
+    }
+
+
     // Note that we're not using StringUtils.strip(s, "'\"") here because we want to make sure we only get rid of
     // matched pairs of quotes/double-quotes.
     static String trimQuotes(String s) {
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -337,8 +337,9 @@ public class Utils {
         }
         return credsTuples
     }
-
-
+    
+    // Note that we're not using StringUtils.strip(s, "'\"") here because we want to make sure we only get rid of
+    // matched pairs of quotes/double-quotes.
     static String trimQuotes(String s) {
         if ((s.startsWith('"') && s.endsWith('"')) ||
             (s.startsWith("'") && s.endsWith("'"))) {
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -318,11 +318,26 @@ public class Utils {
     public static String prepareForEvalToString(String s) {
         String toEval = s ?: ""
         if (!toEval.startsWith('"') || !toEval.endsWith('"')) {
-            toEval = '"' + toEval + '"'
+            if (toEval.indexOf('\n') == -1) {
+                toEval = '"' + toEval + '"';
+            } else {
+                toEval = '"""' + toEval + '"""';
+            }
         }
 
         return toEval
     }
+
+    static List<List<String>> getEnvCredentials(Environment environment, CpsScript script) {
+        List<List<String>> credsTuples = new ArrayList<>()
+        if (environment != null) {
+            credsTuples.addAll(environment?.getCredsMap(script)?.collect { k, v ->
+                [k, v]
+            })
+        }
+        return credsTuples
+    }
+
 
     static String trimQuotes(String s) {
         if ((s.startsWith('"') && s.endsWith('"')) ||
@@ -338,14 +353,14 @@ public class Utils {
             Environment env = new Environment()
 
             Map<String, Environment.EnvValue> inMap = [:]
-            Map<String, CredentialWrapper> credMap = [:]
+            Map<String, Environment.EnvValue> credMap = [:]
             inEnv.variables.each { k, v ->
                 if (v instanceof ModelASTInternalFunctionCall) {
                     ModelASTInternalFunctionCall func = (ModelASTInternalFunctionCall)v
                     // TODO: JENKINS-41759 - look up the right method and dispatch accordingly, with the right # of args
-                    String credId = func.args.first().value
-                    CredentialsBindingHandler handler = CredentialsBindingHandler.forId(credId, r)
-                    credMap.put(k.key, new CredentialWrapper(credId, handler.getWithCredentialsParameters(credId)))
+                    Environment.EnvValue envVal = new Environment.EnvValue(isLiteral: func.args.first().isLiteral(),
+                        value: func.args.first().value)
+                    credMap.put(k.key, envVal)
                 } else {
                     ModelASTValue val = (ModelASTValue)v
                     Environment.EnvValue envVal = new Environment.EnvValue(isLiteral: val.isLiteral(), value: val.value)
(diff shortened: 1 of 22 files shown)


khobbits:
Thanks for getting this done so quickly!  Out of curiosity, what is the release cycle for something like this?  The previous release broke a good number of our jobs that were reliant on this behaviour.

abayer:
@rsandell Needs re-review now that I've, well, refactored. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -318,31 +318,59 @@ public class Utils {
...
+            }
         }
 
         return toEval
     }
 
+    static List<List<String>> getEnvCredentials(Environment environment, CpsScript script) {
+        List<List<String>> credsTuples = new ArrayList<>()
+        if (environment != null) {
+            credsTuples.addAll(environment?.getCredsMap(script)?.collect { k, v ->
```
rsandell: _nit_ You already know that `environment != null` so you don't need the first `?.`

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -318,31 +318,59 @@ public class Utils {
...
+        }
+        return credsTuples
+    }
+    
+    // Note that we're not using StringUtils.strip(s, "'\"") here because we want to make sure we only get rid of
+    // matched pairs of quotes/double-quotes.
+    static String trimQuotes(String s) {
+        if ((s.startsWith('"') && s.endsWith('"')) ||
+            (s.startsWith("'") && s.endsWith("'"))) {
+            return trimQuotes(s.substring(1, s.length() - 1))
```
rsandell: _nit_ `s[1..-1]` should also work I think.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
```diff
@@ -35,31 +36,64 @@ import org.jenkinsci.plugins.workflow.cps.EnvActionImpl
  * @author Andrew Bayer
  */
 @SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
-public class Environment extends MappedClosure<String,Environment> implements PropertiesToMap {
+public class Environment implements Serializable {
+    Map<String,EnvValue> valueMap = new TreeMap<>()
+    // TODO: Actually do stuff with creds again
```
rsandell: Yes that would be nice if they continued to work

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
```diff
@@ -35,31 +36,64 @@ import org.jenkinsci.plugins.workflow.cps.EnvActionImpl
...
+    Map<String,EnvValue> credsMap = new TreeMap<>()
```
rsandell: New fields in runtime data might break running pipelines when upgrading.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -247,9 +250,26 @@ public class ModelInterpreter implements Serializable {
         }
     }
 
+    private Map<String,CredentialWrapper> processCredentials(@Nonnull List<List<String>> varsAndIds) {
```
rsandell: `List<List<String>>` is a very convoluted data structure.
`Map.Entry` is serializable iirc.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Good point.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: `s[1..-2]`, as it turns out.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
abayer: Old comment. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
abayer: Yup, this needs upgrade testing.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: I'll play with it. `HashMap.Entry` is definitely not serializable, can't remember in re: `TreeMap.Entry`.

abayer: Nope, `TreeMap.Entry` isn't serializable either.

abayer:
So syntax-wise, we're good on backwards compatibility - no syntax changes are involved here, all existing tests pass, etc. It's runtime-upgrade testing that's needed, and that's...hard to figure out. Working on it.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
```diff
@@ -65,31 +65,86 @@ public class Environment implements Serializable {
...
+                unsuccessfulCount++
+                unsetKeys.add(nextKey)
+            }
+        }
+
+        return alreadySet
+    }
+
+    private boolean containsVariable(String var, Set<String> keys) {
+        def group = (var =~ /(\$\{.*?\})/)
```
rsandell: Iirc Environment.groovy isn't executed in the CPS engine so a regex should be safe here.

abayer: Yup.

abayer:
Suspicious that there's a hung test in that last commit. Digging into it.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
```diff
@@ -94,19 +97,69 @@ public class Environment implements Serializable {
         unsetKeys.addAll(overrides.keySet())
 
         GroovyShell shell = new GroovyShell(new Binding(alreadySet))
-        shell.setProperty("params", (Map<String,String>)script.getProperty("params"))
+        shell.setProperty("params", (Map<String, String>) script.getProperty("params"))
+
+        Map<String, String> preCreds = roundRobin(shell, alreadySet, overrides, unsetKeys)
+
+        Run<?, ?> r = script.$build()
```
rsandell: What does this do? Looks sacred

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
```diff
@@ -94,19 +97,69 @@ public class Environment implements Serializable {
...
+        credsMap.each { k, v ->
+            try {
+                String resolvedCredId = (String) shell.evaluate(Utils.prepareForEvalToString(v.value.toString()))
+                CredentialsBindingHandler handler = CredentialsBindingHandler.forId(resolvedCredId, r)
+
+                if (handler != null) {
+                    handler.toBindings(k, resolvedCredId).each { b ->
+                        if (!b.descriptor.requiresWorkspace()) {
+                            // We don't actually bother resolving anything that needs a workspace at this point.
+                            MultiBinding.MultiEnvironment bindingEnv = b.bind(r, null, null, TaskListener.NULL)
```
rsandell: I can't really pretend to understand what half of this is supposed to do.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/CredentialsBindingHandler.java
```diff
@@ -69,6 +71,9 @@ public boolean handles(StandardCredentials c) {
     }
 
     @Nonnull
+    public abstract List<MultiBinding<C>> toBindings(String varName, String credentialsId);
```
rsandell: This can't be abstract as it is an addition to an already released API signature.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
abayer: Meaning I need to comment it up better.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/CredentialsBindingHandler.java
abayer: Fair point. I'll move it out.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
abayer: It's getting the build - simpler then using `currentBuild` in a non script context.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/CredentialsBindingHandler.java
abayer: Or rather, I'll add a non-abstract default.

rumpl:
Hi all, 

Any ETA on when this will be released ?
Our Jenkins is red since the last update.

Thanks for the great work.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -247,9 +264,26 @@ public class ModelInterpreter implements Serializable {
...
+        Map<String,CredentialWrapper> creds = new TreeMap<>()
+        RunWrapper currentBuild = script.getProperty("currentBuild")
+
+        for (int i = 0; i < varsAndIds.size(); i++) {
+            String key = varsAndIds.get(i)?.get(0)
+            if (key != null) {
+                String id = (String)script.evaluate(Utils.prepareForEvalToString(varsAndIds.get(i)?.get(1)))
```
rsandell: this looks definitely like a sandbox escape.

abayer: Good catch - fixed.

abayer:
@rumpl As soon as it passes review it'll get merged and released. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Feck, that broke All The Things. Digging.

abayer: Oh, hey, I actually did the sandboxing elsewhere. Here is not actually a problem - `CpsScript.evaluate(script)` actually goes through `CpsScript.$getShell().evaluate(script)`, and that uses the `CpsThread.current().getExecution().getShell()`, which is not trusted, so will fail with `RejectedAccessException`s when appropriate. Phew! I did get rid of the `GroovyShell` we were using directly in `Environment` in favor of wrapping around `SecureGroovyScript.evaluate(classloader, binding)` so that area is sandboxed as well.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
```diff
@@ -266,4 +253,11 @@ public class Environment implements Serializable {
         public boolean isLiteral
         public Object value
     }
+
+    private String resolveAsScript(Binding binding, String script) {
+        String toRun = Utils.prepareForEvalToString(script)
+        SecureGroovyScript toExec = new SecureGroovyScript(toRun, true)
+            .configuring(ApprovalContext.create().withCurrentUser().withKey(Utils.getCurrentJobName()))
```
rsandell: `withKey`: _Associates a unique key with this approval. If not null, any previous approval of the same kind with the same key will be canceled and replaced by this one._

Since you seem to be calling this several times in the same script with the same key they would overwrite eachother. Is that intended?

abayer: Good point. Need to make that unique...

abayer: Hmm, actually, https://github.com/jenkinsci/script-security-plugin/blob/master/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ApprovalContext.java#L104 says that's only relevant for whole script approvals...since we're sandboxing the `SecureGroovyScript`, we are doing signature approvals, not whole script approvals, so this should be ok.

rsandell: So I think you should remove the usage of `whithKey` then to avoid future confusion, since it's not used anyway.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -627,4 +673,8 @@ public class Utils {
             return Collections.singletonMap(UninstantiatedDescribable.ANONYMOUS_KEY, _args)
         }
     }
+
+    static String getCurrentJobName() {
```
rsandell: Do you still need this one?

abayer: Nope, and I also had a superfluous parenthesis. Fixed both.

abayer:
Tentatively believe the test failure is transient noise but verifying.

abayer:
Yup, didn't show up when run locally, but waiting for the build here to finish before merging.

abayer:
Aaaaand we're good. Woot.

PR #149, opened 2017-03-30
feat: allow to user variables (again) in environment declaration

Twister42:
* Description:
    * We can't use variables in environment declarations since 1.1 release
    * This fix resolves the second point in the provided JenkinsFile.
```
def BAR = "bar";
pipeline {
    environment {
        FOO = "${BAR}" // Binding Error
        FOO = BAR          // Environment variable values must either be strings or function calls
    }
    agent {
        label "some-label"
    }

    stages {
        stage("foo") {
            steps {
                sh 'echo "FOO is $FOO"'
            }
        }
    }
}
```


abayer:
So the first point is covered in https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/147 (along with a barrel o' other `environment` related issues). The second...well, honestly, I'm not sure we want to support that.

Twister42:
Damn, didn't saw #147 .
Ok if first case is solved i'm fine with it

PR #150, opened 2017-04-06
[FIXED JENKINS-43404] Escape/unescape strings being evaluated

abayer:
* JENKINS issue(s):
    * [JENKINS-43404](https://issues.jenkins-ci.org/browse/JENKINS-43404)
* Description:
    * Not doing this resulted in unescaped double quotes and therefore errors, so let's make sure we always escape strings when they're passed in to be evaluated and unescape them after the evaluation.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -29,6 +29,7 @@ import com.google.common.cache.CacheBuilder
 import com.google.common.cache.CacheLoader
 import com.google.common.cache.LoadingCache;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
+import groovy.json.StringEscapeUtils
 import hudson.ExtensionList
 import hudson.model.Describable
 import hudson.model.Descriptor
@@ -68,7 +69,6 @@ import org.jenkinsci.plugins.workflow.graphanalysis.LinearBlockHoppingScanner
 import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode
 import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode
 import org.jenkinsci.plugins.workflow.graph.FlowNode
-import org.jenkinsci.plugins.workflow.job.WorkflowJob
 import org.jenkinsci.plugins.workflow.job.WorkflowRun
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor
 import org.jenkinsci.plugins.workflow.support.steps.StageStep
@@ -319,9 +319,9 @@ public class Utils {
         String toEval = s ?: ""
         if (!toEval.startsWith('"') || !toEval.endsWith('"')) {
             if (toEval.indexOf('\n') == -1) {
-                toEval = '"' + toEval + '"';
+                toEval = '"' + StringEscapeUtils.escapeJava(toEval) + '"';
             } else {
-                toEval = '"""' + toEval + '"""';
+                toEval = '"""' + StringEscapeUtils.escapeJava(toEval) + '"""';
             }
         }
 
--- a/pipeline-model-definition/src/test/resources/envQuotesInQuotes.groovy
+++ b/pipeline-model-definition/src/test/resources/envQuotesInQuotes.groovy
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016, CloudBees, Inc.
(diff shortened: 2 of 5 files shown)


abayer:
Noooope, regressions ahoy. Gotta handle escaping outside GStrings but not *inside* them. Youch.

abayer:
Ok, this version passes all of `EnvironmentTest`, which is the biggest hurdle. Making sure everything else is clear...

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -319,15 +320,30 @@ public class Utils {
...
-                toEval = '"""' + toEval + '"""';
+                toEval = '"""' + escapeForEval(toEval) + '"""';
             }
         }
 
         return toEval
     }
 
+    static String escapeForEval(String s) {
+        s = StringUtils.replace(s, '\\\\', '\\\\')
```
rsandell: This seems to do nothing.

pipeline-model-definition/src/test/resources/envBackslashes.groovy
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    stages {
+        stage("foo") {
+            steps {
+                echo "SIMPLE_BACKSLASH is ${SIMPLE_BACKSLASH}"
+                echo "NESTED_BACKSLASH is ${NESTED_BACKSLASH}"
+                echo "HAS_TAB is ${HAS_TAB}"
+                echo "NESTED_HAS_TAB is ${NESTED_HAS_TAB}"
+            }
```
rsandell: Should maybe try some with `sh "echo ..."` as well?

pipeline-model-definition/src/test/resources/envQuotesInQuotes.groovy
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+pipeline {
+    agent any
+
+    environment {
+        // Regrettably needed since GitSampleRepoRule doesn't set BRANCH_NAME.
+        BRANCH_NAME = "master"
+
+        GRADLE_OPTIONS = "--no-daemon --rerun-tasks -PBUILD_NUMBER=${env.BUILD_NUMBER} -PBRANCH=\"${env.BRANCH_NAME}\""
```
rsandell: Probably try something similar with single quotes? as well as multi line single quotes and multi line doubles.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Amusingly, it does do something! Basically it replaces `\\` with `\\\\`. It just looks funny here.

pipeline-model-definition/src/test/resources/envBackslashes.groovy
abayer: Could, but wouldn't make a difference, I don't think.

pipeline-model-definition/src/test/resources/envQuotesInQuotes.groovy
abayer: Yup yup.

pipeline-model-definition/src/test/resources/envBackslashes.groovy
rsandell: Well it would be a check for a `$` without `{}` to make sure normal sh isn't broken. E.g.

```
  sh 'echo "shell NESTED_HAS_TAB is $NESTED_HAS_TAB"'
```

rsandell: Well with multiline I meant

```
  """Something
like
this
"""
```

abayer: Isn't the multiline covered in https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/150/files#diff-4b9c896342cca18f47042db619b3a311R38?

abayer: Made the change to the shell echo, but fwiw, that won't make any difference. =)

rsandell: OK, I just wanted to make sure we covered it somehow since it has bitten out behind before :)

abayer: Very fair point! =)

PR #151, opened 2017-04-11
[FIXED JENKINS-43486] Handle non-String parameters in env resolution

abayer:
* JENKINS issue(s):
    * [JENKINS-43486](https://issues.jenkins-ci.org/browse/JENKINS-43486)
* Description:
    * non-`String` parameters (most notably `booleanParam`) were causing `ClassCastException`s when we tried to add them to the map for `environment` resolution, so do an explicit `toString()` on them.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
@@ -60,7 +60,11 @@ public class Environment implements Serializable {
 
     public Map<String,String> getCredsMap(CpsScript script) {
         Map<String,String> resolvedMap = new TreeMap<>()
-        EnvVars contextEnv = new EnvVars((Map<String,String>)script.getProperty("params"))
+        EnvVars contextEnv = new EnvVars()
+        ((Map<String,Object>)script.getProperty("params")).each { k, v ->
+            contextEnv.put(k, v.toString())
+        }
+
         contextEnv.overrideExpandingAll(((EnvActionImpl)script.getProperty("env")).getEnvironment())
 
         credsMap.each { k, v ->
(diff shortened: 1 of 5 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Environment.groovy
```diff
@@ -60,7 +60,11 @@ public class Environment implements Serializable {
 
     public Map<String,String> getCredsMap(CpsScript script) {
         Map<String,String> resolvedMap = new TreeMap<>()
-        EnvVars contextEnv = new EnvVars((Map<String,String>)script.getProperty("params"))
+        EnvVars contextEnv = new EnvVars()
+        ((Map<String,Object>)script.getProperty("params")).each { k, v ->
+            contextEnv.put(k, v.toString())
```
oleg-nenashev: 🐛 will 💥 on null parameter, which seems to be possible in Groovy

abayer: Turns out a null parameter causes other problems entirely -
```
java.lang.NullPointerException
	at org.jenkinsci.plugins.workflow.cps.persistence.IteratorHack.writeReplace(IteratorHack.java:52)
	at org.jenkinsci.plugins.workflow.cps.ParamsVariable.addValue(ParamsVariable.java:88)
	at org.jenkinsci.plugins.workflow.cps.ParamsVariable.getValue(ParamsVariable.java:75)
	at org.jenkinsci.plugins.workflow.cps.CpsScript.getProperty(CpsScript.java:121)
```
I'll add some NPE checks but that's a bug in `workflow-cps`, I think.

abayer: [JENKINS-43511](https://issues.jenkins-ci.org/browse/JENKINS-43511) opened for that.

abayer: And error-checking/test added.

abayer: That was already fixed in [JENKINS-42367](https://issues.jenkins-ci.org/browse/JENKINS-42367), so I'm bumping to a newer `workflow-cps` version...

emilegg:
Is this fixed ?

abayer:
@oleg-nenashev Re-review?

PR #152, opened 2017-04-13
[FIXED JENKINS-41334] Adding parallel stages.

abayer:
* JENKINS issue(s):
    * [JENKINS-41334](https://issues.jenkins-ci.org/browse/JENKINS-41334)
* Description:
    * This implements [the JENKINS-41334 RFC](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/rfc/JENKINS-41334-parallel-stage-execution.md), adding support for one level of nested parallel `stage`s. The backend implementation does not actually use the `stage` step at this point, just `parallel` branches, but the user experience with configuration, etc is the same.
* Documentation changes:
    * Will be needed, obviously.
* Users/aliases to notify:
    * @reviewbybees 
    * @rsandell 
    * @michaelneale 
    * @i386 
    * @kzantow 
    * @HRMPW


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
```diff
@@ -102,31 +116,37 @@ public String toGroovy() {
...
-        if (branches.size() > 1) {
-            result.append("parallel(");
-            boolean first = true;
-            for (ModelASTBranch branch: branches) {
-                if (first) {
-                    first = false;
-                } else {
-                    result.append(',');
+        if (branches.isEmpty() && parallelStages != null) {
+            result.append("parallelStages {\n");
```
rsandell: This does not follow the RFC.
And revenge of the camelCase :) Why not go with `parallel` as the RFC states? Confusion? There are plenty of synonyms or other contextually applickable things to `parallel` to choose from if you don't want to use that word; `threads`, `simultaneous`, `concurrent`. But I personally think that reusing the word `stages` is best. It self explains in a way that it is sub stages to this stage although that could add to the meaning as it is not apparent they would run in parallel so it would change the implementation a bit since probably an attribute should then be added to declare them to be executed in parallel.

rsandell: And in that last case, why not allow all stages to potentially be run in parallel, even the outer ones :)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -168,6 +143,62 @@ public class ModelInterpreter implements Serializable {
...
+    def evaluateStage(Root root, Agent parentAgent, Stage thisStage, Throwable firstError, Stage parentStage = null) {
+        return {
+            try {
+                // NOTE - this will switch to script.stage in the future.
+                if (firstError != null) {
+                    Utils.logToTaskListener("Stage '${thisStage.name}' skipped due to earlier failure(s)")
+                    Utils.markStageSkippedForFailure(thisStage.name)
+                } else if (skipUnstable(root.options)) {
+                    Utils.logToTaskListener("Stage '${thisStage.name}' skipped due to earlier stage(s) marking the build as unstable")
+                    Utils.markStageSkippedForUnstable(thisStage.name)
```
rsandell: Will this work for nested stages since there is no actual stage to mark?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -168,6 +143,62 @@ public class ModelInterpreter implements Serializable {
...
+            } catch (Exception e) {
+                script.getProperty("currentBuild").result = Result.FAILURE
+                Utils.markStageFailedAndContinued(thisStage.name)
+                if (firstError == null) {
+                    firstError = e
+                }
+            }
+
+            if (firstError != null) {
+                throw firstError
```
rsandell: This looks a bit iffy whether nested stages will get their post run or not. Some tests should be added that checks for exceptions in various parts of a nested stage tree.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
abayer: Compromise, basically. I believe that `parallel` is too overloaded for sure, the synonyms for `parallel` are completely new terms for Pipeline and so I'd prefer to avoid them (both to prevent the need for learning new terminology and, more importantly, to keep those terms free for possible later use!), and just `stages` is too overloaded as well. 

With this, `stages` ends up meaning "execute these stages linearly" while `parallel` means, as in Scripted, "execute these arbitrary blocks in parallel", and the new `parallelStages` means, fairly explicitly, "execute these stages concurrently". While I don't have any further use for those distinctions now, I can envision some possible use cases in the future.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -80,35 +80,10 @@ public class ModelInterpreter implements Serializable {
                                     for (int i = 0; i < root.stages.getStages().size(); i++) {
                                         Stage thisStage = root.stages.getStages().get(i)
                                         try {
+                                            // NOTE that this will eventually be moved into evaluateStage, once BO can
+                                            // handle visualizing stage-inside-parallel.
```
abayer: Yup.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Good question! Needs a test.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: agreed.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
bitwiseman: `parallelStages` seems reasonable to me.    But we it is probably worth getting buy off from more people (and updating the RFC?).  Questions like: can `parallelStages` be used instead of `stages` at the top of pipeline?  

The first thing someone is going to do when they see `stage('A') { parallelStages { stage ('B') ...` is expect to be able to do `stage('A') { parallelStages { stage ('B') { agent { ...`.  And that is not supported in this implementation, right? 

I'd almost rather push this out and do this in a way that is consistent with the `stage` behavior elsewhere.

abayer:
RFC already was updated. =)

Lemme think on the top-level idea. Not sure about that.

bitwiseman:
@abayer - making all `stage` calls have the same syntax and I'd describe it less as "top-level" more more "`stages` and `parallelStages` should be syntactically interchangeable - anywhere I can put one, I should be able to put the other".   Otherwise, it will confuse and annoy people.

HRMPW:
@bitwiseman  I think it will get more confusing trying to nest `stages` and `parallelStages`.  If they are the same you can nest `parallelStages` inside `stages` then it follows that you can nest `stages` inside `stages` and `parallelStages` inside `parallelStages`.

The alternative could be to make them both top-level only:

`
pipeline {
  agent any
  stages {
    stage ("A") { }
  }
  parallelStages {
    stage("P1") {}
    stage("P2") {} 
 }
 stages { 
    stage("B") {}
}
`

The big disadvantage I see with treating `stages` and `parallelStages` as equivalents is that you lose the wrapping `stage` currently proposed around `parallelStages`. This lets you set environment, agent, when, and post for the entire `parallelStages` block. You completely lose that if you by promoting `parallelStages`.

abayer:
Oh, we're *definitely* not supporting multiple `stages` at the top level in any form. I'm still leaning hard towards just allowing `parallelStages` within a `stage` and having that syntax disconnect since it is meaningful - `stage`s inside `parallelStages` themselves can't contain `parallelStages`, so there is a difference. Plus, Blue Ocean is still going to want the actual parallel "stages" to be nested within a parent stage, so we'd have to create a synthetic stage for top-level parallel stages.

HRMPW:
I agree that we should leave this inside a `stage` but to @bitwiseman 's point I do think calling it `parallelStages` does imply that it is the same as `stages`.  Alternatives: 

- parallel - might be overloaded but it is simple
- furcate - too obscure?
- split 
- children

bitwiseman:
> `stage`s inside `parallelStages` themselves can't contain `parallelStages`  

There's a test that contains exactly that. ... Oh, I see, it's an error case.  Sigh.

@HRMPW  
I'm with Andrew  on having a single top-level entry.  I'd like to allow `parallelStages` or `stages`, but that may have been a misunderstanding on my part. 

@abayer
What Blue Ocean wants is not sufficient justification (IMHO) for how we design the declarative syntax.  😄  
 
I do feel strongly that we don't want to have `stage` have different syntax depending on where it shows up.   That syntax disconnect would be worse than having to wait for this feature.

abayer:
@bitwiseman `stage` *has* to have slightly different syntax unless we instead go with an entirely different term for an individual parallel "stage". The difference is just that an individual parallel "stage" can't itself contain further parallel "stages".

I'm not yet persuaded that parallel-at-top-level is a good idea.

bitwiseman:
@abayer @HRMPW @rsandell - Seems like this is something that would benefit from a meeting.

abayer:
Wait, @bitwiseman, re-reading your original comment, I think you may have misunderstood - you can't do 
```groovy
pipeline {
    stages {   
         stage("foo") {
             parallelStages {
                 stage("first") {
                     parallelStages {
                         stage("first-and-one") {
```
just
```groovy
pipeline {
    stages {   
         stage("foo") {
             parallelStages {
                 stage("first") {
```
i.e., a `stage` within `parallelStages` cannot itself contain `parallelStages`. We only support one level of nesting.

bitwiseman:
@abayer  - Here's my core problem with the `stage` syntax disconnect - if you say "all `stage` syntax and behavior is the same", then you don't need to test as thoroughly.   Because if it works in `stage` one place it works everywhere.   

If `stage` is going to have different behavior here than elsewhere, I'd like to see basically *all* the tests you have for `stages-stage` duplicated here for `parallelStages-stage`, with the ones that specifically aren't supported marked as such.

abayer:
@bitwiseman I disagree. The only thing that needs to be tested is that you can't use `parallelStages` in an already nested `stage` - the execution paths are identical, the only difference is at validation time.

abayer:
Well, ok, *technically* they're not identical behind the scenes in that in `stages->stage`, it's actually invoking a closure inside a `stage` step, while in `parallelStages->stage`, it's invoking a closure inside a `parallel` step, but the content of that closure (i.e., everything about `stage` in Declarative other than "actually running the `stage` step") is identical.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: And done.

bitwiseman:
@abayer - Ah. I see.  From a test perspective, I would never trust what you just said.  😃  How about let's compromise and add a test or two showing `parallelStages->stage` using other elements of `stage` than just `steps`.  Show it with an `agent`,  an `environment`, etc.   Think of the test as  functional documentation.  They test framework you have works really well in this respect.

abayer:
@bitwiseman Happy to do so!

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Yyyup, this is buggy. Working on it.

abayer: So the nested stages were doing `post` fine, but the *parent* stage was not. So...juggling things around. Hooray.

abayer:
@bitwiseman And `environment` is thoroughly busted, so good call on the additional testing needed. =)

abayer:
@rsandell @bitwiseman And now we have testing of nested stages with `agent`, `post`, `environment`, `tools`, and `when`. And they all work, which some of them (`post` for parent stages, `environment` and `when` for nested stages) did not previously. Now if you'll excuse me, I'm going to go collapse for a while. =)

bitwiseman:
@abayer - Excellent! Thanks!

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -168,6 +143,68 @@ public class ModelInterpreter implements Serializable {
...
+                    // agent, we do the reverse per-stage. Why? So that the per-stage
+                    // environment is populated before we evaluate any when condition,
+                    // and so that we don't go into a per-stage agent if the when condition
+                    // isn't satisfied.
+                    withEnvBlock(thisStage.getEnvVars(root, script, parentStage)) {
+                        if (evaluateWhen(thisStage.when)) {
+                            if (thisStage.parallelStages != null) {
+                                def parallelStages = [:]
+                                for (int i = 0; i < thisStage.parallelStages.stages.size(); i++) {
+                                    Stage parallelStage = thisStage.parallelStages.getStages().get(i)
```
rsandell: Can you please stick to one syntax type, either `parallelStages.stages` or `parallelStages.getStages()`. I would prefer `parallelStages.stages[i]` as well instead of `parallelStages.stages.get(i)`

abayer:
Rebased on `master`, didn't have a chance to verify tests passed, so will loop back in tomorrow.

michaelneale:
oh thats a shame, went with parallelStages vs just "parallel" (I guess it wasn't possible to repurpose the existing parallel function?)

i386:
Will have to agree with @michaelneale here - this does not line up with the RFC so 🐜   from me too.

rsandell:
@michaelneale @i386 I believe it is technically possible to call it just `parallel` and I suggested it as well, but @abayer doesn't like it, you need to convince him ;)

abayer:
@i386 @michaelneale We discussed this offline a few weeks back? I'm still open to alternative names, but vehemently opposed to overloading `parallel` for comprehension reasons, as we discussed.

HRMPW:
I don't think it has to be `parallel` but I think `parallelStages` creates more comprehension problems than `parallel` does; it is too easily confused with the top-level `stages`.

abayer:
bah! I am outvoted. Back to `parallel` it is.

michaelneale:
👏

what about `parallel_stages`

*ducks*

I'll show myself out 💣

i386:
@abayer thank you 🙏 👍

bitwiseman:
@abayer - 🐝  When does this get merged?  :)

i386:
Soon as we integrate it into Blue Ocean - likely going to pick this up in the next week or so 🙌

abayer:
@i386 Does that mean we can release this now so that you can work against the release, or do you want to work against SNAPSHOTs?

kzantow:
Definitely release it! =)

abayer:
Actually, I can't release it until we doc it - cc @bitwiseman @rtyler - any chance we could get some cycles to work on documenting this next week? I'm kinda hit and miss in re availability for the week.

i386:
Ahh no... if we release and doc now people will be able to write Jenkinsfiles that they cannot visualize. Poor users :'(

kuhnroyal:
I am not using the visualizations anyways, developers are fine with Jenkinsfile's... just those DevOps... xD

i386:
@abayer could we get a beta released to the experimental update center that contains this change? That should allow for Blue Ocean to start integrating it.

i386:
Scheduling changes for visualization and editor so we can unblock asap

abayer:
@i386 Definitely prioritize the editor - the visualization can lag without any problem, the only *real* pressing issue would be to have the editor able to round-trip cleanly. Beta coming now.

abayer:
`1.2-beta-1` is releasing as we speak.

i386:
@abayer seems like its not in the maven repository?

```
Downloading: https://nexus-internal.cloudbees.com/content/groups/mirror/org/jenkinsci/plugins/pipeline-model-definition/1.2-beta-1/pipeline-model-definition-1.2-beta-1.pom
[WARNING] The POM for org.jenkinsci.plugins:pipeline-model-definition:jar:1.2-beta-1 is missing, no dependency information available
Downloading: https://nexus-internal.cloudbees.com/content/groups/mirror/org/jenkinsci/plugins/pipeline-model-definition/1.2-beta-1/pipeline-model-definition-1.2-beta-1.jar
```

abayer:
Trying again as `1.2-beta-2`.

quasarea:
@abayer will it support generation of dynamic parallel stages? i.e. I could have a method outside of pipeline that returns list of stages to execute in parallel based on some conditions.

i386:
@quasarea not as part of this change. You might want to try a `script` step and put your parallel within that instead.

pipeline-model-api/src/main/resources/ast-schema.json
```diff
@@ -445,7 +445,7 @@
       "additionalProperties": false
     },
     "stage": {
-      "description": "A single Pipeline stage, with a name and one or more branches",
+      "description": "A single Pipeline stage, with a name and either one or more branches or one or more nested stages",
```
oleg-nenashev: Oxford commas maybe? My built-in parser gave up here

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -601,20 +601,41 @@ class ModelValidatorImpl implements ModelValidator {
         return valid
     }
 
-    public boolean validateElement(@Nonnull ModelASTStage stage) {
+    public boolean validateElement(@Nonnull ModelASTStage stage, boolean isNested) {
```
oleg-nenashev: Add default value to retain compatibility? OTOH the class is not public

vkotovv:
Is there a fixed date for 1.2 release? I mean, when could I try it in release, not beta?

abayer:
No fixed date, no. @i386, once https://github.com/jenkinsci/pipeline-stage-step-plugin has a release, we can do a `1.2-beta-3` release, but we do need that release.

kuhnroyal:
Skipped parallel stages are marked as successful in blue-ocean instead of the question mark. Is that being tracked somewhere?

i386:
@kuhnroyal I can get someone to test that for us on the blue ocean side.

i386:
@kuhnroyal could you provide a minimal Jenkinsfile example please?

i386:
@jglick could we please get a release of https://github.com/jenkinsci/pipeline-stage-step-plugin ?

kuhnroyal:
@i386 I am doing something like this:

```
pipeline {
  stages {
    stage('Build') {
      parallel {
        stage('Verify') {
          when {
            anyOf {
              branch 'master'
              branch 'PR**'
              branch 'feature/**'
            }
          }
          steps {
            echo "mvn verify"
          }
        }
        stage('Deploy') {
          when {
            branch 'develop'
          }
          steps {
             echo "mvn deploy"
          }
        }
        stage('Release') {
          when {
            branch 'release/**'
          }
          steps {
            echo "mvn release"
          }
        }
      }
    }
  }
}
```

Only the matching parallel stage is execute but all of them show the green checked icon in blue-ocean.

abayer:
Reverting out 9e729c47b87333ebc96dd38f7597298cbfc9b831 for now until there's a `pipeline-stage-step` release, so I can get a `1.2-beta-3` beta release out.

abayer:
@i386 @michaelneale FYI, I'm releasing 1.1.9 from master right now and then will be merging this to master, with the intention of not doing another release from master until Blue Ocean is ready for this. It's just been too damned long sitting unmerged. =)

rtyler:
omigosh

abayer:
@rtyler Not released yet! I just decided that it made no sense to sit unmerged any longer. I still owe a doc update PR too.

bitwiseman:
@abayer - Still, every step toward this being merged is a win.

wagnst:
stumbled upon this one yesterday as well, would love to see it bein released 😄

samgomena:
I'm not sure if this is still open for discussion, but I currently have this pipeline setup: 

pipeline_before

However, I need a pipeline that looks like this: 

pipeline_after
(Note: this pipeline was made possible by the use of Photoshop)

Is it possible to achieve this architecture with this update?

kzantow:
@samgomena no, this only changes `parallel` to accept `stage`s, however I believe what you're asking for is planned, just not necessarily a path forward yet.

morganchristiansson:
I've noticed that the stages are reported in different order sometimes which disruptes the pipeline stage view history.

If stages within parallel were always reported in the order they are defined it would work better.

michaelneale:
Morgan do you mind opening that as a new JIrA as I have seen that too.

quasarea:
@kuhnroyal I have created an issue about not rendering properly skipped stages https://issues.jenkins-ci.org/browse/JENKINS-47219

PR #153, opened 2017-04-21
[FIXED JENKINS-43339] Properly set result from FlowInterruptedException

abayer:
* JENKINS issue(s):
    * [JENKINS-43339](https://issues.jenkins-ci.org/browse/JENKINS-43339)
* Description:
    * We should be setting the build result to `FlowInterruptedException.getResult()` when we catch a `FlowInterruptedException`, since there's actually a relevant result there. Also added a `notBuilt` post status while I was here.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @rsandell 
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
@@ -23,12 +23,23 @@
  */
 package org.jenkinsci.plugins.pipeline.modeldefinition;
 
+import com.gargoylesoftware.htmlunit.html.HtmlPage;
 import hudson.model.Result;
+import hudson.model.queue.QueueTaskFuture;
+import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.support.steps.input.InputAction;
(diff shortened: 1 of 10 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
```diff
@@ -99,4 +110,89 @@ public void buildConditionOrdering() throws Exception {
...
+        WorkflowRun run1 = job.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("wait/1", run1);
+        WorkflowRun run2 = job.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("wait/2", run2);
+
+        SemaphoreStep.success("wait/2", null);
+
+        j.waitForCompletion(run2);
+
+        j.assertBuildStatus(Result.NOT_BUILT, j.waitForCompletion(run1));
```
rsandell: Might be prudent to check that `run2` was successful as well.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
```diff
@@ -99,4 +110,89 @@ public void buildConditionOrdering() throws Exception {
...
+
+        j.assertLogContains("Job not built due to milestone", run1);
```
rsandell: Might be prudent to check that `run2` does not contain this text.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
abayer: Whoops!

PR #156, opened 2017-05-16
[FIXED JENKINS-43055] Change SyntheticStageGraphListener to an Extension

abayer:
* JENKINS issue(s):
    * [JENKINS-43055](https://issues.jenkins-ci.org/browse/JENKINS-43055)
* Description:
    * Use the new `GraphListener` `ExtensionPoint` to have a single `SyntheticStageGraphListener` for all runs rather than one per run, with some logic for only firing on Declarative runs. Downstream of the now-merged https://github.com/jenkinsci/workflow-api-plugin/pull/38 and https://github.com/jenkinsci/workflow-cps-plugin/pull/137
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
@@ -87,7 +87,7 @@ public final class SyntheticStageGraphListener implements GraphListener {
     @Extension
     public static class FlowExecutionListenerImpl extends FlowExecutionListener {
         @Override
-        public void onRunning(FlowExecution execution) {
+        public void onRunning(FlowExecution execution, boolean resumed) {
             if (execution != null && !execution.isComplete()) {
                 // Only attach the graph listener if we know the run this execution is for is a Declarative run,
                 // i.e., if it has an ExecutionModelAction.
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
@@ -25,6 +25,9 @@
 package org.jenkinsci.plugins.pipeline.modeldefinition;
 
 import hudson.Extension;
+import hudson.model.InvisibleAction;
+import hudson.model.Run;
+import jenkins.model.RunAction2;
 import org.jenkinsci.plugins.pipeline.SyntheticStage;
 import org.jenkinsci.plugins.pipeline.modeldefinition.actions.ExecutionModelAction;
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
@@ -39,6 +42,7 @@ import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.support.steps.StageStep;
 
 import java.io.IOException;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.WARNING;
@@ -96,9 +100,22 @@ public final class SyntheticStageGraphListener implements GraphListener {
                         }
                     }
                 } catch (IOException e) {
-                    // Do nothing
+                    LOGGER.log(Level.WARNING, "Failure to get run for FlowExecution: {0}", e);
                 }
             }
+        }
+    }
+
+    @Deprecated
+    public static class GraphListenerAction extends InvisibleAction implements RunAction2 {
+        @Override
+        public void onLoad(Run<?,?> r) {
+            // no-op
+        }
+
+        @Override
+        public void onAttached(Run<?, ?> r) {
+            // no-op
         }
     }
 }
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
@@ -107,15 +107,6 @@ public final class SyntheticStageGraphListener implements GraphListener {
     }
 
     @Deprecated
-    public static class GraphListenerAction extends InvisibleAction implements RunAction2 {
-        @Override
-        public void onLoad(Run<?,?> r) {
-            // no-op
-        }
-
-        @Override
-        public void onAttached(Run<?, ?> r) {
-            // no-op
-        }
+    public static class GraphListenerAction extends InvisibleAction {
     }
 }
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.pipeline.modeldefinition;
 
 import hudson.Extension;
+import hudson.model.Actionable;
 import hudson.model.InvisibleAction;
 import org.jenkinsci.plugins.pipeline.SyntheticStage;
 import org.jenkinsci.plugins.pipeline.modeldefinition.actions.ExecutionModelAction;
@@ -35,7 +36,6 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
-import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.support.steps.StageStep;
 
 import javax.annotation.Nonnull;
@@ -57,12 +57,7 @@ public final class SyntheticStageGraphListener implements GraphListener {
     public void onNewHead(FlowNode node) {
         if (node != null && node instanceof StepStartNode &&
                 ((StepStartNode) node).getDescriptor() instanceof StageStep.DescriptorImpl) {
-            FlowExecution execution = node.getExecution();
-            if (!declarativeRuns.containsKey(execution)) {
-                declarativeRuns.put(execution, isDeclarativeRun(execution));
-            }
-
-            if (declarativeRuns.get(execution)) {
+            if (isDeclarativeRun(node.getExecution())) {
                 LabelAction label = node.getPersistentAction(LabelAction.class);
                 if (label != null &&
                         (SyntheticStageNames.preStages().contains(label.getDisplayName()) ||
@@ -94,19 +89,24 @@ public final class SyntheticStageGraphListener implements GraphListener {
         }
     }
 
-    private boolean isDeclarativeRun(@Nonnull FlowExecution execution) {
-        try {
-            FlowExecutionOwner owner = execution.getOwner();
-            if (owner != null && owner.getExecutable() instanceof WorkflowRun) {
-                if (((WorkflowRun) owner.getExecutable()).getAction(ExecutionModelAction.class) != null) {
-                    return true;
+    private synchronized boolean isDeclarativeRun(@Nonnull FlowExecution execution) {
+        if (!declarativeRuns.containsKey(execution)) {
+            boolean isDeclarative = false;
+            try {
+                FlowExecutionOwner owner = execution.getOwner();
+                if (owner != null && owner.getExecutable() instanceof Actionable) {
+                    if (((Actionable) owner.getExecutable()).getAction(ExecutionModelAction.class) != null) {
+                        isDeclarative = true;
+                    }
                 }
+            } catch (IOException e) {
+                LOGGER.log(Level.WARNING, "Error loading WorkflowRun for FlowNode: {0}", e);
             }
-        } catch (IOException e) {
-            LOGGER.log(Level.WARNING, "Error loading WorkflowRun for FlowNode: {0}", e);
+
+            declarativeRuns.put(execution, isDeclarative);
         }
 
-        return false;
+        return declarativeRuns.get(execution);
     }
 
     @Deprecated
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
@@ -26,13 +26,16 @@ package org.jenkinsci.plugins.pipeline.modeldefinition;
 
 import hudson.Extension;
 import org.jenkinsci.plugins.pipeline.SyntheticStage;
+import org.jenkinsci.plugins.pipeline.modeldefinition.actions.ExecutionModelAction;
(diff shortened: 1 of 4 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -79,29 +77,12 @@ private void attachTag(FlowNode currentNode, String syntheticContext) {
...
-
-        private void attachListener(WorkflowRun run) {
-            if (run != null) {
-                FlowExecution exec = run.getExecution();
-                if (exec != null && !exec.isComplete()) {
-                    exec.addListener(new SyntheticStageGraphListener());
-                }
+        public void onRunning(FlowExecution execution) {
+            if (execution != null && !execution.isComplete()) {
+                execution.addListener(new SyntheticStageGraphListener());
```
rsandell: So this means you add the listener to all pipelines, not only declarative?

abayer: ...huh. Good point.

abayer: Ok, tweaked to only add the `SyntheticStageGraphListener` if the corresponding `WorkflowRun` has the `ExecutionModelAction` on it - i.e., we know it's a Declarative Pipeline.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -79,28 +80,23 @@ private void attachTag(FlowNode currentNode, String syntheticContext) {
         }
     }
 
-    public static class GraphListenerAction extends InvisibleAction implements RunAction2 {
```
jglick: Note that this may trigger some `OldDataMonitor` warnings. Harmless I think, though if you want to avoid it you can keep the action around, but deprecated and a no-op.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -79,28 +80,23 @@ private void attachTag(FlowNode currentNode, String syntheticContext) {
...
+                try {
+                    FlowExecutionOwner owner = execution.getOwner();
+                    if (owner != null && owner.getExecutable() instanceof WorkflowRun) {
+                        WorkflowRun r = (WorkflowRun)owner.getExecutable();
+                        if (r.getAction(ExecutionModelAction.class) != null) {
+                            execution.addListener(new SyntheticStageGraphListener());
+                        }
+                    }
+                } catch (IOException e) {
+                    // Do nothing
```
jglick: Log a warning.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -79,28 +80,23 @@ private void attachTag(FlowNode currentNode, String syntheticContext) {
...
-                    exec.addListener(new SyntheticStageGraphListener());
+        public void onRunning(FlowExecution execution) {
+            if (execution != null && !execution.isComplete()) {
+                // Only attach the graph listener if we know the run this execution is for is a Declarative run,
+                // i.e., if it has an ExecutionModelAction.
+                try {
+                    FlowExecutionOwner owner = execution.getOwner();
+                    if (owner != null && owner.getExecutable() instanceof WorkflowRun) {
+                        WorkflowRun r = (WorkflowRun)owner.getExecutable();
+                        if (r.getAction(ExecutionModelAction.class) != null) {
```
jglick: Hmm, this is not right. Perhaps `FlowExecutionOwner` needs a `getAction` method. But it seems you are already hard-coding the assumption that we are dealing with a `WorkflowRun`? That needs to be fixed.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -79,28 +80,23 @@ private void attachTag(FlowNode currentNode, String syntheticContext) {
...
+                            execution.addListener(new SyntheticStageGraphListener());
```
jglick: I will point out that this is actually more code than simply making `SyntheticStageGraphListener` an `@Extension` and having its `onNewHead` check whether its `owner` has `ExecutionModelAction`. (I suppose you could cache that calculation with a `CacheBuilder` with weak keys, though it should be very cheap anyway since it involves no I/O or significant computation.)

abayer: Correct me if I'm wrong, but is it actually more code? With `onNewHead`, we'd have to go `node` -> `execution` -> `owner` -> `WorkflowRun` anyway, so basically the same code as here (even with caching, we'd still have to do it once, same as here).

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
abayer: I'm confused about the second part of that comment. Yes, I'm assuming we're dealing with a `WorkflowRun` - because that's the only context where I'd care about the listener.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
abayer: Fair point. Fixing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
jglick: Yes you would have to follow the same chain of APIs to determine if this `FlowNode` is from a Declarative build. What _is_ simpler is that you are just reacting passively to events. There is no listener to attach, no risk that you have the timing wrong and there is some kind of subtle race condition, no risk of attaching the same listener twice, etc. So the patch would be shorter (and IMO clearer and safer) than what you have now.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -96,9 +100,22 @@ public void onRunning(FlowExecution execution) {
...
                 }
             }
         }
     }
+
+    @Deprecated
+    public static class GraphListenerAction extends InvisibleAction implements RunAction2 {
+        @Override
+        public void onLoad(Run<?,?> r) {
+            // no-op
```
jglick: You could just drop `implements RunAction2`:

```java
@Deprecated
public static class GraphListenerAction extends InvisibleAction {}
```

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
jglick: Not really a problem in this PR, just something I noticed: `attachDeclarativeActions` is wrong to assume that the owner is a `WorkflowRun`. Really this plugin should have no `workflow-job` dependency at all.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
abayer: Done.

abayer:
Ok, I'm now reluctantly considering `GraphListener` as `ExtensionPoint` again...

jglick:
> I'm now reluctantly considering `GraphListener` as `ExtensionPoint` again...

Still, or no longer?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -54,15 +52,31 @@
...
                 }
-                if (SyntheticStageNames.postStages().contains(label.getDisplayName())) {
-                    attachTag(node, SyntheticStage.getPost());
+            } catch (IOException e) {
+                LOGGER.log(Level.WARNING, "Error loading WorkflowRun for FlowNode: {0}", e);
+            }
+
+            // If the action isn't present, this isn't a Declarative run, so remove the listener.
+            if (action == null) {
+                node.getExecution().removeListener(this);
```
jglick: Well this definitely looks more brittle than just using `GraphListener` as an `@Extension`.

abayer: You're probably right. Sigh.

abayer:
re-ping @jglick for re-review now that we're on `GraphListener`-as-`ExtensionPoint`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -32,41 +32,37 @@
...
-            try {
-                FlowExecutionOwner owner = node.getExecution().getOwner();
-                if (owner != null && owner.getExecutable() instanceof WorkflowRun) {
-                    action = ((WorkflowRun) owner.getExecutable()).getAction(ExecutionModelAction.class);
-                }
-            } catch (IOException e) {
-                LOGGER.log(Level.WARNING, "Error loading WorkflowRun for FlowNode: {0}", e);
+            FlowExecution execution = node.getExecution();
+            if (!declarativeRuns.containsKey(execution)) {
+                declarativeRuns.put(execution, isDeclarativeRun(execution));
```
jglick: :bug: I think you need to have `synchronized` access to this singleton map.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -98,23 +94,19 @@ private void attachTag(FlowNode currentNode, String syntheticContext) {
...
-        }
-
-        private void attachGraphListener(FlowExecution execution) {
-            if (execution != null && !execution.isComplete()) {
-                execution.addListener(new SyntheticStageGraphListener());
+    private boolean isDeclarativeRun(@Nonnull FlowExecution execution) {
+        try {
+            FlowExecutionOwner owner = execution.getOwner();
+            if (owner != null && owner.getExecutable() instanceof WorkflowRun) {
+                if (((WorkflowRun) owner.getExecutable()).getAction(ExecutionModelAction.class) != null) {
```
jglick: Could just use `Run` or even `Actionable`—no need for `WorkflowRun`, unless you are already relying on that elsewhere here.

pom.xml
```diff
@@ -75,17 +75,17 @@
...
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-cps</artifactId>
-        <version>2.29</version>
+        <version>2.33-20170525.192309-1</version> <!-- TODO: update to release once 2.33 is released -->
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-job</artifactId>
-        <version>2.9</version>
+        <version>2.11</version>
```
jglick: This is no longer necessary IIUC?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
abayer: Good point, fixing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
abayer: Ok.

pom.xml
abayer: Correct, but I'd rather leave it at the newer version anyway - I'd prefer to be testing/building against the newest version for our baseline (2.7.3), so why not leave it there?

jglick: No problem, just making sure you did not forget to update dependencies to match the new upstream PRs.

abayer: Nope, I deliberately switched from the `2.12-TIMESTAMPYSTUFF` upstream PR of `workflow-job` (that will hopefully still get released at some point, since I do still think there's a valid use case for `FlowExecutionListener`, if not here) to `2.11`. =)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/SyntheticStageGraphListener.java
```diff
@@ -24,40 +24,50 @@
...
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.WARNING;
 
+@Extension
 public final class SyntheticStageGraphListener implements GraphListener {
     private static final Logger LOGGER = Logger.getLogger(SyntheticStageGraphListener.class.getName());
 
+    private final transient Map<FlowExecution,Boolean> declarativeRuns = new WeakHashMap<>();
```
jglick: BTW `transient` is meaningless here—the class is not serializable.

abayer: Worth removing or fine to keep here?

jglick: Not a problem, just FYI.

jglick:
Can you update the PR description please? It is pointing to the wrong upstreams.

pom.xml
```diff
@@ -75,17 +75,17 @@
...
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-api</artifactId>
-        <version>2.11</version>
+        <version>2.16-20170524.173224-1</version> <!-- TODO: update to release once 2.16 is released -->
```
jglick: released

pom.xml
```diff
@@ -75,17 +75,17 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-cps</artifactId>
-        <version>2.29</version>
+        <version>2.33-20170525.192309-1</version> <!-- TODO: update to release once 2.33 is released -->
```
jglick: 2.33 released

PR #159, opened 2017-05-25
Validate converted Jenkinsfile before returning it.

abayer:
* JENKINS issue(s):
    * Related to [JENKINS-44039](https://issues.jenkins-ci.org/browse/JENKINS-44039) but not exactly the same.
* Description:
    * There are possible edge cases where the editor may not have done a validation of the JSON before saving it - so just to be safe, let's do a validation of the converted Jenkinsfile before returning it from `toJenkinsfile`.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/endpoints/ModelConverterAction.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/endpoints/ModelConverterAction.java
@@ -98,8 +98,17 @@ public class ModelConverterAction implements RootAction {
                 ModelASTPipelineDef pipelineDef = parser.parse();
 
                 if (!collectErrors(result, parser.getErrorCollector())) {
-                    result.accumulate("result", "success");
-                    result.accumulate("jenkinsfile", pipelineDef.toPrettyGroovy());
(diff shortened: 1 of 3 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/endpoints/ModelConverterAction.java
```diff
@@ -98,8 +98,17 @@ public HttpResponse doToJenkinsfile(StaplerRequest req) {
                 ModelASTPipelineDef pipelineDef = parser.parse();
 
                 if (!collectErrors(result, parser.getErrorCollector())) {
-                    result.accumulate("result", "success");
-                    result.accumulate("jenkinsfile", pipelineDef.toPrettyGroovy());
+                    try {
+                        Converter.scriptToPipelineDef(pipelineDef.toPrettyGroovy());
```
kzantow: ... assuming this does the groovy validation?

abayer: Yup, it does.

PR #160, opened 2017-06-01
[FIXED JENKINS-44621] Only remove job properties from Jenkinsfiles

abayer:
* JENKINS issue(s):
    * [JENKINS-44621](https://issues.jenkins-ci.org/browse/JENKINS-44621)
* Description:
    * Rather than actually using the properties(...) step directly, we
instead will now do this behind the scenes by interacting with the job
directly. We'll also record what job properties, triggers, and
parameters were defined in the Jenkinsfile for a given run, so that we
can refer back in the next build to see what existing job properties,
triggers, or parameters were defined through the Jenkinsfile vs
defined outside of the Jenkinsfile. This way, we can preserve the
defined-outside-Jenkinsfile properties et al.
    * Note that this won't *quite* work right if the first run of a job
after upgrading removes triggers/properties/etc, because there is no
record for builds beforehand, so all properties etc on the job at the
time of upgrade will be treated as if they were defined
externally. However, after that first build, you can remove properties
etc from the Jenkinsfile with the expected behavior.
* Documentation changes:
    * Could probably use a release note of some sort for the "first build after upgrade" caveat above. 
* Users/aliases to notify:
    * @reviewbybees 
    * @vivek @i386 (since this removes the sneaky `properties` step execution)
    * @bwalding because he stumbled across this in the first place. =)

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -30,17 +30,23 @@ import com.google.common.cache.CacheLoader
 import com.google.common.cache.LoadingCache;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
 import groovy.json.StringEscapeUtils
+import hudson.BulkChange
 import hudson.ExtensionList
 import hudson.model.Describable
 import hudson.model.Descriptor
+import hudson.model.JobProperty
+import hudson.model.ParameterDefinition
 import hudson.model.ParametersDefinitionProperty
 import hudson.model.Result
+import hudson.triggers.Trigger
+import jenkins.model.BuildDiscarderProperty
 import org.apache.commons.codec.digest.DigestUtils
 import org.apache.commons.lang.StringUtils
 import org.jenkinsci.plugins.pipeline.StageStatus
 import org.jenkinsci.plugins.pipeline.StageTagsMetadata
 import org.jenkinsci.plugins.pipeline.SyntheticStage
 import org.jenkinsci.plugins.pipeline.modeldefinition.actions.ExecutionModelAction
+import org.jenkinsci.plugins.pipeline.modeldefinition.actions.JobPropertyTrackerAction
 import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTEnvironment
 import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTInternalFunctionCall
 import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTPipelineDef
@@ -75,10 +81,12 @@ import org.jenkinsci.plugins.workflow.graph.FlowNode
 import org.jenkinsci.plugins.workflow.job.WorkflowJob
 import org.jenkinsci.plugins.workflow.job.WorkflowRun
 import org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty
+import org.jenkinsci.plugins.workflow.multibranch.BranchJobProperty
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor
 import org.jenkinsci.plugins.workflow.support.steps.StageStep
 
+import javax.annotation.CheckForNull
 import javax.annotation.Nonnull
 import javax.annotation.Nullable
 import javax.lang.model.SourceVersion
@@ -231,7 +239,7 @@ public class Utils {
                     return true
                 }
             } else {
-                return true
+                return !(p instanceof BranchJobProperty)
             }
         }
     }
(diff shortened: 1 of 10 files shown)


jglick:
> We'll also record what job properties, triggers, and
> parameters were defined in the Jenkinsfile for a given run, so that we
> can refer back in the next build to see what existing job properties,
> triggers, or parameters were defined through the Jenkinsfile vs
> defined outside of the Jenkinsfile.

Why not do this in the `properties` step generally?

abayer:
@jglick Good question. Lemme think on that for a day or so.

i386:
@abayer any progress on this one?

abayer:
@jglick So I think something similar to this for `properties` does make sense, but I think there are two reasons to do this here separately as well: handling triggers and parameters separately, since they're distinct in Declarative syntax, and avoiding actually invoking the `properties` step (and thereby confusing Blue Ocean) when we really don't need to.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -735,4 +743,249 @@ public class Utils {
...
+     * @param toInstantiate The list of either instances of the class or {@link UninstantiatedDescribable}s that can be
+     * instantiated to instances of the class.
+     * @return The list of instances. May be empty.
+     */
+    @Nonnull
+    private static <T extends Describable> List<T> instantiateList(Class<T> clazz, List<Object> toInstantiate) {
+        List<T> l = []
+        toInstantiate.each { t ->
+            if (t instanceof UninstantiatedDescribable) {
+                l.add((T) t.instantiate())
```
jglick: I think you meant to pass `clazz` as an argument. Otherwise the cast is unsafe or you might get an `UnsupportedOperationException` about ambiguous symbols.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/JobPropertyTrackerAction.java
```diff
@@ -0,0 +1,95 @@
...
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Invisible action used for tracking what {@link JobProperty}s, {@link Trigger}s, and {@link ParameterDefinition}s were
+ * defined in the Jenkinsfile for a given run.
```
jglick: I have a problem with this because it is potentially complicating the migration to a situation where the `properties` step is handling this issue more generally.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/JobPropertyTrackerAction.java
```diff
@@ -0,0 +1,95 @@
...
+    private final Set<String> jobProperties = new HashSet<>();
+    private final Set<String> triggers = new HashSet<>();
+    private final Set<String> parameters = new HashSet<>();
+
+    public JobPropertyTrackerAction(@CheckForNull List<JobProperty> rawJobProperties,
+                                    @CheckForNull List<Trigger> rawTriggers,
+                                    @CheckForNull List<ParameterDefinition> rawParameters) {
+        if (rawJobProperties != null) {
+            for (JobProperty p : rawJobProperties) {
+                jobProperties.add(p.getClass().getName());
```
jglick: Should be using

```java
jobProperties.add(p.getDescriptor().getId());
```

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/JobPropertyTrackerAction.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+        if (rawTriggers != null) {
+            for (Trigger t : rawTriggers) {
+                triggers.add(t.getClass().getName());
```
jglick: ditto

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -735,4 +743,249 @@ public class Utils {
...
+                previousTriggers.addAll(previousAction.getTriggers())
+                previousParameters.addAll(previousAction.getParameters())
+            }
+        }
+
+        List<JobProperty> jobPropertiesToApply = []
+        Set<Class<? extends JobProperty>> seenClasses = new HashSet<>()
+        if (rawJobProperties != null) {
+            jobPropertiesToApply.addAll(rawJobProperties)
+            seenClasses.addAll(rawJobProperties.collect { it.class })
```
jglick:
```groovy
it.descriptor.id
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -735,4 +743,249 @@ public class Utils {
...
+     */
+    @Nonnull
+    private static List<Trigger> getTriggersToApply(@CheckForNull List<Trigger> newTriggers,
+                                                    @Nonnull List<Trigger> existingTriggers,
+                                                    @Nonnull Set<String> prevDefined) {
+        Set<Class<? extends Trigger>> seenTriggerClasses = new HashSet<>()
+        List<Trigger> toApply = []
+        if (newTriggers != null) {
+            toApply.addAll(newTriggers)
+            seenTriggerClasses.addAll(newTriggers.collect { it.class })
```
jglick: ditto

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -735,4 +743,249 @@ public class Utils {
...
+        }
+
+        // Find all existing triggers that aren't of classes we've explicitly defined, *and* aren't
+        // in the set of classes of triggers defined by the Jenkinsfile in the previous build. Add those too.
+        toApply.addAll(existingTriggers.findAll {
+            !(it.class in seenTriggerClasses) && !(it.class.name in prevDefined)
```
jglick: ditto

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ParametersTest.java
```diff
@@ -74,4 +79,32 @@ public void paramsRemoved() throws Exception {
...
+
+        List<ParameterDefinition> newParams = new ArrayList<>();
+        newParams.addAll(paramProp.getParameterDefinitions());
+        newParams.add(new StringParameterDefinition("DO_NOT_DELETE", "something"));
+        job.removeProperty(paramProp);
+        job.addProperty(new ParametersDefinitionProperty(newParams));
+
+        job.setDefinition(new CpsFlowDefinition(pipelineSourceFromResources("propsTriggersParamsRemoved"), true));
+        WorkflowRun b2 = job.scheduleBuild2(0).waitForStart();
+        j.assertBuildStatusSuccess(j.waitForCompletion(b2));
```
jglick:
```java
WorkflowRun b2 = j.buildAndAssertSuccess(job);
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Oh yeaaaaah, good catch.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -735,4 +743,249 @@ public class Utils {
...
+     * instantiated to instances of the class.
+     * @return The list of instances. May be empty.
+     */
+    @Nonnull
+    private static <T extends Describable> List<T> instantiateList(Class<T> clazz, List<Object> toInstantiate) {
+        List<T> l = []
+        toInstantiate.each { t ->
+            if (t instanceof UninstantiatedDescribable) {
+                l.add((T) t.instantiate())
+            } else if (clazz.isInstance(t)) {
```
abayer: Good point. Originally I had this check ahead of the `UninstantiatedDescribable` check, so it made sense, but that didn't work at all for whatever reason, so I flipped them and didn't change this. Fixing!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: *nod*

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ParametersTest.java
abayer: Yarp.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/JobPropertyTrackerAction.java
abayer: Like I said above, my reasoning for doing this here rather than in `properties` is, primarily, that we're handling `Trigger`s and `ParameterDefinition`s rather than `PipelineTriggersJobProperty` and `ParametersDefinitionProperty`, while `properties` is, well, just handling the `JobProperty`s themselves, not their contents. I'm wary of crossing the streams, as it were.

abayer: Oh, and compatibility with `properties` isn't really relevant, since we block use of `properties` in Declarative - any interaction with `JobProperty`s is going to be mediated by our code in the first place, so if we end up deciding down the road to do something similar to this just for `JobProperty`s in the `properties` step, we can make this logic here aware of the `properties`-specific tracking logic and subsume it, etc. 

That said, I am going to rename the action here in case of overlap in the future.

PR #161, opened 2017-06-11
[FIXED JENKINS-44809] Clean up existing job property handling.

abayer:
* JENKINS issue(s):
    * [JENKINS-44809](https://issues.jenkins-ci.org/browse/JENKINS-44809)
* Description:
    * Two aspects - stop re-adding existing job properties defined outside of the Jenkinsfile, since they can't have changed, and remove all existing instances of a job property class defined inside the Jenkinsfile before we add the new instance, to avoid duplication.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -826,7 +826,7 @@ public class Utils {
                     // instance of the class that we've seen so far. Ideally we'd be ignoring it completely, but due to
                     // JENKINS-44809, we've created situations where tons of duplicate job property instances exist,
                     // which need to be nuked, so go through normal cleanup.
-                    if (!jobPropertiesToApply.any { p.class.isInstance(it) }) {
+                    if (!jobPropertiesToApply.any { p.descriptor == it.descriptor }) {
                         jobPropertiesToApply.add(p)
                     }
                 } else {
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -813,18 +813,14 @@ public class Utils {
             jobPropertiesToApply.addAll(rawJobProperties)
             seenClasses.addAll(rawJobProperties.collect { it.descriptor.id })
         }
-        // Find all existing job properties that aren't of classes we've explicitly defined, *and* aren't
-        // in the set of classes of job properties defined by the Jenkinsfile in the previous build. Add those too.
-        // Oh, and ignore the PipelineTriggersJobProperty and ParameterDefinitionsProperty - we handle those separately.
-        // And stash the property classes that should be removed aside as well.
+        // Find all existing job properties that aren't of classes we've explicitly defined, *and* are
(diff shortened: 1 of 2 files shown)


abayer:
@jglick - Any suggestions on how to reproduce the original bug would be appreciated, but I think it may be impossible to do so in a unit test...

abayer:
Ok, thanks to a later comment I know for sure what the root problem is (duplicate `JobProperty` instances) so testing is a bit easier.

abayer:
@jglick FYI, I'm going to have two follow-ups to this - moving the non-triggers/parameters logic into `properties`, and then moving this to use that (with some special sauce for triggers/parameters) instead of its own implementation. But I wanted to get this bug fixed first.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -822,8 +822,13 @@ public class Utils {
             // We only care about classes that we haven't already seen in the new properties list.
             if (!(p.descriptor.id in seenClasses)) {
                 if (!(p.descriptor.id in previousProperties)) {
-                    // This means it's a job property defined outside of our scope, so leave it there.
-                    jobPropertiesToApply.add(p)
+                    // This means it's a job property defined outside of our scope, so retain it, if it's the first
+                    // instance of the class that we've seen so far. Ideally we'd be ignoring it completely, but due to
+                    // JENKINS-44809, we've created situations where tons of duplicate job property instances exist,
+                    // which need to be nuked, so go through normal cleanup.
+                    if (!jobPropertiesToApply.any { p.class.isInstance(it) }) {
```
jglick:
```groovy
p.descriptor == it.descriptor
```

is what I think you mean.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -854,6 +859,11 @@ public class Utils {
 
             // Now add all the other job properties we know need to be added.
             jobPropertiesToApply.each { p ->
+                // Remove the existing instance(s) of the property class before we add the new one. We're looping and
+                // removing multiple to deal with the results of JENKINS-44809.
+                while (j.getProperty(p.class) != null) {
+                    j.removeProperty(p.class)
```
jglick: OK for now though these methods in `Job` should actually be deprecated and replaced by ones which operate on `Descriptor.id`.

abayer: Agreed wholeheartedly. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Ah, yeah, that should work. Trying it.

abayer: And done.

abayer:
Merging/releasing once the tests pass, just to be safe.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: ...and they should probably remove all instances with the given `Descriptor.id` rather than just the first...

PR #162, opened 2017-06-13
Minor cleanup of job property cleanup changes.

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * Shouldn't actually matter, but makes me feel a bit more comfortable.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -797,7 +797,7 @@ public class Utils {
         Set<String> previousParameters = new HashSet<>()
 
         DeclarativeJobPropertyTrackerAction previousAction = null
-        WorkflowRun previousBuild = r.getPreviousBuild()
+        WorkflowRun previousBuild = r.getPreviousCompletedBuild()
```
jglick: Careful—tracking may not work correctly with concurrent builds.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -877,7 +877,7 @@ public class Utils {
         } finally {
             bc.abort();
             // Roll back and use the same action tracking as last build, if any.
-            if (previousAction != null) {
+            if (r.getAction(DeclarativeJobPropertyTrackerAction.class) == null && previousAction != null) {
```
jglick: Is there a testable scenario for this?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: So that's a good question that I'd love your input on - I've got something similar at https://github.com/jenkinsci/workflow-multibranch-plugin/pull/65 as well. I honestly can't decide whether we should look at the last completed build or the last build regardless. In the `properties` step side, I have a check for whether `properties` had been run in the previous build (for reasons I explain there) so I thought completed would be better and made it the same here for consistent behavior, but I'm legitimately not sure which would be best.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Actually, yeah, I think so. Working on it.

abayer: Test added - making sure we don't have multiple `DeclarativeJobPropertyTrackerAction`s.

abayer:
@jglick Refactored based on input over on https://github.com/jenkinsci/workflow-multibranch-plugin/pull/65 to put the action on the job instead.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -796,16 +796,21 @@ public class Utils {
...
-        DeclarativeJobPropertyTrackerAction previousAction = null
-        WorkflowRun previousBuild = r.getPreviousBuild()
-        if (previousBuild != null) {
-            previousAction = previousBuild.getAction(DeclarativeJobPropertyTrackerAction.class)
-            if (previousAction != null) {
-                previousProperties.addAll(previousAction.getJobProperties())
-                previousTriggers.addAll(previousAction.getTriggers())
-                previousParameters.addAll(previousAction.getParameters())
+        // First, use the action from the job if it's present.
+        DeclarativeJobPropertyTrackerAction previousAction = j.getAction(DeclarativeJobPropertyTrackerAction.class)
```
jglick: What about 1.1.6 users?

jglick: (FTR this is why I recommended waiting to develop the `workflow-multibranch` fix before pushing anything out—to avoid needing to think about multiple upgrade scenarios.)

abayer: See right below, where we fall back to the previous completed build if the action isn't present on the job.

abayer: And yeah, I understand, but the thing is that this was always going to have be different to deal with `Trigger`s and `ParameterDefinition`s, so I'm not sure if we'll be able to actually use `properties` directly. I think I will end up using `JobPropertyTrackerAction.getJobPropertyDescriptors()` (and create/replace `JobPropertyTrackerAction` itself) from here, but never execute `properties`. Anyway, needs more thought.

jglick: And even if `pipeline-model-definition` were to reuse an API from `workflow-multibranch` and the changes released simultaneously, there would be some people who would update `workflow-multibranch` first, so we always have the possibility of a two-stage upgrade.

What is the deal with `Trigger`s and `ParameterDefinition`s? My understanding of Declarative was that these were kept in dedicated sections rather than `options` for UX reasons but internally it was all boiling down to the same `properties` call. Am I missing some subtlety?

abayer: The `Trigger`s and `ParameterDefinition`s are defined via their `JobProperty`s, yes, but I need to be able to say "Ok, which `Trigger`s were defined in the `Jenkinsfile` vs which ones were defined in the UI", which is a level of granularity that doesn't function with `properties`. I couldn't find a way to do that with just `JobPropertyTrackerAction`, and I don't feel like it makes sense to introspect the properties passed to `properties` for `PipelineTriggersJobProperty` and `ParametersDefinitionProperty` and their contents, etc...

jglick:
> which `Trigger`s were defined in the `Jenkinsfile` vs which ones were defined in the UI

Wait, you allow these to be mixed? Seems a bit crazy. I would say that either you define all triggers in the script, or none. Same for parameters.

abayer:
The specific case that prompted all this insanity was around UI triggers...but yeah, not sure if they needed that mix. Lemme think on it

abayer:
Yeah, screw it. That's crazy. New PR incoming. =)

abayer:
Actually, unscrew it. I'm going to leave this all as is for the next few weeks rather than rewriting the world every few days. What we've got now is functional, if imperfect.

PR #163, opened 2017-06-14
Work around PluginFirstClassLoader+GroovyClassLoader problem

abayer:
* JENKINS issue(s):
    * Related to and a workaround for [JENKINS-44898](https://issues.jenkins-ci.org/browse/JENKINS-44898)
* Description:
    * Note that this may be too insane and/or goofy to actually do. =)
    * Relevant problem discussed at https://github.com/jenkinsci/kubernetes-plugin/pull/127.
    * `PluginFirstClassLoader` doesn't have a functional `findResource` method, which for some reason I have yet to determine breaks `GroovyClassLoader.loadClass` when trying to fetch a groovy file from a `PluginFirstClassLoader` via the `UberClassLoader`. This works around that problem by catching the resulting `ClassNotFoundException` and trying to instead use `getResource` to get the URL for the groovy file and then parse it directly.
    * Also whitelists the fields on `WithScriptScript` so they can be accessed by children outside of these plugins properly.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
@@ -61,14 +61,10 @@ public abstract class WithScriptDescribable<T extends WithScriptDescribable<T>> 
         } catch (ClassNotFoundException e) {
             // This is special casing to deal with PluginFirstClassLoaders, which don't have a functional findResource method.
             // That results in GroovyClassLoader.loadClass failing to find resources to parse and load.
-            PluginWrapper pw = Jenkins.getInstance().getPluginManager().whichPlugin(getDescriptor().getClass());
-            if (pw != null) {
-                URL res = pw.classLoader.getResource(getDescriptor().getScriptClass().replace('.', '/') + ".groovy");
-                if (res != null) {
-                    clz = ((GroovyClassLoader) cpsScript.getClass().getClassLoader()).parseClass(new GroovyCodeSource(res));
-                }
-            }
-            if (clz == null) {
+            URL res = getDescriptor().getClass().getClassLoader().getResource(getDescriptor().getScriptClass().replace('.', '/') + ".groovy");
+            if (res != null) {
+                clz = ((GroovyClassLoader) cpsScript.getClass().getClassLoader()).parseClass(new GroovyCodeSource(res));
+            } else {
                 throw e;
             }
         }
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
@@ -24,11 +24,16 @@
 
 package org.jenkinsci.plugins.pipeline.modeldefinition.withscript;
 
+import groovy.lang.GroovyClassLoader;
+import groovy.lang.GroovyCodeSource;
+import hudson.PluginWrapper;
 import hudson.model.AbstractDescribableImpl;
+import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.cps.CpsScript;
 import org.jenkinsci.plugins.workflow.cps.CpsThread;
 
 import java.io.Serializable;
+import java.net.URL;
 
 /**
  * Implementations for {@link WithScriptDescriptor} - pluggable script backends for Declarative Pipelines.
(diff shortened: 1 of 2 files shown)


pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
```diff
@@ -50,10 +55,24 @@ public WithScriptScript getScript(CpsScript cpsScript) throws Exception {
...
-                .getConstructor(CpsScript.class, this.getClass())
+        Class clz = null;
+        try {
+            clz = cpsScript.getClass().getClassLoader().loadClass(getDescriptor().getScriptClass());
+        } catch (ClassNotFoundException e) {
+            // This is special casing to deal with PluginFirstClassLoaders, which don't have a functional findResource method.
+            // That results in GroovyClassLoader.loadClass failing to find resources to parse and load.
+            PluginWrapper pw = Jenkins.getInstance().getPluginManager().whichPlugin(getDescriptor().getClass());
+            if (pw != null) {
+                URL res = pw.classLoader.getResource(getDescriptor().getScriptClass().replace('.', '/') + ".groovy");
```
jglick: `Jenkins.instance.pluginManager.whichPlugin(descriptor.class).classLoader` seems a roundabout way of saying `descriptor.class.classLoader` unless I am missing something?

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
```diff
@@ -50,10 +55,24 @@ public WithScriptScript getScript(CpsScript cpsScript) throws Exception {
...
+                if (res != null) {
+                    clz = ((GroovyClassLoader) cpsScript.getClass().getClassLoader()).parseClass(new GroovyCodeSource(res));
```
jglick: This cannot return null, can it? So, assuming the simplification above to not bother with `PluginWrapper`, the `throw` should be in an `else` clause (`res == null`).

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
abayer: Hmm. Good point.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
abayer: That's right - good catch.

abayer:
Patch for core bug up at https://github.com/jenkinsci/jenkins/pull/2916 - the approach here on this PR will handle the failures until you're on the newer core version.

abayer:
@jglick Thoughts now that https://github.com/jenkinsci/jenkins/pull/2916 has been merged? I'd like to get this in as a bandaid for earlier core versions.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
```diff
@@ -50,10 +55,20 @@ public WithScriptScript getScript(CpsScript cpsScript) throws Exception {
...
 
-        return (WithScriptScript) cpsScript.getClass()
-                .getClassLoader()
-                .loadClass(getDescriptor().getScriptClass())
-                .getConstructor(CpsScript.class, this.getClass())
+        Class clz = null;
+        try {
+            clz = cpsScript.getClass().getClassLoader().loadClass(getDescriptor().getScriptClass());
+        } catch (ClassNotFoundException e) {
+            // This is special casing to deal with PluginFirstClassLoaders, which don't have a functional findResource method.
```
jglick:
```java
// TODO delete JENKINS-44898 workaround as of Jenkins 2.66
```

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptScript.java
```diff
@@ -24,12 +24,15 @@
 
 package org.jenkinsci.plugins.pipeline.modeldefinition.withscript;
 
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted;
 import org.jenkinsci.plugins.workflow.cps.CpsScript;
 
 import java.io.Serializable;
 
 public abstract class WithScriptScript<T extends WithScriptDescribable<T>> implements Serializable {
+    @Whitelisted
```
jglick: How is this related?

abayer: Once I got past the `PluginFirstClassLoader` problem, I discovered that `WithScriptScript.script` and `WithScriptScript.describable` weren't accessible by https://github.com/jenkinsci/kubernetes-plugin/blob/master/src/main/resources/org/csanchez/jenkins/plugins/kubernetes/pipeline/KubernetesDeclarativeAgentScript.groovy, so hey, whitelisting.

abayer: i.e., it's a flaw that was only uncovered after the main fix here and it made sense to just fold it into the same PR.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
abayer: Done.

PR #166, opened 2017-07-06
[FIXED JENKINS-45081] Don't error out for foo.pipeline { ... }

abayer:
* JENKINS issue(s):
    * [JENKINS-45081](https://issues.jenkins-ci.org/browse/JENKINS-45081)
* Description:
    * We were erroring out if you had a Scripted Pipeline with, say, a global variable `foo` with a method on it named `pipeline` that was being called like, say, `foo.pipeline { echo 'Hi there' }`. That...was not right. This fixes that. =)
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


abayer:
@rsandell It's actually kinda nifty - `matchMethodName` returns null if the `MethodCallExpression`'s being called on anything other than `this`!

PR #167, opened 2017-07-06
[JENKINS-42082] Verify that you can put TEST in an env var name

abayer:
* JENKINS issue(s):
    * [JENKINS-42082](https://issues.jenkins-ci.org/browse/JENKINS-42082)
* Description:
    * Just verifying that this isn't an actual bug, at least as of now.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/test/resources/envVarContainsTEST.groovy
+++ b/pipeline-model-definition/src/test/resources/envVarContainsTEST.groovy
@@ -37,6 +37,8 @@ pipeline {
             steps {
                 echo "TEST_VAR is ${TEST_VAR}"
                 echo "VAR_TEST is ${VAR_TEST}"
+                sh 'echo "TEST_VAR from shell is ${TEST_VAR}"'
+                sh 'echo "VAR_TEST from shell is ${VAR_TEST}"'
             }
         }
     }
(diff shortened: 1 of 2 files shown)


pipeline-model-definition/src/test/resources/envVarContainsTEST.groovy
```diff
@@ -0,0 +1,46 @@
...
+    agent {
+        label "some-label"
+    }
+
+    stages {
+        stage("foo") {
+            steps {
+                echo "TEST_VAR is ${TEST_VAR}"
+                echo "VAR_TEST is ${VAR_TEST}"
+            }
```
rsandell: Just to be safe, should add
```
  sh "echo SH_TEST_VAR is $TEST_VAR"
  sh "echo SH_VAR_TEST is $VAR_TEST"
```
or similar.

pipeline-model-definition/src/test/resources/envVarContainsTEST.groovy
```diff
@@ -37,6 +37,8 @@ pipeline {
             steps {
                 echo "TEST_VAR is ${TEST_VAR}"
                 echo "VAR_TEST is ${VAR_TEST}"
+                sh 'echo "TEST_VAR from shell is ${TEST_VAR}"'
+                sh 'echo "VAR_TEST from shell is ${VAR_TEST}"'
```
rsandell: Well, the `${}` are valid in bash but in this case unnecessary.
It might cause confusion that groovy is expected to expand it, at least I had to read it three times to get that :)

PR #168, opened 2017-07-06
[FIXED JENKINS-42338] Install root-level tools on stage agents.

abayer:
* JENKINS issue(s):
    * [JENKINS-42338](https://issues.jenkins-ci.org/browse/JENKINS-42338)
* Description:
    * If there's a top-level `tools` directive, make sure to install those tools on any `agent` specific to a `stage`. Override the root-level tools with any stage-level tool definitions of the same type.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Tools.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Tools.groovy
@@ -29,6 +29,7 @@ import hudson.tools.ToolDescriptor
 import org.jenkinsci.Symbol
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 
+import javax.annotation.CheckForNull
 import javax.annotation.Nonnull
 
 /**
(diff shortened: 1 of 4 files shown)


imod:
👍 love it!

oleg-nenashev:
@reviewbybees done

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Tools.groovy
```diff
@@ -57,6 +58,23 @@ public class Tools extends MappedClosure<String,Tools> implements Serializable {
     }
 
     /**
+     * Merges the tool entries from another instance into this one, defaulting to the current instance's values.
+     *
+     * @return A list of type/name tuples
+     */
+    public List<List<Object>> mergeToolEntries(@CheckForNull Tools other) {
```
rsandell: Since you annotate the nullability of the parameter you should annotate the return value as well. In this case `@Nonnull`. It confused me in the code below that you before expected something could be null, but now you seemed to know that it was non null.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -306,14 +306,20 @@ public class ModelInterpreter implements Serializable {
...
      * @param body The closure to execute
      * @return The return of the resulting executed closure
      */
-    def toolsBlock(Agent agent, Tools tools, Closure body) {
+    def toolsBlock(Agent agent, Tools tools, Root root = null, Closure body) {
+        def toolsList = []
+        if (tools != null) {
+            toolsList = tools.mergeToolEntries(root?.tools)
+        } else if (root?.tools != null) {
+            toolsList = root.tools.getToolEntries()
```
rsandell: Can't see if `getToolEntries()` is non null or not. But you seem to know that ;)

abayer: Fixed both of these with more annotations. =)

PR #170, opened 2017-07-06
[FIXED JENKINS-45098] Properly validate tool-without-version

abayer:
* JENKINS issue(s):
    * [JENKINS-45098](https://issues.jenkins-ci.org/browse/JENKINS-45098)
* Description:
    * We were still trying to do stuff with a variable after we determined it was null. That was dumb. Fixed.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


abayer:
@rsandell good idea - sadly, I don't know if we can do it against `ModelInterpreter` and friends, but some of it for sure...

PR #172, opened 2017-07-18
[FIXED JENKINS-45546] Add jacoco code coverage to Declarative builds

abayer:
* JENKINS issue(s):
    * [JENKINS-45546](https://issues.jenkins-ci.org/browse/JENKINS-45546)
* Description:
    * Note that we're not using the Jacoco plugin - it's not installed on ci.jenkins.io in the first place, and I think the real value is probably going to come from the XML file anyway, which can be slurped in by other tools.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,9 +35,11 @@ pipeline {
         }
         success {
             archive "**/target/*.hpi"
+            archive "**/target/site/jacoco/jacoco.xml"
         }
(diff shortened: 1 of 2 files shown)


kshultzCB:
Thanks for doing this, Andrew. :bee:

PR #173, opened 2017-07-23
WORK-IN-PROGRESS - DO NOT MERGE: [JENKINS-42753] Runtime model generation via AST transformation

abayer:
* JENKINS issue(s):
* Description:
    * Experimental work on generating `Root` and the rest of the runtime model via AST transformation, rather than runtime closure evaluation with some insane `script.evaluate(...)` calls tossed in for fun. 
    * This is close to functional, but contains a *ton* of debug statements, dead code, messy code, different ways of generating the relevant Groovy `ASTNode`s, etc...it's missing some test coverage as well. I'm going to be refactoring all the new stuff into as tight and reusable chunks of code as I can, and documenting the hell out of how it works before we proceed. And I'm going to squash this down to get rid of the giant pile of failed experiment commits too.
    * The key value of this change is not anything *new* (though it will re-enable references to variables and functions defined in the `Jenkinsfile` outside of the `pipeline { ... }` block from within `environment` variables and `when { expression { ... } }`s. Rather, what matters is that it's going to get rid of the various chunks of code for runtime translation currently held together with spit and baling wire, replacing them with a coherent, documented, maintainable framework. Also it'll cut a second or two off the startup/run time of Declarative builds, but that's just a pleasant side effect.
    * All existing syntax will continue to be valid. It's *possible* I may miss some expressions that need to be translated in some edge cases, in which case that'll get noticed and thrown as an informative error message asking for a ticket to be opened to add support. 
    * Unless I'm persuaded otherwise, we will *not* be guaranteeing support for running builds at upgrade time. Most things will work fine, but some (job properties, triggers, parameter definitions, `skipCheckout` and `skipUnstable` options, `when` conditions, `environment` and `credentials` that haven't been evaluated yet, probably more) will not. The runtime model has to change some to support the new approach, and, well, them's the breaks. We'll throw up compatibility warnings on upgrade, and I expect to get some hassle over this, but I simply don't think it's worth the effort of adding a compatibility shim just for the running jobs. 
* Documentation changes:
    * God willing, the only thing users will care about here is the running builds issue.
* Users/aliases to notify:
    * No one yet.


abayer:
I've moved the `transform*` methods into a new class, `RuntimeASTTransformer`, rather than each being in `Root`, `Agent`, etc, and think I'll be changing from the `new AstBuilder().buildFromSpec { ... }` approach to using the convenience methods in `GeneralUtils` (like `closureX(...)`) and straight-up `new WhateverExpression(...)` when needed. Consistency, simpler, easier to move to Java if we so choose, etc.

abayer:
Replaced by #174

PR #174, opened 2017-07-25
[FIXED JENKINS-42753] Use AST transformation to generate runtime model

abayer:
* JENKINS issue(s):
* Description:
    * There've been any number of problems in the past due to the closure translation approach for going from `pipeline { ... }` to the `Root` object we need to actually run. I won't bother going into them here, just...yeah. There's lots. Until this change goes in, there are 4 special CPS-transformed classes for the translation, plus truly bizarre special-casing logic for handling environment and when expressions via looping circuits of script.evaluate calls. I'm serious. It's sad.
    * This replaces all that by instead generating Root and its children via an AST transformation at parsing/validation time. It gets us away from ever having to worry about translating or evaluating within the CPS code, using lazily evaluated closures to support things like declaring cross-referencing environment variables out of order, or environment variables depending on credentials variables, etc...
    * Anyway, it works. It's more maintainable. It's cleaner. It's actually very slightly faster (since we've eliminated the closure translation stuff from CPS execution, and got rid of a now-redundant additional parse/validate run to get the execution model). And it's completely compatible syntax-wise (hopefully! It's not impossible that I missed some weird expressions in environment variable values, but we'll cross that bridge if we get to it).
    * *NOTES FOR REVIEWERS*
        * For some reason, the following files were "renamed" and had their content completely replaced, rather than removed and new files added. So ignore the deletions, and generally don't worry about them much in the first place since they're just test `Jenkinsfile`s:
            * `ClosureContentsChecker.groovy` -> `backslashReductionInEnv.groovy`
            * `MethodMissingWrapper.groovy` -> `backslashesFromExistingVar.groovy`
            * `PropertyOptionContainer.groovy` -> `credentialsUsedInWhenEnv.groovy`
            * `LibrariesTranslator.groovy` -> `credentialsUsedInWhenExpression.groovy`
            * `MethodsToList.groovy` -> `additionalInvalidExpressionsInEnvironment.groovy`
            * `StepBlockWithOtherArgs.groovy` - `errors/multipleStageLevelSections.groovy`
            * `PropertiesToMap.groovy` -> `stagePost.groovy`

* Documentation changes:
    * This will need some form of notification that currently-running builds at upgrade time may not work - any new build will work fine, but I've opted to not add a compatibility shim to the runtime logic to handle some fairly significant changes in the model classes.
* TODO:
  - [x] Tests of outside-of-`pipeline`-defined variables and functions being referenced in `environment` or `when { expression { ... } }`
  - [x] More thorough testing of `environment` and `credentials` variables inside `when { expression { ... } }`, just to be safe.
  - [x] Cleanup any kruft that may remain in re: dead code.
  - [x] Expanded coverage of possible `environment` expressions.
  - [x] Developer documentation on the Declarative AST interactions/transformations.
  - [x] Set `compatibleSinceVersion` properly.
* Users/aliases to notify:
    * @reviewbybees


abayer:
Still needs some expansion of the dev docs, but I think this is ready for review.

rsandell:
Whoa

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
```diff
@@ -58,73 +57,32 @@ public class Root implements NestedModel, Serializable {
...
      */
-    List<String> getEnvVars(CpsScript script) {
+    Map<String,Closure> getEnvVars(CpsScript script) {
         if (environment != null) {
-            return environment.resolveEnvVars(script, true).findAll {
-                it.key in environment.getMap().keySet()
-            }.collect { k, v ->
-                "${k}=${v}"
-            }
+            environment.envResolver.setScript(script)
```
rsandell: from a first scan over the class this line does not seem to be needed, it doesn't seem to do anything in particular for the `closureMap`.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
```diff
@@ -54,67 +55,31 @@ public class Stage implements NestedModel, Serializable {
...
      */
-    List<String> getEnvVars(Root root, CpsScript script, Stage parentStage = null) {
+    Map<String,Closure> getEnvVars(CpsScript script) {
         if (environment != null) {
-            return environment.resolveEnvVars(script, true, root.environment, parentStage).findAll {
-                it.key in environment.getMap().keySet()
-            }.collect { k, v ->
-                "${k}=${v}"
-            }
+            environment.envResolver.setScript(script)
```
rsandell: same here?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -1157,98 +1170,17 @@ class ModelParser implements Parser {
...
-    /**
-     * Works like a regular Java cast, except if the value doesn't match the type, return null
-     * instead of throwing an exception.
-     */
-    protected <X> X castOrNull(Class<X> type, Object value) {
-        if (type.isInstance(value))
-            return type.cast(value);
-        return null;
-    }
+    public static final boolean AST_DEBUG_LOGGING = SystemProperties.getBoolean(ModelParser.class.getName()+".astDebugLogging")
```
rsandell: `|| isHpiRun()` ?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
abayer: I'll double check, but I'm 99% certain (on first waking up) that this is needed - the various closures in that map can need it to be set to call `getScriptPropOrParam`

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: Worthy addition.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -0,0 +1,759 @@
...
+    Expression transformOptions(@CheckForNull ModelASTOptions original) {
+        if (isGroovyAST(original) && !original.options.isEmpty()) {
+            List<ModelASTOption> jobProps = new ArrayList<>()
+            List<ModelASTOption> options = new ArrayList<>()
+            List<ModelASTOption> wrappers = new ArrayList<>()
+
+            SymbolLookup symbolLookup = SymbolLookup.get()
+
+            original.options.each { o ->
+                if (symbolLookup.findDescriptor(JobProperty.class, o.name) != null) {
```
rsandell: shouldn't there be some blacklisting of stuff here or is that done earlier?

abayer: Yeah, that's done in `ModelValidatorImpl` - we never get here if there are invalid `JobProperty`s etc.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: ...is that an actual method I can call? If so, where is it? =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -534,17 +502,37 @@ public class ModelInterpreter implements Serializable {
...
+     * @param instanceType
+     * @return A list of instances
+     */
+    private <Z> List<Z> instancesFromClosure(Closure rawClosure, Class<Z> instanceType) {
+        rawClosure.delegate = script
+        rawClosure.resolveStrategy = Closure.DELEGATE_FIRST
+
+        List<Z> instanceList = []
+
+        rawClosure.call().each { inst ->
```
rsandell: A bit more groovy way I guess would be

```groovy
return rawClosure.call().collectMany { inst ->
    if (instanceType.isInstance(inst)) {
                return [instanceType.cast(inst)]
    } else {
        return []
    }
}
```

pipeline-model-definition/src/test/resources/errors/additionalInvalidExpressionsInEnvironment.groovy
```diff
@@ -22,24 +22,30 @@
...
  */
 
+pipeline {
+    environment {
+        FOO =
+            "B" +
+                "A" -
+                "R"
+        BAR = 1 + credentials("foo")
+        _UNDERSCORE = "VALID" + "FOO" ?: "BAR"
```
rsandell: looks weird, if the literal is not empty use it?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
rsandell: http://javadoc.jenkins.io/hudson/Main.html#isDevelopmentMode

But is has been iffy for me in the past, so I think I looked for some maven thing in `System.getProperty()` or something. I'll see what I can find.

rsandell: https://github.com/jenkinsci/dockerhub-notification-plugin/blob/master/src/main/java/org/jenkinsci/plugins/registry/notification/webhook/dockerhub/DockerHubWebHook.java#L51

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AnyOfConditional.java
```diff
@@ -48,5 +52,10 @@ public AnyOfConditional(List<DeclarativeStageConditional<? extends DeclarativeSt
         public int getAllowedChildrenCount() {
             return -1;
         }
+
+        @Override
+        public Expression transformToRuntimeAST(@CheckForNull ModelASTWhenContent original) {
```
abayer: `ASTParserUtils#transformWhenContentToRuntimeAST(ModelASTWhenContent)` is in `pipeline-model-definition`, so it's not available to `pipeline-model-extensions`. But I do need to think about that some.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: I always forget `collectMany`!

pipeline-model-definition/src/test/resources/errors/additionalInvalidExpressionsInEnvironment.groovy
abayer: This is literally just here to blow up on the `?:` elvis expression - the content is kinda irrelevant.

rsandell: But it is definitely not empty.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: Ah, found `hudson.hpi.run` system property.

pipeline-model-definition/src/test/resources/errors/additionalInvalidExpressionsInEnvironment.groovy
abayer: Doesn't matter - we fail at validation, before we'd ever execute it.

rsandell: But it looks very deliberate, so I got stuck on it for a while reading it :)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
rsandell: Me too, I had to look it up :p

rsandell:
:bee:

Do you need more reviews or can this be merged soon?

abayer:
Ok, I'm gonna merge it. =)

stephendonner:
ship it squirrel

PR #175, opened 2017-08-01
Switch to actually calling stage('foo') for parallel stages

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * I discovered that this won't actually mess with Blue Ocean visualization, since Blue Ocean already discards stages inside parallel branches but keeps the steps inside said stages. So...let's simplify things.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -148,7 +148,6 @@ public class ModelInterpreter implements Serializable {
         return {
             script.stage(thisStage.name) {
                 try {
-                    // NOTE - this will switch to script.stage in the future.
                     if (firstError != null) {
                         Utils.logToTaskListener("Stage '${thisStage.name}' skipped due to earlier failure(s)")
                         Utils.markStageSkippedForFailure(thisStage.name)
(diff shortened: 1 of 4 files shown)


abayer:
@kzantow @vivek - I cc'd you on this to make sure that I'm not glaringly wrong from the Blue Ocean side. I checked `PipelineNodeTest` there, which has a nested stages test, and this behavior should match what's expected there.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -150,66 +146,68 @@ public class ModelInterpreter implements Serializable {
...
-                                    script.parallel(parallelStages)
-                                }
-                            }
-                        } else {
-                            Utils.logToTaskListener("Stage '${thisStage.name}' skipped due to when conditional")
-                            Utils.markStageSkippedForConditional(thisStage.name)
-                        }
+            script.stage(thisStage.name) {
+                try {
+                    // NOTE - this will switch to script.stage in the future.
```
rsandell: The future is now!

abayer:
@vivek I'll do a SNAPSHOT deploy and open a PR against Blue Ocean using that SNAPSHOT to go through the tests/ATH there.

abayer:
Opened https://github.com/jenkinsci/blueocean-plugin/pull/1308 using the deployed SNAPSHOT.

abayer:
One legit failure in Blue Ocean so far - we're not marking the parallel branch above the nested stage as skipped etc, so that needs to be added. There were also some step counts that needed to be fixed now that the `properties` step isn't always executed.

vivek:
Sounds good, thanks @abayer

abayer:
@vivek Ok, think I got it - I tweaked things to propagate the `StageStatus` tags from the nested `stage` inside the `parallel` branch to the `parallel` branch too, and added logic so that if a `stage` containing nested `stage`s is skipped due to failure/unstable/conditional, we'll still "run" its child stages but mark them as skipped for the same reason (just as we would with a non-nested stage). I updated the Blue Ocean PR too.

abayer:
@vivek Since Blue Ocean looks good with that PR, is this good to go from your perspective? If so, I'll merge this, update #174 on top of master, do a new SNAPSHOT deploy, and update the Blue Ocean PR to use that SNAPSHOT so that it's even closer to what'll actually be Declarative 1.2.

PR #176, opened 2017-08-02
[FIXED JENKINS-45955] Switch to pulling blacklisted steps from extensions

abayer:
* JENKINS issue(s):
    * [JENKINS-45955](https://issues.jenkins-ci.org/browse/JENKINS-45955)
* Description:
    * Moves blacklisted steps to extensions rather than hardcoding in one place.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
@@ -36,11 +36,17 @@ import java.util.Map;
 public abstract class DeclarativeBlockedSteps implements ExtensionPoint {
     /**
      * Steps which are blocked in "steps", "post" conditions, and the like.
+     *
+     * @return A map of step names (which can also be function or method names that aren't actual steps) to strings
+     * explaining the reason for the step being blocked.
      */
     public abstract Map<String,String> blockedInSteps();
 
     /**
      * Steps which are blocked in method calls like "options" entries.
+     *
+     * @return A map of step names (which can also be function or method names that aren't actual steps) to strings
+     * explaining the reason for the step being blocked.
      */
     public abstract Map<String,String> blockedInMethodCalls();
 
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
@@ -50,7 +50,7 @@ public abstract class DeclarativeBlockedSteps implements ExtensionPoint {
     public static Map<String,String> allBlockedInSteps() {
         Map<String,String> map = new LinkedHashMap<>();
 
-        for (DeclarativeBlockedSteps dbs : ExtensionList.lookup(DeclarativeBlockedSteps.class)) {
+        for (DeclarativeBlockedSteps dbs : ExtensionList.lookup(DeclarativeBlockedSteps.class).reverseView()) {
             map.putAll(dbs.blockedInSteps());
         }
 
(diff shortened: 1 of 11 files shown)


pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public abstract Map<String,String> blockedInMethodCalls();
+
+    /**
+     * Get all blocked-in-steps steps across all extensions.
+     */
+    public static Map<String,String> allBlockedInSteps() {
+        Map<String,String> map = new LinkedHashMap<>();
+
+        for (DeclarativeBlockedSteps dbs : ExtensionList.lookup(DeclarativeBlockedSteps.class)) {
```
rsandell: I think you should traverse the list in reverse so that we are somewhat consistent with how ordinals normally work i.e. higher ordered extensions override lower sorted ones.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
```diff
@@ -0,0 +1,72 @@
...
+        return map;
+    }
+
+    /**
+     * Get all blocked-in-method-calls steps across all extensions.
+     */
+    public static Map<String,String> allBlockedInMethodCalls() {
+        Map<String,String> map = new LinkedHashMap<>();
+
+        for (DeclarativeBlockedSteps dbs : ExtensionList.lookup(DeclarativeBlockedSteps.class)) {
```
rsandell: same here

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
abayer: Will do, though it doesn't make a big difference here since the only thing that can be overridden is the message.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Map;
+
+/**
+ * Extension point for providing steps, and messages, which are blocked in Declarative.
+ */
+public abstract class DeclarativeBlockedSteps implements ExtensionPoint {
+    /**
+     * Steps which are blocked in "steps", "post" conditions, and the like.
+     */
+    public abstract Map<String,String> blockedInSteps();
```
jglick: You need to document what the keys and values are. I would never have guessed had I not read sources of an implementation.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Steps which are blocked in method calls like "options" entries.
+     */
+    public abstract Map<String,String> blockedInMethodCalls();
```
jglick: Rather than only allowing an implementation to block some steps globally, these methods should accept a `FlowExecution` or something, allowing certain steps to be blocked in certain places but not others.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
abayer: Good point.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeBlockedSteps.java
abayer: Interesting thought. What sort places are you thinking of?

rsandell: pipelines in the folder `admin-secret-hideyhole` can run the steps but not others?

jglick: Something along those lines. Hardly any extra effort to add this flexibility.

abayer: Sounds good to me. Adding.

abayer:
Subsumed into #179

PR #177, opened 2017-08-04
[FIXED JENKINS-45198] When possible, add scm vars to environment

abayer:
* JENKINS issue(s):
    * [JENKINS-45198](https://issues.jenkins-ci.org/browse/JENKINS-45198)
* Description:
    * This will not result in any change when run on core versions <2.60 (or with workflow-scm-step <2.6, git <3.3.1, etc), but will Just Work if you've got the right core/etc versions installed.
    * BasicModelDefTest#scmEnvVars will start failing when we bump core etc dependencies for this plugin directly.
* Documentation changes:
    * Honestly not sure if there should be doc changes - @bitwiseman, thoughts?
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
@@ -57,7 +57,6 @@ import org.jenkinsci.plugins.workflow.libs.SCMSourceRetriever;
 import org.jenkinsci.plugins.workflow.pipelinegraphanalysis.GenericStatus;
 import org.jenkinsci.plugins.workflow.pipelinegraphanalysis.StatusAndTiming;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.jvnet.hudson.test.Issue;
 
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
@@ -57,6 +57,7 @@ import org.jenkinsci.plugins.workflow.libs.SCMSourceRetriever;
 import org.jenkinsci.plugins.workflow.pipelinegraphanalysis.GenericStatus;
 import org.jenkinsci.plugins.workflow.pipelinegraphanalysis.StatusAndTiming;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.jvnet.hudson.test.Issue;
 
@@ -853,5 +854,18 @@ public class BasicModelDefTest extends AbstractModelDefTest {
                 .go();
 
 
+    }
+
+    @Issue("JENKINS-45198")
+    @Test
+    public void scmEnvVars() throws Exception {
+        // The change to support checkout scm returning a map and that map being added to the environment works fine with
+        // older core etc, but just doesn't do anything, since checkout scm isn't returning anything yet. But with newer
+        // core, etc, it'll Just Work.
+        expect("scmEnvVars")
+                // TODO: switch to .logNotContains("GIT_COMMIT is null") once we've moved to core 2.60+,
+                // workflow-scm-step 2.6+, git 3.3.1+
+                .logContains("GIT_COMMIT is null")
+                .go();
     }
 }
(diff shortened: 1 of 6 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -57,6 +57,7 @@
 import org.jenkinsci.plugins.workflow.pipelinegraphanalysis.GenericStatus;
 import org.jenkinsci.plugins.workflow.pipelinegraphanalysis.StatusAndTiming;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
```
jglick: Unused?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -854,4 +855,17 @@ public void parallelStagesAgentEnvWhen() throws Exception {
...
+
+    @Issue("JENKINS-45198")
+    @Test
+    public void scmEnvVars() throws Exception {
+        // The change to support checkout scm returning a map and that map being added to the environment works fine with
+        // older core etc, but just doesn't do anything, since checkout scm isn't returning anything yet. But with newer
+        // core, etc, it'll Just Work.
+        expect("scmEnvVars")
+                // TODO: switch to .logNotContains("GIT_COMMIT is null") once we've moved to core 2.60+,
+                // workflow-scm-step 2.6+, git 3.3.1+
```
jglick: Maybe a follow-on PR preparing those changes…

pipeline-model-extensions/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/CheckoutScript.groovy
```diff
@@ -44,17 +44,24 @@ public class CheckoutScript implements Serializable {
...
+                        checkoutMap = script.checkout script.scm
                     }
                 } else {
                     // No stage when we're in a nested stage already
-                    script.checkout script.scm
+                    checkoutMap = script.checkout script.scm
                 }
             }
-            body.call()
+            if (checkoutMap != null && checkoutMap instanceof Map) {
```
jglick: Either

```groovy
if (checkoutMap != null) { // assumed to be a Map
```

or

```groovy
if (checkoutMap instanceof Map) { // hence, not null
```

Probably you also want to avoid an extra block step if we are in legacy mode:

```groovy
if (checkoutMap instanceof Map && !checkoutMap.isEmpty()) {
```

which could also be simplified to just

```groovy
if (checkoutMap) { // handles null and empty maps
```

pipeline-model-extensions/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/CheckoutScript.groovy
```diff
@@ -44,17 +44,24 @@ public class CheckoutScript implements Serializable {
...
+                script.withEnv(checkoutMap.collect { k, v -> "${k}=${v}" }) {
```
jglick: Feeling confident in your `CpsDefaultGroovyMethods` there?

`"$k=$v"` suffices I think.

abayer: I am! Or more accurately, I am with newer versions of `workflow-cps`, which is already being bumped over in #174, but I'll incorporate that change here too.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
abayer: Nah, we've got to be really conservative on core versions here, since Blue Ocean depends on Declarative. Gotta coordinate core bumps with them, so I'll wait 'til there's something I legitimately can't do without a core bump.

jglick: I meant a follow-on PR to run tests, not to be merged any time soon. Or simply a branch.

rsandell: So this test will then fail when run in the PCT with the correct combinations?

PR #179, opened 2017-08-08
[FIXED JENKINS-46065] Add extension point for contributing validation

abayer:
* JENKINS issue(s):
    * [JENKINS-46065](https://issues.jenkins-ci.org/browse/JENKINS-46065)
* Description:
    * Adds an extension point for contributing additional validation checks.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


rsandell:
Possibly getting the run as a parameter would be good to have, to be able to do more validation in particular places, or only do validation on multibranch etc.

rsandell:
For example in the case of the `branch` when condition we know that only works for multi branch and when the Jenkinsfile is grabbed from an scm and so we would like to have that info when validating if `branch` can be used or not.

abayer:
@rsandell There are validation contexts where we don't have a `run` - i.e., CLI linting or the Editor - so I'd prefer not to have validation that only fires when we're in the actual run. I want consistency there if at all possible.

rsandell:
So it is better to fail on stage 45 because the `BRANCH_NAME` env var is missing than failing the build straight away but missing that check when linting over cli?

abayer:
@rsandell Lemme think on it...

abayer:
I've added `FlowExecution`s as `CheckForNull` parameters to the `DeclarativeValidatorContributor#validateElement` methods. I've also rewritten the logic for [JENKINS-45955](https://issues.jenkins-ci.org/browse/JENKINS-45955) to replace #176 by implementing blocked steps/methods as a `DeclarativeValidatorContributor`. There's a little bit of a harmless loophole in that if a different extension blocks a new method call (i.e., a wrapper step), it'll get past one validation check (`ModelValidatorImpl#validateElement(ModelASTOption)` checks to see if the name is in `Options#getAllowedOptionTypes()`'s keys - and those keys have steps blocked in `BlockedStepsAndMethodCalls#blockedInMethodCalls()` filtered out) but will still fail when `DeclarativeValidatorContributor#validateElement(ModelASTMethodCall, FlowExecution)` gets called for the new extension. Also, the offending step will still show up in the list of possible option names for validation errors. Imperfect, but I couldn't find a more elegant way to pull this off without keeping #176 as well as this, and that seemed silly.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
```diff
@@ -65,122 +66,122 @@
      * Fallback for any unknown element type. Always returns true, cannot be overridden.
      */
     @CheckForNull
-    public final String validateElement(@Nonnull ModelASTElement element) {
+    public final String validateElement(@Nonnull ModelASTElement element, @CheckForNull FlowExecution execution) {
```
rsandell: Documenting the `execution` would be usefull, esp. when it's `null` and not and why that might matter.
Should be enough to do on a class level so you don't need to cut & paste to all the methods :)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -705,6 +702,26 @@ class ModelValidatorImpl implements ModelValidator {
                 }
             }
         }
-        return valid
+        return validateFromContributors(agent, valid)
+    }
+
+    private boolean validateFromContributors(ModelASTElement element, boolean isValid, boolean isNested = false) {
+        boolean contributorsValid = DeclarativeValidatorContributor.all().every { contributor ->
+            String error = null
```
jglick: unused initializer

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -705,6 +702,26 @@ class ModelValidatorImpl implements ModelValidator {
...
+            if (error != null) {
+                errorCollector.error(element, error)
+                return false
+            } else {
+                return true
+            }
+        }
+        if (isValid) {
+            return contributorsValid
+        }
```
jglick: `else` what do you `return`?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/BlockedStepsAndMethodCalls.java
```diff
@@ -0,0 +1,87 @@
...
+public class BlockedStepsAndMethodCalls extends DeclarativeValidatorContributor {
+    /**
+     * Get the map of step names to rejection messages. Exposed statically for testing purposes.
+     */
+    public static Map<String, String> blockedInSteps() {
+        LinkedHashMap<String, String> map = new LinkedHashMap<String, String>();
+        map.put("stage", Messages.BlockedStepsAndMethodCalls_BlockedSteps_Stage());
+        map.put("properties", Messages.BlockedStepsAndMethodCalls_BlockedSteps_Properties());
+        map.put("parallel", Messages.BlockedStepsAndMethodCalls_BlockedSteps_Parallel());
+        return map;
```
jglick: `ImmutableMap.of`

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/BlockedStepsAndMethodCalls.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Get the map of method or step names to rejection messages. Exposed statically for testing purposes and for use
+     * in more granular validation of job properties vs declarative options vs wrapper steps.
+     */
+    public static Map<String, String> blockedInMethodCalls() {
+        Map<String, String> map = new LinkedHashMap<>();
+        map.put("node", Messages.BlockedStepsAndMethodCalls_BlockedSteps_Node());
+        map.putAll(blockedInSteps());
```
jglick: `Collections.singletonMap`

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * Override a {@code validateElement} method in your extension to have that check run at the end of normal 
+ * Each method should return a string containing the error message if validation fails, and null otherwise.
+ */
+public abstract class DeclarativeValidatorContributor implements ExtensionPoint {
+    /**
+     * Fallback for any unknown element type. Always returns true, cannot be overridden.
+     */
+    @CheckForNull
+    public final String validateElement(@Nonnull ModelASTElement element, @CheckForNull FlowExecution execution) {
```
jglick: Using method overloading this way is a very bad idea. Pick a distinct name for each method here.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * Extension point for contributing additional {@link ModelASTElement} validation checks to {@link ModelValidator} runs.
+ *
+ * Override a {@code validateElement} method in your extension to have that check run at the end of normal 
+ * Each method should return a string containing the error message if validation fails, and null otherwise.
+ */
+public abstract class DeclarativeValidatorContributor implements ExtensionPoint {
+    /**
+     * Fallback for any unknown element type. Always returns true, cannot be overridden.
```
jglick: So why is it in the API at all?

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
```diff
@@ -0,0 +1,197 @@
...
+import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * Extension point for contributing additional {@link ModelASTElement} validation checks to {@link ModelValidator} runs.
+ *
+ * Override a {@code validateElement} method in your extension to have that check run at the end of normal 
+ * Each method should return a string containing the error message if validation fails, and null otherwise.
```
jglick: Document the `execution` parameters also.

abayer:
@jglick - No downstream uses at this moment, just internal ones.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
abayer: ha, that's a mistake.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
abayer: Safety - we need to be able to handle cases like a new `ModelASTElement` being added to the validator without breaking compatibility with an extension of `DeclarativeValidatorContributor` that predates that new `ModelASTElement`.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
abayer: Going with method overloading to match up with how `ModelValidator` works.

abayer:
Bah, types.

PR #181, opened 2017-08-17
[FIXED JENKINS-46276] post stage not running inside stage agent block

rsandell:
When agent none is set globally and agent docker in the stage

* JENKINS issue(s):
    * [JENKINS-46276](https://issues.jenkins-ci.org/browse/JENKINS-46276)
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * @abayer 
    * @reviewbybees


abayer:
Pretty sure I got this fixed now - can't actually test it directly 'cos I'm on OS X and docker tests don't work there, but the problem would exist for `agent any` as well, which I did test.

PR #182, opened 2017-08-17
[FIXED JENKINS-46277] Do a pull for stage-level docker agent too

abayer:
* JENKINS issue(s):
    * [JENKINS-46277](https://issues.jenkins-ci.org/browse/JENKINS-46277)
* Description:
    * When we're at the top level, we want do do a pull of the docker image, but we want to do it within a synthetic stage. That's all well and good, but until now, the way this was done meant we weren't doing an image pull if we were at the stage level. That was dumb. Fixed!
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
@@ -50,6 +50,9 @@ public class DockerPipelineScript extends AbstractDockerPipelineScript<DockerPip
                 }
             }
             try {
+                if (Utils.withinAStage()) {
+                    script.getProperty("docker").image(describable.image).pull()
+                }
                 script.getProperty("docker").image(describable.image).inside(describable.args, {
                     body.call()
                 })


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
```diff
@@ -50,6 +50,9 @@ public class DockerPipelineScript extends AbstractDockerPipelineScript<DockerPip
                 }
             }
             try {
+                if (Utils.withinAStage()) {
+                    script.getProperty("docker").image(describable.image).pull()
+                }
```
rsandell: But doesn't the line below pull the image if it doesn't exist already?

abayer: If it doesn't exist, yes, but it doesn't update if there's a newer version.

oysols:
If you do not use a docker registry this change breaks the pipeline. Previously it was enough to keep the images locally, but this `pull` obviously fails if the images only exist locally.

The simple workaround is to patch the plugin to revert this change, but it would be nice to avoid that. If you think `pull` by default is the right thing, fine, but I think it makes sense to make it configurable.

Am I missing something? Is there some way to make it work with local images?

abayer:
@oysols Argh! Very sorry to hear that - could you open an issue at issues.jenkins-ci.org? I'm planning a 1.2.1 release for probably end of this week and will try to find a good option here.

oysols:
Opened an issue: https://issues.jenkins-ci.org/browse/JENKINS-47106
Let me know if it needs more specifics.

Danack:
Thank-you @oysols  for opening the issue.

I think it would be appropriate for the priority to be updated to be 'major' or 'critical', and for there to be some announcement about this being a breaking bug.

Upgrading to this version of the plugin, when you have images that are stored locally, apparently breaks the build pipeline. 

So anyone using the previous version and having local images in their pipeline need to be aware this bug will break their build.

(apologies in advance for cross-posting, but this caused quite a bit of headaches for us, which other people shouldn't have to suffer.)

abayer:
I'll be fixing the issue today or tomorrow, with a release planned for Friday, FYI.

PR #184, opened 2017-08-18
[WiP]Validate credentials id

rsandell:
Just putting it here for now, not even sure it works yet

* JENKINS issue(s):
    * Issue link or n/a
* Description:
    * Validate credentials id so we don't get a nasty surprise way down in the deploy stage.
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/CredentialsValidator.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/CredentialsValidator.java
@@ -1,0 +1,85 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+package org.jenkinsci.plugins.pipeline.modeldefinition.validator;
+
+import hudson.Extension;
+import hudson.model.Queue;
+import hudson.model.Run;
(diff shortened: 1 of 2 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/CredentialsValidator.java
```diff
@@ -0,0 +1,85 @@
...
+public class CredentialsValidator extends DeclarativeValidatorContributor {
+
+    @CheckForNull
+    @Override
+    public String validateElement(@Nonnull ModelASTEnvironment environment, @CheckForNull FlowExecution execution) {
+        final FlowExecutionOwner owner = execution != null ? execution.getOwner() : null;
+        if (owner != null) {
+            try {
+                final Queue.Executable executable = owner.getExecutable();
+                if (executable != null && executable instanceof Run) {
```
abayer: In practice, this will be `WorkflowRun` in any situation where it's not null, but eh, doesn't really matter. =)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/CredentialsValidator.java
```diff
@@ -0,0 +1,85 @@
...
+                    for (Map.Entry<ModelASTKey, ModelASTEnvironmentValue> entry : environment.getVariables().entrySet()) {
+                        if (entry.getValue() instanceof ModelASTInternalFunctionCall) {
+                            ModelASTInternalFunctionCall func = (ModelASTInternalFunctionCall) entry.getValue();
+                            if (func.getName().equals("credentials")) {
+                                if (func.getArgs().isEmpty()) {
+                                    return Messages.CredentialsValidator_noArgs();
+                                } else {
+                                    final ModelASTValue idValue = func.getArgs().get(0);
+                                    if (idValue.isLiteral() && !idValue.getValue().toString().startsWith("${")) {
+                                        //In this context I guess we can only evaluate literals, not runtime computed values
```
abayer: Correct. I still need to decide what to do about non-literal values in other places we do validation, like `tools` - right now, we don't allow non-literal values there because it'd break validation, but we've already crossed that bridge with `credentials`, so I'd say that yeah, we can't exactly do validation of non-literals and that's just the price of doing business.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/CredentialsValidator.java
```diff
@@ -0,0 +1,85 @@
...
+    @CheckForNull
+    @Override
+    public String validateElement(@Nonnull ModelASTEnvironment environment, @CheckForNull FlowExecution execution) {
+        final FlowExecutionOwner owner = execution != null ? execution.getOwner() : null;
+        if (owner != null) {
+            try {
+                final Queue.Executable executable = owner.getExecutable();
+                if (executable != null && executable instanceof Run) {
+                    //This might show the error not exactly in the right place
+                    for (Map.Entry<ModelASTKey, ModelASTEnvironmentValue> entry : environment.getVariables().entrySet()) {
```
abayer: This raises an interesting point - the validation extensions don't let you drill down to say "apply this validation to the value of environment variables", since those are going to be either `ModelASTValue`s or `ModelASTInternalFunctionCall`s. The `validateElement` methods for those don't give the context in which they exist, so there's no easy way to know "Oh yeah, that's being called from an environment variable, so..." Need to think on this more down the road, but this is as good as we can do right now.

bitwiseman:
@rsandell  
Tests for this? 
Could we get a JIRA to track this? 
Also, over 18-months old now, are we going to finish this up or should we close it?

rsandell:
@bitwiseman This was mostly a proof of concept to check the feasability of the validator extension point. I personally don't plan to work on it more, but I guess it's a good base for someone to continue working on, as it's something that I really think is needed.
But I think the API needs to be extended a bit to allow the validator to know the context of the thing it's validating as abayer mentioned.

bitwiseman:
I've created an issue for this and linked this PR and a new branch. 
https://github.com/jenkinsci/pipeline-model-definition-plugin/tree/issue/JENKINS-60046
https://issues.jenkins-ci.org/browse/JENKINS-60046
Closing for now.

PR #186, opened 2017-08-22
[FIXED JENKINS-46359] Reset failure status in retry option

abayer:
* JENKINS issue(s):
    * [JENKINS-46359](https://issues.jenkins-ci.org/browse/JENKINS-46354)
* Description:
    * Also change the build condition responder extension point to take an exception, which we can use to decide what the build status actually is. Backwards compatibility is retained (see Unstable.groovy and Always.groovy - they don't need to do anything with the exception, so I left them as is to demonstrate the backwards compatibility).
    * I don't *love* this. It appears to work and is probably smarter long term - gets us away from depending on the build status quite as much. But if someone's got `currentBuild.result` in their Pipeline somewhere, it's always going to be null/success. This definitely needs more thought. I'm aiming post-1.2 for this.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
```diff
@@ -45,13 +46,45 @@
...
+    public boolean meetsCondition(WorkflowRun r) {
+        if (Util.isOverridden(BuildCondition.class, getClass(), "meetsCondition", WorkflowRun.class)) {
+            return getClass().cast(this).meetsCondition(r);
+        } else {
+            return true;
+        }
+    }
+
+    public boolean meetsCondition(WorkflowRun r, Exception error) {
+        if (Util.isOverridden(BuildCondition.class, getClass(), "meetsCondition", WorkflowRun.class, Exception.class)) {
```
rsandell: This ... looks ... a bit strange.
If the method is overridden this code wouldn't run anyway?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
```diff
@@ -37,8 +38,9 @@ import org.jenkinsci.plugins.workflow.job.WorkflowRun
 @Extension(ordinal=800d) @Symbol("aborted")
 public class Aborted extends BuildCondition {
     @Override
-    public boolean meetsCondition(WorkflowRun r) {
-        return r.getResult() != null && r.getResult().equals(Result.ABORTED)
+    public boolean meetsCondition(WorkflowRun r, Exception error) {
+        return (error != null && Utils.getResultFromException(error) == Result.ABORTED) ||
```
svanoort: A bit hacky but it is our best option at the moment.

abayer:
@rsandell Huh, yeah, that does work. Not sure I like it, but huh.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
svanoort: Yeaaaaaah

abayer: Oh yeah. See, I keep getting this logic wrong. =)

abayer:
...I don't like this. Closing it, will reconsider and reapproach in the future.

PR #189, opened 2017-08-31
[JENKINS-43911] Process creds and env before agent

abayer:
* JENKINS issue(s):
    * [JENKINS-43911](https://issues.jenkins-ci.org/browse/JENKINS-43911)
* Description:
    * This allows use of environment variables in agent fields.
    * I'm not yet sure about this - this creates some potential problems with, say, file-based `credentials`, which would no longer work, or node environment variable properties on an agent, which you could no longer refer to in `environment`. So...need to make some decisions. It's also possible that we could do some kind of repeat injection of `environment` and `credentials`, but I don't love that idea either. And it'd still flop for file-based credentials, actually. So...yeah. I'd like some discussion on this.
* Documentation changes:
    * Depends on what we end up doing - we may want to clarify the ordering.
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -63,9 +63,9 @@ public class ModelInterpreter implements Serializable {
                 executeProperties(root)
 
                 // Entire build, including notifications, runs in the agent.
-                inDeclarativeAgent(root, root, root.agent) {
-                    withCredentialsBlock(root.environment) {
-                        withEnvBlock(root.getEnvVars(script)) {
+                withCredentialsBlock(root.environment) {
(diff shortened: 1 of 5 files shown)


abayer:
Yeah. I don't like it. Definitely not getting into 1.2 - a lot more thought is needed.

rsandell:
Well you changed this behavior once before, it used to be almost like you describe this change to be, but changed it, I believe due to a user request, to be done inside node.

abayer:
Closing this sucker. Will go back to trying to find a workaround instead.

Buzzard44:
"It's also possible that we could do some kind of repeat injection of environment and credentials"

^I think this is the way to go. If we evaluate before we enter the agent and then a second time after the agent, we can use variables to pick agents (as there exist scenarios currently where the only workaround is to not use declarative pipeline for all or some large piece of your pipeline), and then once inside the node we resolve with all the nice context of being inside the node. It might mean nesting functions a couple layers deeper, but I believe the functionality would be what most people expect (and want).

PR #192, opened 2017-09-12
[JENKINS-46831] Test behaviour when agent defined globally and on stage

alecharp:
* JENKINS issue(s):
    * n/a
* Description:
    * I faced a problem when the agent on the global level is a Docker container and when the agent on the stage is a normal agent. This ends up with a very odd issue. I'd like to make sure it's coming from my end and not from the plugin.
* Users/aliases to notify:
    * @abayer

### DIFF
--- a/pipeline-model-definition/src/test/resources/agentDockerGlobalThenLabel.groovy
+++ b/pipeline-model-definition/src/test/resources/agentDockerGlobalThenLabel.groovy
@@ -1,0 +1,48 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


pipeline-model-definition/src/test/resources/agentDockerGlobalThenLabel.groovy
```diff
@@ -0,0 +1,48 @@
...
+            image 'maven:3.5.0-jdk-8-alpine'
+            label 'docker'
+        }
+    }
+    stages {
+        stage("Maven") {
+            steps { sh 'echo "first agent = $WHICH_AGENT"' }
+        }
+        stage("Agent") {
+            agent { label 'some-label' }
```
abayer: This should be `other-docker` or else everything hangs. =)

abayer:
So in a local test, even after I changed `some-label` to `other-docker` to un-hang things, I got a failure due to workspace whackiness, with a mention of [JENKINS-33510](https://issues.jenkins-ci.org/browse/JENKINS-33510). I think the problem is due to the fake agents created by `JenkinsRule.createOnlineSlave()` using local paths for everything, but not sure.

abayer:
Nope, it's a general problem - reproduced it by hand. It's specifically that you can't invoke `sh`/`bat`/`ps` inside a `node` block within a `docker.image(...).inside` block due to [JENKINS-33510](https://issues.jenkins-ci.org/browse/JENKINS-33510). Huh!

JonnyPower:
I'm seeing similar behaviour and was writing a similar test - I believe the container id from the outer agent is expected on the inner agent which makes no sense (the container ids are not going to be the same on different nodes)

[JENKINS-33510](https://issues.jenkins-ci.org/browse/JENKINS-33510) is supposedly resolved, but the test here still fails - and my own experience still sees the issue this test replicates

PR #193, opened 2017-09-12
[JENKINS-46547] Allow loading pipeline {} blocks from shared libraries

abayer:
* JENKINS issue(s):
    * [JENKINS-46547](https://issues.jenkins-ci.org/browse/JENKINS-46547)
* Description:
    * This adds support for loading `pipeline { ... }` blocks from shared libraries. A `call()` method in a shared library "step" will now be able to include one or more `pipeline { ... }` blocks, though only one should be executed.
    * Only can be used in `vars/*.groovy` - not in `src/...` or via `load(...)`. 
    * Validation works - in fact, you can do `if (something) { pipeline { ... } } else { pipeline { ... } }` and both blocks will be validated. 
    * Only the actually executing `pipeline { ... }` block will be available via the `ExecutionModelAction.getStages()` method, which is used by Blue Ocean to render the execution plan.

- [x] Tests to verify we won't stomp on a shared library "step" named `pipeline`. (EDIT: Reconsidering this - that won't work anyway currently, since we're already inserting `pipeline` as a "step"/global variable. The only scenario I can think of where this would be relevant would be a `Jenkinsfile` or shared library that has a closure it trampolines a la the old parser for Declarative, evaluating everything before ever getting to CPS evaluation. And frankly, I'm not sure it's worth worrying about that weird use case, especially since we're now only transforming `pipeline` blocks if they contain `agent` and `stages` anyway.)
- [x] Decide whether to fold [JENKINS-42224](https://issues.jenkins-ci.org/browse/JENKINS-42224) into this change or tackle that separately later. (EDIT: Deferring - we'll do that later)
- [x] Try to find more ways to guarantee we're only parsing/validating/transforming `Jenkinsfile` or `vars/*.groovy` and nothing else. (EDIT: Now only parsing etc for `WorkflowScript` and classes without packages whose name matches a `GlobalVariable`, since `vars/*.groovy` get added as `GlobalVariable`s before they get compiled - nifty!)
- [x] Support multiple `pipeline` blocks in a shared library "step" - i.e., if statements etc.
- [x] Determine how `ExecutionModelAction` will know which `stages` instance is the one that's actually running if there are more than one. (EDIT: done - using `ModelASTStages.hashCode()` stored in the actual `Root` instance we execute to figure out what's run at runtime, and then having `ExecutionModelAction.getStages()` use that hashcode to figure out which specific instance to return)
- [x] Ensure we only execute one `pipeline { ... }` block - probably by failing the build if we attempt to execute a second?

* Documentation changes:
    * Docs will be needed.
* Users/aliases to notify:
    * @reviewbybees


abayer:
@jglick - would appreciate your thoughts on this as well, if you have a chance.

abayer:
Moving this to 1.2 - while `pipeline { ... }` has never been officially supported in shared libraries, it seems kind of ridiculous to do a release that breaks that usage and then later another release that fixes it and makes it supported, when the fix is already up for review.

rsandell:
But why should this be fixed?
I just fear we start supporting something that we shouldn't. Just one more thing that can break with imho close to no value?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -263,6 +264,18 @@ public class Utils {
...
 
+    static void markExecutedStagesOnAction(CpsScript script, int astHashCode) throws Exception {
+        WorkflowRun r = script.$build()
+        ExecutionModelAction action = r.getAction(ExecutionModelAction.class)
+        if (action != null) {
+            if (action.stagesHashCode > 0) {
+                throw new IllegalStateException("Only one pipeline { ... } block can be executed in a single run.")
+            }
+            action.setStagesHashCode(astHashCode)
+            r.save()
```
rsandell: Do you have to save here? Pipelines are hitting the master's storage too much as it is.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
```diff
@@ -57,9 +57,11 @@ public class Root implements Serializable {
 
     Libraries libraries
 
+    int astHashCode
```
rsandell: `final`?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Yeah, we do - the action already exists, so just modifying it won't result in it being saved. Also, this literally happens only once per build (since if this method gets called a second time, we throw that exception above!)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
abayer: Will do.

abayer:
@rsandell So that was my thought for a long time too, but a couple things changed my mind.

* I know people are using `pipeline { ... }` in shared libraries currently due to bug reports about `environment` and `when` `expression`s not working from there (due to the implementation of the old parsing logic).
* In 1.2, *all* usage of `pipeline { ... }` in shared libraries will fail, due to the new parser, so this goes from just being an unsupported/undocumented behavior to what will appear to be a regression to people already using it.
* On the positive side, the new parser made implementing the guts of supporting `pipeline { ... }` in shared libraries extremely easy - no need to worry about the `Closure` trampoline/translation stuff applying there, etc, just do the same transformation magic we do in `Jenkinsfile`. The harder parts were making sure we don't try to parse/transform code we shouldn't, limiting where the parsing/transformation can be done to where it'd actually work at runtime (i.e., to `vars`, not `src`), and not breaking the contract with Blue Ocean for the stages execution plan that'll actually be executed being available via `ExecutionModelAction.getStages()`. That's all done, so...

I don't think this adds much maintenance or exposure for bugs that wasn't there before - the parsing is exactly the same as in `Jenkinsfile`s, once you get past a tiny bit of additional logic for handling the potential of multiple `pipeline { ... }` blocks to be parsed (but not executed - i.e., `if (whatever) { pipeline { ... } } else { pipeline { ... } }`), and the logic for determining whether to transform a Groovy file is something I'm really confident in.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Also, this is writing the run's `config.xml`, not the XML storage for an individual `FlowNode`, so it's no worse than, say, setting the build result, etc.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -139,16 +142,38 @@ class ModelParser implements Parser {
...
+                List<Statement> pipelineSteps = finder.pipelineSteps
+
+                if (!pipelineSteps.isEmpty()) {
+                    List<ModelASTPipelineDef> pipelineDefs = pipelineSteps.collect { p ->
+                        return parsePipelineStep(p, secondaryRun)
+                    }
+                    // Even if there are multiple pipeline blocks, just return the first one - this return value is only
+                    // used in a few places: tests, where there will only ever be one, and linting/converting, which also
+                    // are guaranteed to only have one, since we don't intend to support linting of pipelines defined in
+                    // shared libraries.
```
rsandell: So why waste time parsing them all?

abayer: Because I can't tell which one is going to be executed. Basically, the return value of `parse` is irrelevant except for those very few places, none of which are applicable to shared libraries in the first place, so we can keep just returning the first one we parse. But we do need to parse all of them in case of the `if (whatever) { pipeline { ... } } else { pipeline { ... } }` case.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
rsandell: Well my other concern was around badly coded `SaveableListener`s out there that might get confused with more than one save per build.
I would guess the build is saved at the end as for freestyles but there might be more to it here if it should survive restarts.

abayer: Oh, no, `run.save()` gets called a *bunch* already - like I said, any time the build result gets set, a new action is added, you name it.

abayer:
Hmm, couldn't reproduce the test failure locally. Will investigate why it's being flaky.

abayer:
Nope, even weirder - when I run just the one test via Maven, it passes. But when I run all of `BasicModelDefTest`, the test fails. There's some lurking state somewhere...

abayer:
Oh geez I'm dumb. Hashcodes can be negative. *facepalm*

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -606,7 +610,8 @@ class RuntimeASTTransformer {
                     transformOptions(original.options),
                     transformTriggers(original.triggers),
                     transformParameters(original.parameters),
-                    transformLibraries(original.libraries)))
+                    transformLibraries(original.libraries),
+                    constX(original.stages.hashCode())))
```
jglick: Humor me and use CRC32 if not SHA-*. `String.hashCode` is pretty weak. Probably this is not a security consideration but are you willing to stake your reputation on that?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
```diff
@@ -33,15 +46,34 @@ public void configureCompiler(@CheckForNull final CpsFlowExecution execution, Co
...
+                    doModelParsing = true;
+                } else if (execution != null && classNode.getPackageName() == null) {
+                    // Second, if we've got an execution and there's no package name, we'll parse if this is a global
+                    // variable, which we can determine by looking to see if its name is in the list of global variables.
+                    // Note that the combination of no package name plus global variable name should keep us from
+                    // parsing src/org/whatever/Foo.groovy in shared libraries *or* global variables defined in plugins.
+                    try {
+                        FlowExecutionOwner owner = execution.getOwner();
+                        if (owner != null && owner.getExecutable() instanceof WorkflowRun) {
+                            WorkflowRun run = (WorkflowRun)owner.getExecutable();
```
jglick: `Run` suffices here

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
```diff
@@ -33,15 +46,34 @@ public void configureCompiler(@CheckForNull final CpsFlowExecution execution, Co
...
+                            for (GlobalVariable v : GlobalVariable.forRun(run)) {
```
jglick: Hmm, should probably make an overload taking `FlowExecution[Owner]`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
```diff
@@ -33,15 +46,34 @@ public void configureCompiler(@CheckForNull final CpsFlowExecution execution, Co
...
-
-                // but until that happens,
-                // Using getNameWithoutPackage to make sure we don't end up executing without parsing when an inadvertent package name is put in the Jenkinsfile.
-                if (classNode.getNameWithoutPackage().equals(Converter.PIPELINE_SCRIPT_NAME)) {
+                boolean doModelParsing = false;
+
+                // First, we'll parse if this is WorkflowScript, i.e., a Jenkinsfile, or it's an autogenerated script
+                // with no package, i.e., Script1
+                if (classNode.getNameWithoutPackage().equals(Converter.PIPELINE_SCRIPT_NAME) ||
+                        (classNode.getPackageName() == null && classNode.getNameWithoutPackage().matches("Script\\d+"))) {
```
jglick: Going to blow up when someone fixes [JENKINS-31838](https://issues.jenkins-ci.org/browse/JENKINS-31838). Find a better way. `CodeSource`?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -139,16 +142,38 @@ class ModelParser implements Parser {
...
         }
 
-        if (pst==null) {
-            // Check if there's a 'pipeline' step somewhere nested within the other statements and error out if that's the case.
-            src.statementBlock.statements.each { checkForNestedPipelineStep(it) }
-            return null; // no 'pipeline', so this doesn't apply
+        if (pst != null) {
+            return parsePipelineStep(pst, secondaryRun)
+        } else {
+            // Look for the pipeline step inside methods named call.
```
jglick: Oof.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
abayer: So the only reason I'm actually using `hashCode` is that it's trivial to get off an object. If there's a better method I should be calling, point me to it and I will *happily* use it.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
abayer: This would bust the line above too, right? Which we've had in place for ages...hrm. Lemme think on it for a bit.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
abayer: Gotcha. At some point I was doing something that actually cared about `WorkflowRun`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
abayer: Would be handy!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: Actually a conscious decision on my part - we're only going to support `pipeline {...}` blocks in the `call()` method of `vars/*.groovy`. Limits the scope of transformation and the ways things could go weird.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
abayer: And also, yes, I actually am willing to stake my reputation that this is not a security consideration - the hash code is just used to try to determine which of the possibly multiple `ModelASTStages` instances in `ExecutionModelAction` is the one we actually invoked, so that we can make `ExecutionModelAction.getStages()` return that instance. And that method is only used by Blue Ocean to pre-render the stages that *will* run before they've actually run. So worst case is that the wrong group of stages renders initially.

That said, I'd be fine with literally anything as a unique identifier for a `ModelASTStages` instance - if there's a good way to randomly generate a string or whatever at instantiation time that I can add as a field to `ModelASTStages`, that'd be perfect. I just went with `hashCode()` because I knew it'd work...and I knew you'd eventually tell me the right way I should do it. =)

abayer: Switched around to store a `UUID` on `ModelASTStages` that we generate at initialization, and a `String` representation of it on `Root` (to make the sandboxed transformation stuff easier than trying to instantiate a `UUID` from the existing `ModelASTStages.getUuid()` `String`).

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -263,6 +264,18 @@ public class Utils {
         return nodes
     }
 
+    static void markExecutedStagesOnAction(CpsScript script, String astUUID) throws Exception {
```
svanoort: Seems kind of hacky.

abayer: It is a bit hacky, yes, but given that I don't actually know which `pipeline` block (if there are multiple being read from the `Jenkinsfile` and shared libraries) is going to be executed until it actually *is* executed...and hey, it works. If a better idea comes along in the future, I'll take advantage of it. =)

abayer:
Merged, cutting a beta.

casz:
@abayer THANK YOU <3

PR #195, opened 2017-09-14
Fix PCT issues caused by newer durable task step

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * Newer durable task step plugin doesn't seem to know what to do with an escaped $ in an env var, so...bugger the lazy evaluation tricks I was doing, just do the simple thing.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @kshultzCB

### DIFF
--- a/pipeline-model-definition/src/test/resources/environmentWithWorkspace.groovy
+++ b/pipeline-model-definition/src/test/resources/environmentWithWorkspace.groovy
@@ -25,7 +25,7 @@
 pipeline {
     environment {
         FOO = "FOO"
-        BAR = '${WORKSPACE}BAR'
+        BAR = "${WORKSPACE}BAR"
     }
     agent {
         label "some-label"
@@ -34,7 +34,7 @@ pipeline {
     stages {
         stage("foo") {
             environment {
-                BAZ = "\${FOO}BAZ"
+                BAZ = "${FOO}BAZ"
             }
 
             steps {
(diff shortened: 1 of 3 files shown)


abayer:
I'll try to dig down into what exactly the problem is later, but I'm pretty confident it's not Declarative specific...

abayer:
Pretty sure it's in https://github.com/jenkinsci/durable-task-plugin/compare/durable-task-1.12...durable-task-1.13 - specifically, I think https://github.com/jenkinsci/durable-task-plugin/compare/durable-task-1.12...durable-task-1.13#diff-8c4fb36832416fbbbf56e1ac7747491eR90.

That is, the escaping of `$` in the env vars is resulting in an env var with the literal string `${SOMETHING}` not being processed in `sh` execution into the value of `SOMETHING` in the environment. It's probably a reasonable change in general that just happens to bite two test cases in Declarative where I was trying to verify the env var referenced was actually in `sh`'s environment by escaping the `$` in the initial declaration. No one else has reported any problems with this in the 7 months and 1 day since `workflow-durable-task-step` 2.9 was released, so I think we can safely conclude I was just being weird and there's no reason to expect this should behave the way it did previously. Hence this change just being switching the tests around.

jglick:
> an env var with the literal string `${SOMETHING}` not being processed in `sh` execution into the value of `SOMETHING` in the environment

Yes, this was a bug fix.

PR #197, opened 2017-09-22
[FIXED JENKINS-47064] Get rid of .every call for when evaluation

abayer:
* JENKINS issue(s):
    * [JENKINS-47064](https://issues.jenkins-ci.org/browse/JENKINS-47064)
* Description:
    * This can lead to a serialization error in some cases. I haven't nailed down exactly what's distinct about those cases, though, so while this solves the symptom, it doesn't address whatever the underlying problem is that makes `.every` freak out in some situations. More investigation needed there, probably at the `groovy-cps` level.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -193,6 +193,12 @@
 
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>pipeline-utility-steps</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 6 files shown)


abayer:
Closing this since it's better handled over in https://issues.jenkins-ci.org/browse/JENKINS-47071 in `workflow-cps` (specifically in https://github.com/jenkinsci/workflow-cps-plugin/pull/177 and https://github.com/cloudbees/groovy-cps/pull/75)

PR #198, opened 2017-09-25
Make the plugin name match user expectations

rtyler:
When somebody hears about "Declarative Pipeline" they will logically go look for it in the plugins site or Update Center :smile:


rsandell:
Should have changed the title of the configuration UI as well

PR #199, opened 2017-09-28
[FIXED JENKINS-47106] Only do explicit docker pull if specified

abayer:
* JENKINS issue(s):
    * [JENKINS-47106](https://issues.jenkins-ci.org/browse/JENKINS-47106)
* Description:
    * Automatically doing docker pull was breaking usage of local images. It may make sense to re-enable the automatic pull for the top-level agent, though.
    * Also added a way to store arbitrary DeclarativeOptions on the DeclarativeAgent, so that we don't have to hardcode any new options in the future.
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1152
* Users/aliases to notify:
    * @reviewbybees


abayer:
Switched to an `agent` `docker` option rather than a global `DeclarativeOption`, though I also left the logic for arbitrary `DeclarativeOption`s being pulled into `DeclarativeAgent` in place, because I can definitely see places that'd be useful.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
```diff
@@ -66,6 +73,25 @@ public boolean isDoCheckout() {
         return doCheckout;
     }
 
+    public void addOption(@Nonnull DeclarativeOption option) {
```
rsandell: Is there any effect to this? I can't see any.

abayer: Nope, I just left it dangling. On second thought, I'll clean it out.

abayer:
@rsandell Test updated to first build the dummy image tagged as `maven:3-alpine`, verify that without `alwaysDoPull` that's the image that gets used, and then verify that with `alwaysDoPull`, the real `maven:3-alpine` is used instead.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
```diff
@@ -89,6 +89,9 @@ public class Agent extends MappedClosure<Object,Agent> implements Serializable {
                 }
             }
             a.setDoCheckout(doCheckout)
+            root?.options?.options?.each { k, v ->
```
rsandell: Forgot to remove this?

abayer: argh yup

pipeline-model-definition/src/test/resources/dockerPullLocalImage.groovy
```diff
@@ -0,0 +1,61 @@
...
+            steps {
+                sh 'cat /hi-there'
+                sh 'echo "The answer is 42"'
+            }
+        }
+        stage("in pulled image") {
+            agent {
+                docker {
+                    image "maven:3-alpine"
+                    alwaysDoPull true
```
rsandell: missed renaming this

abayer: oh goddammit me.

PR #200, opened 2017-09-28
[FIXED JENKINS-47193] Ignore other classes in Jenkinsfiles

abayer:
* JENKINS issue(s):
    * [JENKINS-47193](https://issues.jenkins-ci.org/browse/JENKINS-47193)
* Description:
    * For transformation purposes, we only are looking at the top level of the Jenkinsfile for pipeline blocks. So ignore classes defined in the Jenkinsfile as well, or we get errors due to things already having been transformed.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
@@ -47,11 +47,12 @@ public class GroovyShellDecoratorImpl extends GroovyShellDecorator {
             @Override
             public void call(SourceUnit source, GeneratorContext context, ClassNode classNode) throws CompilationFailedException {
                 boolean doModelParsing = false;
-
                 // First, we'll parse anything that's not coming from an actual source file on disk - i.e., Jenkinsfiles
                 // (which are read in as strings and then compiled), dynamically loaded/evaluated strings, etc. We can
-                // tell when we've got one of those by looking to see if the code source's location is "file:/groovy/shell".
-                if (context.getCompileUnit().getCodeSource().getLocation().toString().equals("file:/groovy/shell")) {
+                // tell when we've got one of those by looking to see if the code source's location is "file:/groovy/shell"
+                // and if the classNode is a script.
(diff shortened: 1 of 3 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/GroovyShellDecoratorImpl.java
```diff
@@ -47,11 +47,12 @@ public void configureCompiler(@CheckForNull final CpsFlowExecution execution, Co
...
                 boolean doModelParsing = false;
-
                 // First, we'll parse anything that's not coming from an actual source file on disk - i.e., Jenkinsfiles
                 // (which are read in as strings and then compiled), dynamically loaded/evaluated strings, etc. We can
-                // tell when we've got one of those by looking to see if the code source's location is "file:/groovy/shell".
-                if (context.getCompileUnit().getCodeSource().getLocation().toString().equals("file:/groovy/shell")) {
+                // tell when we've got one of those by looking to see if the code source's location is "file:/groovy/shell"
+                // and if the classNode is a script.
+                if (context.getCompileUnit().getCodeSource().getLocation().toString().equals("file:/groovy/shell") &&
+                        classNode.isScript()) {
```
abayer: Note that this basically means that we'll transform a `Jenkinsfile`, something that's been `evaluate`d, etc, but not any other classes in those files. Before this, we'd end up getting called twice if there was a `Class` in the `Jenkinsfile` as well as the `pipeline` block, and things would blow up good at that point - I'm still not entirely clear how `isDeclarativePipelineStep(it)` is true when `parsePipelineStep(pst, secondaryRun)` (where the first `it` ends up as `pst`) passes the transformed statement, not the original one, but hey, that's what was happening and this makes it not happen.

quasarea:
Not sure if its relevant, but enums have the same issue.

```
enum Modes { Debug, Release }
pipeline {
  ...
}
```

abayer:
@quasarea - Yup, same cause, fix applies for both. But for giggles, I added an `enum` to the test as well.

quasarea:
thanks, could you suggest any workaround, i.e. would shared library work? without this my pipelines are pretty useless ;/

abayer:
@quasarea Release is coming in...well, nowish, should be in the update center in a couple hours at most.

quasarea:
thanks, works great!

PR #201, opened 2017-09-28
[FIXED JENKINS-47109] stage-level failFast w/ parallel stages

abayer:
* JENKINS issue(s):
    * [JENKINS-47109](https://issues.jenkins-ci.org/browse/JENKINS-47109)
* Description:
    * Allows setting `failFast` on a stage. Piggybacks on existing parallel branches JSON syntax, so roundtripping is built in. Woo.
    * Might want to also reject if you use failFast in a non-parallel-parent stage, but it's pretty much harmless, so...eh.
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1156
* Users/aliases to notify:
    * @reviewbybees 
    * @kshultzCB

### DIFF
--- a/pipeline-model-definition/src/test/resources/parallelStagesFailFast.groovy
+++ b/pipeline-model-definition/src/test/resources/parallelStagesFailFast.groovy
@@ -1,0 +1,54 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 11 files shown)


pipeline-model-definition/src/test/resources/parallelStagesFailFast.groovy
```diff
@@ -0,0 +1,54 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    agent none
+    stages {
+        stage("foo") {
+            failFast true
```
kzantow: Just to be clear: `failFast false` is the default, is there a reason to ever set that? could this just be `failFast` ?

kzantow: (I was kinda hoping, and maybe it's stupid, to have a syntax like:
```
parallel failFast {
...
```
)

abayer: Yeah, that's stupid and not happening for any number of reasons. =)

First and foremost, we can't do `failfast` as a "field" on its own because then it's not a method call and things blow up in parsing. And to do something like `parallel failFast { ... }`, we'd need to have a "variable" `failFast` defined to make that work syntactically, a la `agent none` and `agent any`, and I ain't goin' down that road again. =)

PR #202, opened 2017-09-29
Build and test on Windows yay

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * Thought I'd finally get the tests all able to pass on Windows, and add Windows to the Jenkinsfile run. Down the road (once ci.jenkins.io is on Declarative 1.2 or later and once the junit-per-step stuff is done), I'll switch to running Windows and Linux builds in parallel.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @jtnord because it's Windows-related so I feel an obligation to namecheck him

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,7 +59,7 @@ pipeline {
                 label "windows"
             }
             steps {
-                bat 'mvn clean install -Dmaven.test.failure.ignore=true -Dcodenarc.skip=true'
+                bat 'mvn clean install -Dconcurrency=1 -Dmaven.test.failure.ignore=true -Dcodenarc.skip=true'
             }
             post {
                 always {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,44 +26,42 @@ pipeline {
     stages {
         // While there is only one stage here, you can specify as many stages as you like!
         stage("build") {
-            agent {
-                label "java"
-            }
-            steps {
-                sh 'mvn clean install -Dmaven.test.failure.ignore=true'
-            }
-            post {
-                // No matter what the build status is, run this step. There are other conditions
-                // available as well, such as "success", "failed", "unstable", and "changed".
-                always {
-                    junit '*/target/surefire-reports/*.xml'
+            parallel {
+                stage("linux") {
+                    agent {
+                        label "java"
+                    }
+                    steps {
+                        sh 'mvn clean install -Dmaven.test.failure.ignore=true'
+                    }
+                    post {
+                        // No matter what the build status is, run this step. There are other conditions
+                        // available as well, such as "success", "failed", "unstable", and "changed".
+                        always {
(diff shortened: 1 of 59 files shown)


pipeline-model-definition/src/test/resources/envDollarQuotes.groovy
```diff
@@ -39,10 +39,10 @@ pipeline {
             }
 
             steps {
-                sh 'echo "FOO is $FOO"'
-                sh 'echo "BAR is $BAR"'
-                sh 'echo "BAZ is $BAZ"'
-                sh 'echo "SPLODE is $SPLODE"'
+                echo "FOO is $FOO"
```
rsandell: This is not really the same. In the former version you tested variable expansion in `sh` this one tests it in groovy-cps.

rsandell: But that probably doesn't matter.

rsandell:
:bee:

pipeline-model-definition/src/test/resources/envDollarQuotes.groovy
abayer: Very good point. @jtnord or anyone else - how would I write to a file from `bat`?

rsandell: `echo FOO is %FOO% > somefile.txt`

abayer: Okiedokie.

abayer: Oh wait, that's for a different test. Ehhhh, I'll update 'em both.

Jenkinsfile
```diff
@@ -25,31 +22,50 @@ pipeline {
...
             }
+            post {
+                // No matter what the build status is, run this step. There are other conditions
+                // available as well, such as "success", "failed", "unstable", and "changed".
+                always {
+                    junit '*/target/surefire-reports/*.xml'
+                }
+                success {
+                    archive "**/target/*.hpi"
+                    archive "**/target/site/jacoco/jacoco.xml"
```
jtnord: why not use the jaccoco plugin?  Also generally better to run with and without coverage in a different lifecycle...
But this is not anything new so - Meh...

jtnord:
Code looks ok, left some drive by comments  - but [failing tests](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fpipeline-model-definition-plugin/detail/PR-202/12/tests) :-)

Jenkinsfile
abayer: Simple answer - at the time we added that, the jacoco plugin didn't have a `@Symbol` yet. =) But it does now, so I'll add that.

abayer: Check that, it still hasn't been released yet.

PR #203, opened 2017-10-03
Enable Docker tests on Mac with directions for configuration

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * Finally figured out the particular incantations needed to enable tests using `docker-workflow` in `JenkinsRule` to actually work on OS X. Woo.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


DEV_GUIDE.md
```diff
@@ -1,5 +1,15 @@
...
+## Docker tests on OS X
+
+Docker for Mac out of the box will not be able to run the Docker tests due
+to mount issues. To fix this, add the following in Docker Preferences ->
+File Sharing, in this order:
+* `/tmp` - this will initially add as `/private/tmp`. Double-click on `/private/tmp` and change to just `/tmp`.
+* `/var/folders` - this will initially add as `/private/var/folders`. 
+Double-click on `/private/var/folders` and change to just `/var/folders`.
+* `/private` - make sure to add this last.
+
```
bitwiseman: How can I verify this works?  Just build the plugin on Mac? 
What is the minimal supported version of docker for these tests?

bitwiseman: I've made no changes to my Docker install on my Mac and get the following: 

screen shot 2017-10-04 at 2 07 27 pm

So, it looks like I'd be maybe in trouble?  I'd have to add `/var/folders` and then remove and add `/private` again?

It is likely that these settings have changed over time.  Perhaps it would better to describe what values must be present and which not?  In any case, there needs to be more here to make it clear what the key points are.

abayer: Ah, yeah, gotta remove `/private` first. It's annoying. I also have no idea what the minimum Docker version is - I think anything works, but I'll note that it works with at least `2017.03` (which is what I have installed).

abayer:
Thanks, @bitwiseman!

PR #204, opened 2017-10-04
[FIXED JENKINS-47202] Switch to HashMaps for env to fix serialization

abayer:
* JENKINS issue(s):
    * [JENKINS-47202](https://issues.jenkins-ci.org/browse/JENKINS-47202)
* Description:
    * TreeMap#Entry isn't serializable, which can bite you in the weirdest places. Like here - if `environment` variable values end up triggering a program save, tada, you get a `NotSerializableException`. Which is bad. I can't reproduce this consistently - maybe 3 times out of 4, it'll fail. But this fix has yet to reproduce the error after 20+ attempts, so I think it's good.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


abayer:
So there *is* via `sleep`, but there's no logical way to do a `sleep` in `environment`...

PR #205, opened 2017-10-04
[FIXED JENKINS-47197] Make sure we don't end up with Java 8 bits

abayer:
* JENKINS issue(s):
    * [JENKINS-47197](https://issues.jenkins-ci.org/browse/JENKINS-47197)
* Description:
    * Groovy does something weird with `@Delegate` on a `Map` with Java 8 but targeting Java 7, resulting in a whole lot of additional methods created using `java.util.function.Function` and friends. That's annoying. To say the least.  Took me a while to find where that reference was being inserted, but yeah, it was `MappedClosure` and its `@Delegate`, so I got rid of `@Delegate` and now things are fine. Phew.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/MappedClosure.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/MappedClosure.groovy
@@ -41,13 +41,21 @@ import groovy.transform.ToString
 @SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
 public abstract class MappedClosure<O,M extends MappedClosure<O,M>> implements Serializable {
 
-    @Delegate Map<String,O> resultMap = [:]
+    Map<String,O> resultMap = [:]
 
     public MappedClosure() {
     }
 
     public MappedClosure(Map<String,O> inMap) {
         this.resultMap.putAll(inMap)
+    }
+
+    public O remove(String p) {
+        return resultMap.remove(p)
+    }
+
+    public add(String k, O v) {
+        resultMap.add(k, v)
     }
 
     public Map<String, Object> getMap() {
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/MappedClosure.groovy
```diff
@@ -50,6 +50,14 @@ public abstract class MappedClosure<O,M extends MappedClosure<O,M>> implements S
         this.resultMap.putAll(inMap)
     }
 
+    public O remove(String p) {
+        return resultMap.remove(p)
+    }
+
+    public add(String k, O v) {
```
rsandell: Isn't [put](https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#put(K,%20V)) the more conventional method to add stuff to a map?

abayer: i r stupid.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/MappedClosure.groovy
```diff
@@ -50,6 +50,14 @@ public abstract class MappedClosure<O,M extends MappedClosure<O,M>> implements S
         this.resultMap.putAll(inMap)
     }
 
+    public O remove(String p) {
+        return resultMap.remove(p)
+    }
+
+    public put(String k, O v) {
+        resultMap.add(k, v)
```
rsandell: So maybe actually using `put` on `resultMap` as well?

abayer: durrrrrrr brain

abayer:
@rsandell Yeah, turns out that the current code doesn't actually ever care about the old `Map` aspects of `MappedClosure` directly, thanks to not needing the closure transformation garbage any more. =)

PR #206, opened 2017-10-09
JENKINS-46403 - fix infinite loop running jobs with a logRotator

thomasgl-orange:
* JENKINS issue(s):
    * [JENKINS-46403](https://issues.jenkins-ci.org/browse/JENKINS-46403)
* Description:
    * a Declarative Pipeline job can live-lock on startup when the (deprecated) Job.logRotator field has a value
    * see [my explanation](https://issues.jenkins-ci.org/browse/JENKINS-46403?focusedCommentId=316154#comment-316154) in the JENKINS issue
* Documentation changes:
    * N/A
* Users/aliases to notify:
    * ?


abayer:
Re-running the tests since I think the issue was transient docker noise.

PR #208, opened 2017-10-20
[FIXED JENKINS-46597] Eradicate TreeMap from runtime

abayer:
* JENKINS issue(s):
    * [JENKINS-46597](https://issues.jenkins-ci.org/browse/JENKINS-46597)
* Description:
    * This was the last lurking bit of TreeMap, which almost certainly was causing the NotSerializableException. So let's get rid of it and switch to the ugly but functional list-of-pairs.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees 
    * @kshultzCB

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Tools.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Tools.groovy
@@ -55,16 +55,18 @@ public class Tools extends MappedClosure<Closure,Tools> implements Serializable 
     /**
      * Merges the tool entries from another instance into this one, defaulting to the current instance's values.
      *
-     * @return A map of type/name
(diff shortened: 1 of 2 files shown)


svanoort:
@abayer Why isn't it Serializing?  TreeMaps should serialize as long as their keys, values, and Comparator are all Serializable.

abayer:
@svanoort TreeMap.Entry isn't serializable. Annoying but true - the map itself does fine but if there's an entry floating around at save time, boom.

svanoort:
@abayer  ah, you're holding a reference to the Entry itself somewhere, ok, yeah, ugh

samirageb:
I don't know where to open an issue, but I added my comments to the bottom of this thread: https://issues.jenkins-ci.org/browse/JENKINS-46597

This issue is most definitely not resolved for us.  We continue to get the same exact Serialization error... Any assistance on this would be greatly appreciated.

PR #210, opened 2017-10-24
Get more idiomatic with Groovy, clean up some warnings

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * Some general code cleanup - I kept meaning to do these things but didn't want to include them in other PRs, so tada.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -82,6 +82,8 @@ import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
 import org.jenkinsci.plugins.workflow.steps.Step
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor
 import org.jenkinsci.plugins.workflow.support.steps.StageStep
+import org.kohsuke.accmod.Restricted
+import org.kohsuke.accmod.restrictions.NoExternalUse
 
 import javax.annotation.CheckForNull
 import javax.annotation.Nonnull
@@ -109,6 +111,7 @@ class Utils {
         }
     }
 
+    @Restricted(NoExternalUse.class)
     static Map<String,Closure> getCredsFromResolver(Environment environment, CpsScript script) {
         if (environment != null) {
             environment.credsResolver.setScript(script)
@@ -125,6 +128,7 @@ class Utils {
      * @return A non-null but possibly empty map of strings to {@link CredentialWrapper}s
      */
     @Nonnull
+    @Restricted(NoExternalUse.class)
     static Map<String, CredentialWrapper> getLegacyEnvCredentials(@Nonnull Environment environment) {
         Map<String, CredentialWrapper> m = [:]
         environment.each {k, v ->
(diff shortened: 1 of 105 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -697,19 +696,19 @@ public class Utils {
...
         for (int x = beginLine; x <= endLine; x++) {
-            String line = sourceUnit.source.getLine(x, null);
+            String line = sourceUnit.source.getLine(x, null)
             if (line == null)
-                throw new AssertionError("Unable to get source line"+x);
+                throw new AssertionError("Unable to get source line"+x)
 
             if (x == endLine) {
-                line = line.substring(0, endLineLastColumn - 1);
+                line = line.substring(0, endLineLastColumn - 1)
```
rsandell: Could potentially be cleaned a bit more using `str.take` and `str.drop`. Depending on your definition of cleaning :)

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -607,9 +614,9 @@ public void skippedStagesForFailure() throws Exception {
                 .logNotContains("I will be skipped", "I also will be skipped", "I have succeeded")
                 .go();
 
-        assertTrue(b.getExecution().getCauseOfFailure() != null);
-
         FlowExecution execution = b.getExecution();
+        assertNotNull(execution);
+        assertTrue(execution.getCauseOfFailure() != null);
```
rsandell: `assertNotNull` provides better failure messages. Or `assertThat(o, not(null()))`

abayer: durrr

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: I have never actually used those yet, so I'll stick with what I'm familiarish with. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -442,6 +454,7 @@ public class Utils {
      * @return The list of instances. May be empty.
      */
     @Nonnull
+    @Restricted(NoExternalUse.class)
```
rsandell: Normally there is no need to restrict package protected methods, but shouldn't hurt. But there is definitely no need to restrict a private method.
otoh this is groovy, so private/public modifiers are just for show :)

rsandell: Or no modifier in Groovy means `public`? I keep forgetting that :)
Anyway it's a good marker to have so others know to stfo ;)

PR #212, opened 2017-11-02
[FIXED JENKINS-47781] Narrow model lookups when appropriate

abayer:
* JENKINS issue(s):
    * [JENKINS-47781](https://issues.jenkins-ci.org/browse/JENKINS-47781)
* Description:
    * The original problem here is that if you just look up describables by the "upstream" symbol, you'll get one from Copy Artifact before you get `ReverseBuildTrigger`. Which...ok. Except that our syntax checking treats kicks in for the one from Copy Artifact! So let's narrow our model lookups when appropriate.
    * I don't *love* this solution - it could be smarter and have a better understanding of context. But this works for now, so I'll go with it.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
@@ -50,6 +50,7 @@ import org.jenkinsci.plugins.structs.describable.DescribableModel
 import org.jenkinsci.plugins.structs.describable.DescribableParameter
 import org.jenkinsci.plugins.workflow.flow.FlowExecution
 
+import javax.annotation.CheckForNull
 import javax.annotation.Nonnull
 
 /**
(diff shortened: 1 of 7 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -386,10 +387,26 @@ class ModelValidatorImpl implements ModelValidator {
...
         if (Jenkins.getInstance() != null) {
-            Descriptor desc = lookup.lookupFunctionFirstThenStep(meth.name)
             DescribableModel<? extends Describable> model
-            if (desc != null) {
-                model = lookup.modelForFunctionFirstThenStep(meth.name)
+
+            List<Class<? extends Describable>> parentDescribables = Utils.parentsForMethodCall(meth)
+
+            if (!parentDescribables.isEmpty()) {
+                model = parentDescribables.collect { p ->
```
rsandell: using `collectMany` should let you remove the `.find` at the end

abayer: *facepalm*

abayer: Actually, we still need that `find` to make sure we get the first non-null one we encounter. Eh, this code's ok. =)

rsandell: Then [findResult](http://docs.groovy-lang.org/2.4.3/html/groovy-jdk/java/util/Collection.html#findResult(groovy.lang.Closure)) is probably what you want :)

PR #216, opened 2017-11-09
[FIXED JENKINS-46854] Do parsing pass on script/expression blocks in JSON validation

abayer:
* JENKINS issue(s):
    * [JENKINS-46854](https://issues.jenkins-ci.org/browse/JENKINS-46854)
* Description:
    * This doesn't catch all possible script/expression errors that'll show up in full Groovy parsing and validation, but it'll catch errors like mismatched quotes in JSON form and report back their location and that compilation failed. Getting a more granular error would be too much text, so I opted for just the "hey, this wasn't valid Groovy syntax" approach.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
@@ -32,8 +32,8 @@ import hudson.tools.ToolDescriptor
 import hudson.tools.ToolInstallation
 import hudson.util.EditDistance
 import jenkins.model.Jenkins
-import org.codehaus.groovy.control.CompilationFailedException
 import org.codehaus.groovy.control.CompilationUnit
+import org.codehaus.groovy.control.MultipleCompilationErrorsException
 import org.codehaus.groovy.control.Phases
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter
 import org.jenkinsci.plugins.pipeline.modeldefinition.DescriptorLookupCache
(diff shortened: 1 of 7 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -359,7 +361,17 @@ class ModelValidatorImpl implements ModelValidator {
...
-            // No validation needed for code blocks like expression and script
+            // Verify that the code block can be parsed - we'll still get garbage for errors around class imports, etc,
+            // but you can't do that from the editor anyway.
+            String codeBlock = step.codeBlockAsString()
+            CompilationUnit cu = new CompilationUnit()
+            cu.addSource(step.name, codeBlock)
+            try {
+                cu.compile(Phases.PARSING)
+            } catch (CompilationFailedException e) {
+                errorCollector.error(step, Messages.ModelValidatorImpl_CompilationErrorInCodeBlock(step.name))
```
kzantow: Can we get a line number or something from the unused exception here?

abayer: You don't actually want the line number - that's from the converted-to-Groovy version. You *are* getting the JSON location, though.

abayer: (i.e., same as any other validation error on a JSON element)

kzantow: What about any other information? I'll show an error to the user but it should have some context about what went wrong so the user can correct it

abayer: See my note in the description - there's no way for me to get a more granular error than `startup failed` without digging into the full error string a la 
```
startup failed:
script: 1: unexpected char: ''' @ line 1, column 63.
    "Sleeping 1S"; sleep 1;' done'
                                 ^
```
If you want that full message, I'll happily give it to you, but I assumed you wouldn't want it due to the size.

kzantow: Yeah, would rather have the error than not.

abayer: Wait, I may have found a way...

abayer: Ok, the error message is now gonna be like:
```
Groovy compilation error(s) in script. Error(s): "unexpected char: '''"
```
In addition to the JSON location, of course.

michaelneale:
Nice 👍 

(I agree with summary of solution on the jira ticket and expect this will do nicely)

PR #217, opened 2017-11-09
[JENKINS-47928] Skip parallel stage post steps for previous errors

tadfisher:
Fixes [JENKINS-47298](https://issues.jenkins-ci.org/browse/JENKINS-47928)

Avoid running `post` steps for parallel stage parents that did not throw errors.


abayer:
Very interesting! I hadn't thought of that approach. Thanks - rerunning the build to see if the error was transient.

samirageb:
I currently run a single post stage for the entire pipeline, which has a single parallel stage with 6 child stages (no child post stages) and continue to get the treemap serialization error listed here: https://issues.jenkins-ci.org/browse/JENKINS-46597

In my case, I'm typically entering the parallel parent stage, but conditionally skipping the child parallel stages. (based on params) I often run into failed builds at the pipeline post stage.

Is my scenario unique? Or will this fix the error as well?

tadfisher:
@abayer Looks like a timeout issue with a Docker agent test? Not sure if this would have affected that.

abayer:
@tadfisher Yeah, it's probably something transient, but let me try it out locally tomorrow...

PR #220, opened 2017-11-20
[FIXED JENKINS-37663] Check execution result in post as well

abayer:
* JENKINS issue(s):
    * [JENKINS-37663](https://issues.jenkins-ci.org/browse/JENKINS-37663)
* Description:
    * The new junit step doesn't set the Run to UNSTABLE directly, it sets the context (and therefore the execution) result to UNSTABLE. We should be checking the execution result as well as the Run result.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
@@ -39,8 +39,7 @@ public class Aborted extends BuildCondition {
     @Override
     public boolean meetsCondition(WorkflowRun r) {
         Result execResult = getExecutionResult(r)
-        return (execResult != null && execResult.equals(Result.ABORTED)) ||
-            (r.getResult() != null && r.getResult().equals(Result.ABORTED))
+        return execResult == Result.ABORTED || r.getResult() == Result.ABORTED
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
```diff
@@ -38,7 +38,9 @@ import org.jenkinsci.plugins.workflow.job.WorkflowRun
 public class Aborted extends BuildCondition {
     @Override
     public boolean meetsCondition(WorkflowRun r) {
-        return r.getResult() != null && r.getResult().equals(Result.ABORTED)
+        Result execResult = getExecutionResult(r)
+        return (execResult != null && execResult.equals(Result.ABORTED)) ||
```
rsandell: You could clean this up a bit to make it more readable. Groovy performs a `.equals` bu default with the `==` operator. But that doesn't really matter since they should all be the same instance anyway, so an `==` operator should catch it in standard java as well.

```
return execResult == Result.ABORTED || r.getResult() == Result.ABORTED
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Changed.groovy
```diff
@@ -46,11 +47,13 @@ public class Changed extends BuildCondition {
         }
         // If the current build's result isn't null (i.e., it's got a specified status), and it's different than the
         // previous build's result, we're changed.
-        else if (r.getResult() != null && !prev.getResult().equals(r.getResult())) {
+        else if ((execResult != null && !prev.getResult().equals(execResult)) ||
```
rsandell: similar here: `(execResult != null && prev.result != execResult) || r.result != prev.result //r.result should always be != null`

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
abayer: Oh right, null safety for free. =)

PR #221, opened 2017-11-20
[FIXED JENKINS-48115] Work around badly configured metasteps

abayer:
* JENKINS issue(s):
    * [JENKINS-48115](https://issues.jenkins-ci.org/browse/JENKINS-48115)
* Description:
    * jira-steps has `isMetaStep()` returning true for all its step descriptors, which is...wrong. As a result, literally every symbol has at least one erstwhile metastep, breaking our logic for determining whether we're looking at a function or a describable. Sigh. While this should be fixed in jira-steps too, we should definitely be working around the super-broad cases like this.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ASTParserUtils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ASTParserUtils.groovy
@@ -422,7 +422,10 @@ class ASTParserUtils {
         Descriptor<? extends Describable> funcDesc = lookupCache.lookupFunction(methodName, descClass)
         StepDescriptor stepDesc = lookupCache.lookupStepDescriptor(methodName)
         // This is the case where we've got a wrapper in options
-        if (stepDesc != null || (funcDesc != null && !StepDescriptor.metaStepsOf(methodName).isEmpty())) {
+        if (stepDesc != null ||
+            (funcDesc != null &&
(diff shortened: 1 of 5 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ASTParserUtils.groovy
```diff
@@ -422,7 +422,10 @@ class ASTParserUtils {
         Descriptor<? extends Describable> funcDesc = lookupCache.lookupFunction(methodName, descClass)
         StepDescriptor stepDesc = lookupCache.lookupStepDescriptor(methodName)
         // This is the case where we've got a wrapper in options
-        if (stepDesc != null || (funcDesc != null && !StepDescriptor.metaStepsOf(methodName).isEmpty())) {
+        if (stepDesc != null ||
+            (funcDesc != null &&
+                // Deal with cases like jira-steps that claim to be metasteps without any metastep arguments!
+                StepDescriptor.metaStepsOf(methodName)?.first()?.metaStepArgumentType != Object.class)) {
```
jglick: Should this not be covered by the `workflow-step-api` fix? If not, why not?

abayer: Yup, I missed a commit.

rsandell:
:bee:

PR #222, opened 2017-11-21
[FIXED JENKINS-46252] Mark Declarative jobs with DeclarativeJobAction

abayer:
* JENKINS issue(s):
    * [JENKINS-46252](https://issues.jenkins-ci.org/browse/JENKINS-46252)
* Description:
    * If we are in a Declarative run, mark the parent job as a Declarative job via DeclarativeJobAction, if it's not already present. Note that we never remove this - if a job was ever once Declarative, we still count it as such. Imperfect, but close enough.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -33,6 +33,7 @@ import hudson.BulkChange
 import hudson.ExtensionList
 import hudson.model.Describable
 import hudson.model.Descriptor
+import hudson.model.Job
 import hudson.model.JobProperty
 import hudson.model.ParameterDefinition
 import hudson.model.ParametersDefinitionProperty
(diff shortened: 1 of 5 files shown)


abayer:
@rsandell Thought about that, decided I'd rather just keep it around forever and not have the additional overhead, especially since I think we may actually *want* it to be around forever.=)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -270,7 +272,7 @@ public class Utils {
         return nodes
     }
 
-    static void markExecutedStagesOnAction(CpsScript script, String astUUID) throws Exception {
+    static void updateRunAndJobActions(CpsScript script, String astUUID) throws Exception {
```
rsandell: should be `@Restricted` but eh.

abayer: I should do a whole chunk of that over in #210...

abayer: And I did! =)

PR #223, opened 2017-11-24
[FIXED JENKINS-48209] Prevent serialization of when expression closure

abayer:
* JENKINS issue(s):
    * [JENKINS-48209](https://issues.jenkins-ci.org/browse/JENKINS-48209)
* Description:
    * This shouldn't really matter, in truth - I believe we actually regenerate the whole Root object from scratch on resume, though I could be misremembering. Regardless, even having the when expression closure be transient, the durability test (with when expressions added) still works right.
    * As to why we need to do this in the first place? Because something weird could be out there trying to some non-standard serialization of Describables. Ok, spoiler, there definitely is such a thing, and it does a straight XStream serialization of every Describable. But we use Describables as more than just UI config stuff - we use them for syntax validation and such for our internal extension points, like when conditions. And if that XStream serialization of a when expression with a sh step kicks in, well, all hell breaks loose. This works around that.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/WhenStageTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/WhenStageTest.java
@@ -40,9 +40,13 @@ import net.sf.json.JSONObject;
 import org.jenkinsci.plugins.pipeline.modeldefinition.endpoints.ModelConverterAction;
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.Stage;
 import org.jenkinsci.plugins.pipeline.modeldefinition.when.ChangeLogStrategy;
+import org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditional;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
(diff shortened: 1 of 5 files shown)


abayer:
fwiw, the problematic pickler is in CloudBees Pipeline Templates, a proprietary plugin.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/WhenStageTest.java
```diff
@@ -386,4 +398,14 @@ protected boolean shouldExamineAllBuilds(@Nonnull SCMHead head) {
             return false;
         }
     }
+
+    @TestExtension
```
rsandell: shouldn't you isolate this to only be active for `whenExprDurableTask`?

abayer: Theoretically, we could do that, but I'm frankly fine with this being run on all the tests in here anyway.

abayer:
Hrm, can't reproduce the test failure locally. Fun.

abayer:
Oh. 'cos I'm not on Windows. Sigh.

PR #224, opened 2017-11-24
Refactoring to allow when condition plugins to not need p-m-d

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * Moving around various methods (and tweaking one of them to not need pipeline-stage-step) so that plugins like declarative-pipeline-when-conditions can avoid depending on pipeline-model-definition.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


pipeline-model-api/pom.xml
```diff
@@ -62,7 +62,18 @@
...
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>script-security</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-cps</artifactId>
```
rsandell: But now you are forcing users of api to get this one. Isn't that more bloated than forcing a dependency on declarative-api?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -173,27 +172,12 @@ class Utils {
         }
     }
 
+    /**
+     * Use {@link ASTParserUtils#isStageWithOptionalName} instead.
```
rsandell: `@deprecated Use {@link ASTParserUtils#isStageWithOptionalName} instead.` will make javadoc generation handle it a bit smarter.

abayer:
So fwiw, the idea here is to make it so that `pipeline-model-definition` can itself depend on `declarative-pipeline-when-conditions` in the future. I know more experimenting/fiddling is needed to actually get this to a point I'm happy with.

abayer:
Giving up on this for now.

PR #226, opened 2017-11-28
[FIXED JENKINS-48229] Use the agent's dir separator for Dockerfile path

abayer:
* JENKINS issue(s):
    * [JENKINS-48229](https://issues.jenkins-ci.org/browse/JENKINS-48229)
* Description:
    * I'm not sure how to test this - the issue occurs when the master's got different directory separators than the agent, which isn't something I can easily do in a unit test.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees


abayer:
@jglick I prefer using a variable someone else defined to make sure I get it right. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfileScript.groovy
```diff
@@ -72,15 +72,16 @@ class DockerPipelineFromDockerfileScript extends AbstractDockerPipelineScript<Do
 
     private Closure buildImage() {
         return {
+            def dockerfilePath = describable.getDockerfilePath(script.isUnix())
```
rsandell: assuming this is called when actually running on the node.

abayer: It is. I double checked. =) All of this script is run within the node block.

PR #227, opened 2017-11-29
[FIXED JENKINS-46809] Add sequential groups of parallel stages

abayer:
* JENKINS issue(s):
    * [JENKINS-46809](https://issues.jenkins-ci.org/browse/JENKINS-46809)
* Description:
    * Basic syntax and backend structure determined. Execution functional, seemingly. Initial test included. More to come!
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1274
* Users/aliases to notify:
    * @reviewbybees 

**_SYNTAX_**
```groovy
stages {
    stage('first-solo') { ... }
    stage('parent') {
        parallel {
            // Order in the parallel block is not significant.
            stage('single-stage') { ... }
            
            stage('multiple-stages') {
                stages {
                    // Order inside a group's stages is significant - we will execute the stages in the order 
                    // they're specified here.
                    stage('first-sequential-stage') { ... }
                    stage('second-sequential-stage') { ... }
                }
            }

            stage('other-single-stage') { ... }

            // etc - arbitrarily long list of either "stage" or "group"
        }
    }
    stage('second-solo') { ... }
}
```

**_RUNTIME "DIAGRAM"_**
Based on the above:
* stage 'first-solo' runs alone
* stage 'parent' starts after 'first-solo' finishes ->
    * runs the following branches in parallel ->
        * branch 'single-stage' ->
            * stage 'single-stage'
        * branch 'multiple-stages' ->
            * stage 'multiple-stages' (yes, there's a stage step here now - runs its multiple stages sequentially) ->
                * stage 'first-sequential-stage'
                * stage 'second-sequential-stage' runs after 'first-sequential-stage' finishes
        * branch 'other-single-stage' ->
            * stage 'other-single-stage'
* stage 'second-solo' runs after 'parent' (and all its nested stages) finishes

**_TESTS TODO_**
- [x] Compatibility tests for JSON containing `parallel` from before this change - should still translate and work the same. (get for free with pre-existing JSON/`parallel` tests!)
- [x] Compatibility tests for Groovy containing `parallel` from before this change - get for free as well. Hooray!
- [x] JSON<->Groovy conversion of new mixed `parallel` (i.e., `stage` and `group`) blocks.
- [x] Upgrade testing - what happens to a running build with old-school `parallel` definition, since that gets migrated behind the scenes? (EDIT: This will be fine - we haven't actually changed `parallel`, we've changed `stage` by adding more fields/functionality. So nothing needs to be tested here, and if it *does* go weird on upgrade while a build is running, well, we don't guarantee that for 1.x versions.)
- [x] Validate that nested stages within `group` can't have `parallel` within them.
- [x] Testing in Blue Ocean to make sure this doesn't Break All The Things visually in versions of Blue Ocean before visualization support for nested stage groups is added. (EDIT: It's not great, but it doesn't blow up horribly, so I'll roll with it.)
- [x] Verify that the new JSON representation is still valid (without any `group`s in the `parallel`) in the Pipeline Editor before it also is updated for nested stage groups. *EDIT*: This turns out to be a no-op. The JSON syntax if you don't include `stage`s with nested `stages` is the same, so...yay?
- [x] `post` behavior for both parent stages containing `parallel` `group`s and for the sequential stages within the `group`s.
- [x] `group` level `agent`
- [x] `group` level `tools`
- [x] `group` level `post`
- [x] `group` level `environment`
- [x] `group` level `when`

**_OTHER TODO_**
- [x] Docs PR (https://github.com/jenkins-infra/jenkins.io/pull/1274)
- [x] HTML help for `group` directive.
- [x] Do we want to add `when` for `group`s, so you can conditionally execute/not execute the whole `group` of `stage`s with one `when` declaration while still executing any other `group`s or `stage`s within the same `parallel`?
- [x] And that brings up the question of whether we should have an actual `stage` step execution for a `group`. I think that if we do `when` on `group`s, we should also nest a `stage` step directly within the `parallel` branch and run the sequential `stage`s within that `group` `stage`, so that things like marking as skipped and the like actually work right.
- [x] Make `group` basically a special case of `stage` where `steps` and `parallel` aren't allowed, a list of `stage`s is, and `agent`, `tools`, `when`, `post`, `environment` are all allowed. This will mean having an actual `stage` step execution for the group.
- [x] Move the list of `stage`s inside `group` into a nested `stages` block now that we actually have other possible fields.
- [x] Validation (and testing thereof) for rejecting empty `group`s (i.e., without `stages` or with empty `stages`). (EDIT: Realized we get this for free in `ModelASTStages` validation. woo!)
- [x] Switch `stage` to be able to contain one and only one of `steps`, `parallel`, or `stages` - i.e., merge together `stage` and `group`.


abayer:
Adding @bitwiseman for feedback as well.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
```diff
@@ -26,18 +27,34 @@
...
 
     public ModelASTStage(Object sourceLocation) {
         super(sourceLocation);
     }
 
+    protected Object readResolve() throws IOException {
+        // If there's already a set of parallel stages defined, add that to the parallel content instead.
+        if (this.parallel != null) {
+            this.parallelContent.addAll(this.parallel.getStages());
+            this.parallel = null;
```
rsandell: So `parallel` should now be `transient` and `@Deprecated`?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
```diff
@@ -53,22 +53,43 @@ class Stage implements Serializable {
...
+    Stage(String name, StepsBlock steps, Agent agent, PostStage post, StageConditionals when, Tools tools,
+          Environment environment, boolean failFast, List<ParallelContent> parallelContent) {
         this.name = name
         this.steps = steps
         this.agent = agent
         this.post = post
         this.when = when
         this.tools = tools
         this.environment = environment
-        this.parallel = parallel
```
rsandell: I think the same goes for this one.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -700,6 +700,47 @@ class RuntimeASTTransformer {
...
 
+    /**
+     * Generates the AST (to be CPS-transformed) for instantiating a list of {@link ParallelContent}.
+     *
+     * @param original The parsed AST model of a stage
+     * @return The AST for a list of {@link ParallelContent}, or the constant null expression if the original
+     * cannot be transformed.
+     */
+    Expression transformParallelContent(@CheckForNull ModelASTStage original) {
+        if (isGroovyAST(original) && !original.parallelContent?.isEmpty()) {
```
rsandell: #210 was merged only 7 days ago :) 
`!original.parallelContent?.isEmpty()` can be shortened to just `!original.parallelContent` iirc "Groovy truthiness" checks for both `null` and non empty for lists.
But I could be remembering wrongly ;)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -700,6 +700,47 @@ class RuntimeASTTransformer {
...
+
+    /**
+     * Generates the AST (to be CPS-transformed) for instantiating {@link ParallelGroup}.
+     *
+     * @param original The parsed AST model
+     * @return The AST for a constructor call for {@link ParallelGroup}, or the constant null expression if the original
+     * cannot be transformed.
+     */
+    Expression transformParallelGroup(@CheckForNull ModelASTParallelStageGroup original) {
+        if (isGroovyAST(original) && !original.stages?.stages?.isEmpty()) {
```
rsandell: same here

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
abayer: Good point. `getParallel()` is deprecated already, and I've re-added `setParallel(ModelASTStages)` solely so I can deprecate it. =) Adding `transient` and `@Deprecated` on the field too.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
abayer: boo, this was actually older code that I copy-pasted into an older iteration of this. Cleaning up!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
abayer: Gotcha.

abayer:
> I also think it makes sense for a group to have when conditions, so You don't have to repeat the same conditions in every group's stage.
Interesting. Do we really want `when` as a possibility on all of the parent `stage`, the `group`, and the individual sequential `stage`s?  I'm not ruling it out, I'm just not sure it's worth the additional parsing/model pain. Hrm. I'll add a note to the TODO to decide that.

abayer:
cc @rpocase and @stephendonner - this all may be of interest to you as well. And I'm looking for any feedback I can get, so hey. =)

rpocase:
:+1: From a current syntax perspective. I don't love the name `group`, but I'm struggling with coming up with anything more fitting that isn't overly verbose.

Another possible TODO decision (that might depend on `when`s on `group` TODO) is if `group` should support post. I can think of a use case where I want something along the lines of

```
group('cluster1') {
    ...
    post { failure { rollBack() } }
}
group('cluster2') {
    ...
    post { failure { rollBack() } }
}
```

This could be handled at the sequential stage level, but I could imagine scenarios where having at the group level reduces repetition.

abayer:
Upon further reflection, yeah. We should make `group` a special case of `stage` - no `steps` or `parallel` allowed in it, 1..n `stage`s required, but `agent`, `tools`, `environment`, `when`, and `post` are all available. 

Here's how I got there: first, I realized we needed to be able to have one `agent` for all the `stage`s in a `group` - otherwise, you can't be sure you'll get the same actual agent and workspace, which could be relevant. And then I started thinking about my eventual dream of a `matrix` directive - and that wants not just `agent` but `when` and `post` too. So if we're going that far, hey, might as well just make `stage` and `group` inherit from a common ancestor that provides `agent`, `tools`, `environment`, `when`, `post`, and a name field. So that'll take some refactoring and fiddling with syntax, but I think it's the right thing to do.

rpocase:
Sounds great! I was leaning towards `group` really feeling like a stage, but wasn't sure if the rest of the directives were too much. 

There is still room for each `stage` contained within a `group` to require specific agents, though. Could group have an implied `agent none` if not present? It's a small enough amount of noise to not be terribly obstructive when I don't need a `group` level `agent`, but would be a nice convenience.

abayer:
Yeah, none of `agent` et al would be required on `group`.

abayer:
ping @cliffmeyers too from the Blue Ocean perspective.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
```diff
@@ -670,6 +670,14 @@ public void parallelStagesDeepNesting() throws Exception {
...
     }
 
+    @Issue("JENKINS-46809")
+    @Test
+    public void parallelStagesGroupsDeepNesting() throws Exception {
+        expectError("parallelStagesGroupsDeepNesting")
+                .logContains(Messages.ModelValidatorImpl_NoNestedWithinNestedStages())
+                .go();
+    }
+
```
kshultzCB: I'd love to see a similar test, which checks the ability to override/inherit `agent`, `tools`, `environment`, `when`, and `post`

abayer: See above in the **TEST TODO**s =)

kshultzCB: Oh, right. That. :)

rtyler:
@rpocase @abayer how does the step name `sequence` strike you both instead of `group`? I think `group` is too ambiguous to accurately convey what's happening within the block.

bitwiseman:
@abayer 
Would this syntax (with `group` directly under `stages` also be allowed? 

```groovy
stages {
    stage('Build') { ... }
    group('Release Train') {
        when { branch 'master' }
        stage("Confirm deploy to Staging") { ... }
        stage("Deploy to Staging") { ... }
        stage("Confirm deploy to Production") { ... }
        stage("Deploy to Production") { ... }
    }
}
```

@rtyler  I mostly agree with `sequence`, especially when under `parallel` but it makes less sense here.   

@abayer  It sounds like `stage` and `group` are basically identical.  
What about having each stage allow only  one of the following: `steps`, `parallel`, or `stages`?  (A reasonable alternative to `stages` would be `sequence`.)

abayer:
The way I'm coding it up is that `Stage` and `ParallelGroup` (the runtime model classes) both inherit from `AbstractParallelContent`, with `AbstractParallelContent` having `agent` et al, `Stage` having `steps`, `parallel`, and `failFast`, and `ParallelGroup` having a list of `stage`s.

I'm not sure what the point of `group` directly under `stages` (or elided into `stage`) would be - the whole point is that you can have 2..N `group`s or `stage`s within a `stage`'s `parallel`. It's not meant to do "nesting" of a sequential list of `stage`s within a `stage` without parallelism involved.

abayer:
@rtyler I don't hate `sequence`. I don't love it either, but I don't hate it. I almost wonder if `stages` would be right, but that'd create some confusion since `parallel`->`stages` could have `agent`, `post`, etc, while `stages` at the root can't...

bitwiseman:
@abayer 
    I'm not sure what the point of group directly under stages

In this case, so  that we can do a `when` clause once on the `group` instead of on multiple stages.

I don't understand what you said about the runtime classes. I'm only talking about the user facing syntax.  A `stage` can currently have either `steps` or `parallel`.   If you allow a `stage` to have either `steps`, `parallel`, or `stages`, that (IMO) has is the least disruptive path.  It also keeps the syntax consistent as it nests - `pipeline -> stages` is serial execution of `stage` items, `stage -> stages` is the same.

bitwiseman:
@abayer I get what you're saying about how you meant to use it, and I don't want to punish you for implementing this feature (which is my personal number 1 desired feature for declarative). 

Maybe what I'm saying is if you open the box of allowing `when`, etc. on `group`, go all the way.  Otherwise, it's going to cause more confusion and frustration.

Also if you allow `group` to have `when`, etc., then group is no long a list like `parallel` or `stages` and there needs to be a `group -> stages`.

bitwiseman:
@abayer  
    but that'd create some confusion since parallel->stages could have agent, post, etc, while stages at the root can't...

Huh?  Example please.

abayer:
@bitwiseman =)

i386:
Liking where this is going but we need to get a designer on how this will work in the Editor before committing to this.

abayer:
Got on the same page with @bitwiseman - the idea is to possibly make `stage` able to contain one-and-only-one-of `steps`, `parallel`, or `stages`. I just moved the `stage` list in `group` to a nested `stages` block now that we have other fields that could be used in `group` (`agent` etc), so what's being proposed is basically this:
* At the top level, a `stage` could contain `steps`, `parallel`, or `stages` - in the last case, that would mean a parent stage step containing a sequentially executed list of individual stages, no parallelism involved.
* Inside a `parallel`, a `stage` could contain `steps` or `stages` - in the last case, that would mean a parent stage step containing a sequentially executed list of individual stages, with that parent stage step run in parallel with any other `stage`s defined inside the `parallel`.
* Inside a `stages` (that is, a `stage(...) { stages { ... } }`), a `stage` could only contain `steps`. No arbitrarily deep nesting of parallelization or nested sequential lists.

The upsides to this approach are that it's simpler syntactically (albeit a bit confusing due to how many different ways you can use `stage`, and particularly for the double usage of `stages`, which we might want to rename in the "sequential list inside a stage" scenario), a sequential stage list within a `parallel` has the same syntax as a sequential stage list at the top level, and, here's where things get interesting, it'd let you do things like say "run this list of stages on a different agent than the top level one, but with all of this list of stages running in the same agent+workspace", since the `node` block for the parent would have all of the `stage->stages` list run inside it. That opens up some very interesting possibilities. I'll write more on that tomorrow when I haven't been at the computer for ~11 hours already. =)

bitwiseman:
Here's sample code that I think conveys what @abayer said above.
```groovy
pipeline {
    agent none
    // list of sequentially executed stages
    stages {
        stage('first-solo') { 
            steps { ... } 
        }
        stage('parent') {
            parallel {
                // Order in the parallel block is not significant.
                stage('single-stage') { steps { ... } }

                // It does need to take a string, though, so we have a name for the parallel branch 
                // it creates to contain its sequential stages. 

                stage('sequential-stages') {
                    // list of sequentially executed stages
                    stages { 
                        // Order inside a stages is significant - we will execute the stages in the order 
                        // they're specified here.
                        stage('first-sequential-stage') { steps { ... } }
                        stage('second-sequential-stage') { steps { ... } }
                    }
                }

                stage('other-single-stage') { steps { ... } }

                // etc - arbitrarily long list of either "stage" or "group"
            }
        }
        stage('grouped-stages') {
            stages {
                stage('first-grouped') { steps { ... } }
                stage('second-grouped') { steps { ... } }
            }
        }
    }
}
```

rsandell:
The overusage of `stage` and `stages` is a little worrying to me though; in terms of readability. You can't easily glance over the example @bitwiseman showed and understand what is happening. The key to understand the context of all the stages is a very small `parallel` that is easy to miss. Using another word, like `group`, makes it visually easy to grasp where the differences in behavior lie when reading it.

Writing it might be easier since the names are consistent with what they do when you write them. It's the other end of the coin I'm concerned with. Like we are trying to make another [Perl](https://en.wikipedia.org/wiki/Write-only_language) ;)

abayer:
Yeah, I don't love the terminology - using `stage` everywhere is one thing (since in this approach, whether it has a "group", a "parallel", or "steps", it's always actually a stage), but using the term `stages` inside `stage` doesn't quite feel right to me. Two reasons: one is the duplication of terminology (yes, `stages` at the top level and `stages` within a `stage` do more or less behave the same, but it does feel weird to have `stages { stage(...) { stages { stage(...) {} } } }`, and the second is that I'd kind of prefer to have the sibling to `parallel` and `steps` be more distinctive than just `stages`. Yes, we know that probably means the same thing (a sequential list of stages to run) at the `stage` level as at the top level, but I feel like a name that more clearly states the purpose in the `stage` context would be better. Maybe `sequence` after all...

bitwiseman:
I understand the objections.  So, let me throw a this alternative out there.

```groovy
pipeline {
    agent none
    // list of stages executed sequentially
    stages {
        stage('first-solo') { 
            // list of steps executed in sequentially
            steps { ... } 
        }
        stageGroup('parent') {
            // list of stages executed in parallel
            parallel {
                // Order in the parallel block is not significant.
                stage('single-stage') { steps { ... } }

                // It does need to take a string, though, so we have a name for the parallel branch 
                // it creates to contain its sequential stages. 

                stageGroup('sequential-stages') {
                    // list of stages executed sequentially
                    stages { 
                        // Order inside a stages is significant - we will execute the stages in the order 
                        // they're specified here.
                        stage('first-sequential-stage') { steps { ... } }
                        stage('second-sequential-stage') { steps { ... } }
                    }
                }

                stage('other-single-stage') { steps { ... } }

                // etc - arbitrarily long list of either "stage" or "group"
            }
        }
        stageGroup('grouped-stages') {
            // list of stages executed sequentially
            stages {
                stage('first-grouped') { steps { ... } }
                stage('second-grouped') { steps { ... } }
            }
        }
    }
}
```

Internally, `stageGroup` will be the same as `stage` but syntactically, a `stage` will have `steps`, and a `stageGroup` will have `parallel` or `stages`.   We can _allow_ `stage` to continue to have `parallel` but make it deprecated.  I know we just got there, but this is what happens when you develop iteratively and don't have future additions mapped out.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/AbstractModelASTParallelContent.java
```diff
@@ -0,0 +1,244 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * Represents an individual Stage and the {@link ModelASTBranch}s it may contain.
+ *
+ * @author Kohsuke Kawaguchi
+ * @author Andrew Bayer
+ * @see ModelASTPipelineDef
+ */
+public abstract class AbstractModelASTParallelContent extends ModelASTElement {
```
rsandell: Bad name, it indicates that classes extending this will have parallel content. But the actual parallel stuff is implemented someplace else.
Should be some thing like `AbstractASTPseudoStage` or something.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
rsandell: It is very strange to add a method just to deprecate it :/

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
```diff
@@ -281,24 +170,24 @@ public ModelASTStages getParallel() {
         return parallel;
     }
 
-    public List<ModelASTParallelContent> getParallelContent() {
+    @Deprecated
+    public void setParallel(ModelASTStages s) {
```
rsandell: eehh? :confused:

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/AbstractParallelContent.groovy
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * An individual stage or group that can be used inside a stage's parallel block
+ *
+ * @author Andrew Bayer
+ */
+@ToString
+@EqualsAndHashCode
+@SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
+abstract class AbstractParallelContent implements Serializable {
```
rsandell: same naming issue here I think.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -302,10 +302,52 @@ class ModelParser implements Parser {
...
+                } else {
+                    def name = parseMethodName(mc)
+
+                    // Here, method name is a "section" name in the "group" closure, which must be unique - other than
+                    // "stage"
+                    if (name != "stage" && !sectionsSeen.add(name)) {
+                        // Also an error that we couldn't actually detect at model evaluation time.
+                        errorCollector.error(r, Messages.Parser_MultipleOfSection(name))
+                    }
+                    // TODO: Might be able to combine some of this with the equivalent in parseStage
```
rsandell: yepp, please try :)

pipeline-model-definition/src/test/resources/postStage/groupLocalAll.groovy
```diff
@@ -0,0 +1,88 @@
...
+                            echo "I AM UNSTABLE"
+                        }
+                    }
+                }
+            }
+        }
+    }
+    post {
+        always {
+            echo "And AAAAIIIAAAIAI"
```
rsandell: It's still around :rofl:

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -194,69 +212,100 @@ class ModelInterpreter implements Serializable {
...
+                            inDeclarativeAgent(stageOrGroup, root, stageOrGroup.agent) {
+                                if (evaluateWhen(stageOrGroup.when)) {
+                                    withCredentialsBlock(stageOrGroup.environment) {
+                                        withEnvBlock(stageOrGroup.getEnvVars(script)) {
+                                            toolsBlock(stageOrGroup.tools, stageOrGroup.agent ?: root.agent, parent?.tools ?: root.tools) {
+                                                if (asStage) {
+                                                    // Execute the actual stage and potential post-stage actions
+                                                    executeSingleStage(root, asStage, parentAgent)
+                                                } else if (asGroup) {
+                                                    evaluateSequentialStages(root, asGroup.stages, firstError, asGroup).call()
```
rsandell: I might have missed it in my previous read through, but have you added validation and tests for empty groups?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/AbstractModelASTParallelContent.java
abayer: Yeah, I am emphatically not in love with the name. =)

abayer: ...fwiw, this will probably get renamed to `ModelASTAbstractStage` or something like that, as we generalize this so that you can actually do a top-level `stage` with a group within it.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/AbstractParallelContent.groovy
abayer: Ahyup.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
abayer: Ahyup. =)

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
abayer: I swear there was a reason for this...probably demented worrying about in-progress-runs during upgrade. Which I think we can disregard, since I seem to have decided that we don't guarantee in-progress runs will work when upgrading a minor version (i.e., we broke that for the 1.1.x to 1.2.x upgrades, so we can do it for 1.2.x to 1.3.x, especially since this would be far less likely to cause problems than the 1.1.x->1.2.x upgrade and I never heard any complaints about that not working anyway).

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: I have not. Good point.

pipeline-model-definition/src/test/resources/postStage/groupLocalAll.groovy
abayer: I copy-paste a lot. =)

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/AbstractModelASTParallelContent.java
abayer: Actually, this gets stupid simpler once we just make it so that a `stage` can contain one of `steps`, `parallel`, or `stages` (or whatever we rename `stages` to in this context), since then it's just one class anyway.

abayer:
Ok, something went weird in that last commit and now not everything's getting its source location removed. Digging.

abayer:
Oh, it was stupid. I inadvertently deleted a `super.removeSourceLocation()` in `ModelASTStage#removeSourceLocation()` when combining stage and group. Oops.

hendrikhalkow:
Hi all, before I saw your work, I created this JIRA ticket:
https://issues.jenkins-ci.org/browse/JENKINS-48422
```groovy
    parallel {
        stages {
            stage('Build on Linux') {
                steps {
                    echo 'Building on Linux'
                }
            }
            stage('Test on Linux') {
                steps {
                    echo 'Testing on Linux'
                }
            }
        }
        stages {
            stage('Build on macOS') {
                steps {
                    echo 'Building on macOS'
                }
            }
            stage('Test on Linux') {
                steps {
                    echo 'Testing on macOS'
                }
            }
        }
    }
```
Please consider omitting the 'stageGroup' level so that there is just 'stages'. I am not that deep in the implementation details of this, but if there needs to be something between 'parallel' and 'stages', the best term would be 'sequential' instead of 'stages':

```groovy
    parallel {
        sequential {
            stages {
                stage('Build on Linux') {
                    steps {
                        echo 'Building on Linux'
                    }
                }
                stage('Test on Linux') {
                    steps {
                        echo 'Testing on Linux'
                    }
                }
            }
        }
        sequential {
            stages {
                stage('Build on macOS') {
                    steps {
                        echo 'Building on macOS'
                    }
                }
                stage('Test on Linux') {
                    steps {
                        echo 'Testing on macOS'
                    }
                }
            }
        }
    }
```
However, the first one would be my favourite.

sheerun:
I support not "group" or "sequential" but just using "stages", just like on top-level.

abayer:
@hendrikhalkow @sheerun See the updated description of the PR - there's an example in there with the current syntax, which is `stages`. 

Basically, `stage` (with this PR) will be required to contain one and only one of the following sections: `steps`, `parallel`, or `stages`. In a perfect world, I'd still use something other than `stages` within `stage` due to `stages { stage(...) { stages { stage(...) { ... } } } }` having more duplication than I like, but *shrug*, I haven't come up with anything better, so it's likely to stay `stages`. 

You'll be able to use `stage(...) { stages { ... } }` in either a top-level stage (i.e., within `pipeline { stages { stage(...) { stages { ... } } } }`) or in a nested stage within a `parallel` in a top-level stage (i.e., `pipeline { stages { stage(...) { parallel { stage(...) { stages { ... } } } } }`). You won't be able to use `parallel` anywhere but in a top-level stage, and you won't be able to use `stages` within a nested stage within a `stages` block in another stage. Phew. 

`stage(...) { stages { ... } }` will behave a lot like `stage(...) { parallel { ... } }` - the parent stage there can't have an `agent` or `tools`, since it's just a container for its nested stages.

steven-foster:
Is it possible to assign an `agent` etc to the "new" stages block, or is that outside the scope of this addition?

abayer:
@steven-foster Oh whooooops, you're right. We actually *do* want the `agent` supported on the sequential stage parent. I'll roll that back in a bit. Sorry for the confusion. =)

rpocase:
I agree with the original proposal for a new keyword (`sequence` has been my favorite suggestion so far) mostly due to readability. I think Identifying scope of `stages` is going to be very taxing for job implementors/code review. I hate to add the new keyword since there are already so many, but clarity is much more important to me here.

abayer:
@rpocase - Yeah, the best argument for `stages` I've heard so far is that the behavior of top-level `stages` is basically the same as what we're talking about within `stage`, but I'm with you: it feels weird. 

If we could get a consensus around `sequence`, I'd change everything right now. =)

abayer:
Boy howdy do I have some merge conflicts to resolve here. =)

ddfreiling:
Would love to see this land for our Jenkins pipeline, any ETA on this merging?

Hokwang:
I am looking forward committing this PR. Please consider 1.2.10 not 1.3 milestone.

abayer:
@rsandell @svanoort @batmat @bitwiseman @kshultzCB - if any of you would like to do a review (or re-review) here, it'd be appreciated. Looking to get this and #228 rolling towards release in the next week or two.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -421,20 +477,30 @@ class ModelInterpreter implements Serializable {
         parameters
     }
 
+    /**
+     * Legacy version to pass the root tools in, rather than directly passing in a tools. Only relevant for in-progress
+     * runs.
+     * TODO: Delete in 1.4? Or maybe just nuke now.
```
rsandell: Don't know either. I don't remember how it behaves when restoring an old running pipeline with this new code. If the old code is what is run or not.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/EnvironmentTest.java
```diff
@@ -106,6 +106,17 @@ public void environmentInStage() throws Exception {
...
+    @Test
+    public void environmentInGroup() throws Exception {
+        expect("environmentInGroup")
+                .logContains("[Pipeline] { (foo)",
+                        "Solo: FOO is BAZ",
+                        "First in group: FOO is BAR",
+                        "Second in group: FOO is BAH")
+                .go();
+    }
+
```
rsandell: credentials in group?

pipeline-model-definition/src/test/resources/errors/topLevelStageGroupsDeepNesting.groovy
```diff
@@ -0,0 +1,52 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
```
rsandell: It's a new year, a new life, a new dawn...

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -446,6 +512,7 @@ class ModelInterpreter implements Serializable {
             } else {
                 toolEnv = actualToolsInstall(toolsList)
             }
+            System.err.println("TOOL ENV: ${toolEnv}")
```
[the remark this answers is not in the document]
abayer: Done.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -421,20 +477,30 @@ class ModelInterpreter implements Serializable {
...
-    def toolsBlock(Agent agent, Tools tools, Root root = null, Closure body) {
+    def toolsBlock(Tools tools, Agent agent, Tools rootTools, Closure body) {
         def toolsList = []
         if (tools != null) {
-            toolsList = tools.mergeToolEntries(root?.tools)
-        } else if (root?.tools != null) {
-            toolsList = root.tools.mergeToolEntries(null)
+            toolsList = tools.mergeToolEntries(rootTools)
+        } else if (rootTools != null) {
+            toolsList = rootTools.mergeToolEntries(null)
```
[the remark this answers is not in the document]
abayer: Nope - if there are stage-level tools defined, we merge the root tools into them. Otherwise, we merge null into the root tools. It works. =)

pipeline-model-definition/src/test/resources/errors/topLevelStageGroupsDeepNesting.groovy
abayer: Ah, but it wasn't when I created that file. =)

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/EnvironmentTest.java
abayer: Good idea.

pipeline-model-definition/src/test/resources/errors/incorrectNestedStagesNPE.groovy
```diff
@@ -0,0 +1,62 @@
...
+                            steps {
+                                sh "ls file-2"
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
```
jsoref: Can you add a newline at the end to avoid Ø?

abayer: Done!

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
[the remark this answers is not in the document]
rsandell: Hmm, from what I can see `rootTools` is either the tools definition from the `parent` or from `root` not both, and I don't see any of the merge results being "written" anywhere. :confused:

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/Messages.properties
```diff
@@ -26,3 +26,4 @@ DirectiveGenerator.DeclarativeDirectivesLink.displayName=Declarative Directive G
 DirectiveGenerator.DeclarativeOnlineDocsLink.displayName=Declarative Online Documentation
 StageDirective.Steps.name=Steps
 StageDirective.Parallel.name=Parallel Stages
+StageDirective.Stages.name=Sequential Stages
```
jsoref: Can you add a newline at the end to avoid Ø?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
```diff
@@ -113,6 +134,15 @@ public void validate(final ModelValidator validator, boolean isNested) {
         if (input != null) {
             input.validate(validator);
         }
+        if (stages != null) {
+            stages.validate(validator, true);
+        }
+        for (ModelASTBranch branch : branches) {
+            branch.validate(validator);
+        }
+        for (ModelASTStage content: parallelContent) {
```
jsoref: probably add a space before `:` to match the previous block

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -575,11 +575,24 @@ class ModelParser implements Parser {
                             stage.environment = parseEnvironment(s)
                             break
                         case 'parallel':
-                            stage.parallel = parseStages(s)
+                            def parallelStmt = matchBlockStatement(s)
+                            if (parallelStmt==null) {
```
jsoref: probably spaces around `==`

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/Messages.properties
```diff
@@ -123,7 +123,7 @@ ModelValidatorImpl.InvalidUnnamedParameterType=Expecting "{0}" but got "{1}" of
 ModelValidatorImpl.NoSteps=No steps specified for branch
 ModelValidatorImpl.RequiredSection=Missing required section "{0}"
 ModelValidatorImpl.NoStageName=Stage does not have a name
-ModelValidatorImpl.NothingForStage=No "steps" or "parallel" to execute within stage "{0}"
+ModelValidatorImpl.NothingForStage=No "steps", "stages", or "parallel" to execute within stage "{0}"
```
jsoref: `Expected one of "steps", "stages", or "parallel" for stage "{0}"`

(__No__ _item series_ + __or__ _item_`...` is a mess in English grammar)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/Messages.properties
```diff
@@ -142,7 +142,7 @@ ModelValidatorImpl.NestedWhenNoArgs=No additional arguments are allowed for chil
 ModelValidatorImpl.NoNestedWhenAllowed=No children when conditions are allowed for when condition "{0}".
 ModelValidatorImpl.NestedWhenWithoutChildren=Nested when condition "{0}" requires at least one child condition.
 ModelValidatorImpl.NestedWhenWrongChildrenCount=Nested when condition "{0}" requires exactly {1} child condition(s).
-ModelValidatorImpl.BothStagesAndSteps=Only one of "parallel" or "steps" allowed for stage "{0}"
+ModelValidatorImpl.TwoOfStepsStagesParallel=Only one of "parallel", "stages", or "steps" allowed for stage "{0}"
```
jsoref: This is better (compare above)

Note to self: The periods in the nearby lines are probably bugs -- or indications of bugs...

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -138,34 +128,82 @@ class ModelInterpreter implements Serializable {
...
+        }
+    }
+
+    /**
+     * Get the map to pass to the parallel step of nested stages to run in parallel for the given stage.
+     *
+     * @param root The root of the Declarative model
+     * @param parentAgent The parent agent definition, which can be null
+     * @param thisStage The current stage we'll look in for parallel stages
+     * @param firstError An error that's already occurred earlier in the build. Can be null.
```
jsoref: The way this is written contrasts with `parentAgent` two lines earlier

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -138,34 +128,82 @@ class ModelInterpreter implements Serializable {
...
                     }
                 })
             } else if (skippedForWhen) {
-                parallelStages.put(parallelStage.name, {
-                    script.stage(parallelStage.name) {
-                        Utils.logToTaskListener("Stage '${parallelStage.name}' skipped due to when conditional")
-                        Utils.markStageSkippedForConditional(parallelStage.name)
+                parallelStages.put(content.name, {
+                    script.stage(content.name) {
+                        Utils.logToTaskListener("Stage '${content.name}' skipped due to when conditional")
```
jsoref: could `when` be marked up in some way?

pipeline-model-definition/src/test/resources/json/agentOnGroup.json
```diff
@@ -0,0 +1,81 @@
...
+    ]
+  }],
+  "agent":   {
+    "type": "label",
+    "argument":     {
+      "isLiteral": true,
+      "value": "some-label"
+    }
+  }
+}}
```
jsoref: Can you add a newline at the end to avoid Ø?

-- There may be others, best if your editor always includes them...

abayer: Yeah, this was a copy-paste case - my editor catches them normally.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: I try to avoid markup in the console log.

abayer:
Alrightie, I am finally merging this. =)

ader1990:
Hello,

Is there a working example of this feature?
I see more than one version of this feature in the above comments.

Thanks.

steven-foster:
there's a good blogpost about it here
https://jenkins.io/blog/2018/07/02/whats-new-declarative-piepline-13x-sequential-stages/

note that blue ocean does not support visualization of the stages separately yet, but it will still work properly. I'm personally waiting for that before adopting it.

nabertrand:
There's an extra number in the JIRA link above. Should be https://issues.jenkins-ci.org/browse/JENKINS-38442

PR #228, opened 2017-12-10
[JENKINS-45455] Restart from a specific stage in a previous build

abayer:
* JENKINS issue(s):
    * [JENKINS-45455](https://issues.jenkins-ci.org/browse/JENKINS-45455)
* Description:
    * Adds the ability to restart a Declarative Pipeline from any top-level stage that actually ran, once the initial Pipeline has completed. This also provides the option to preserve `stash`es from a defined number of completed builds, so that `stash`ed artifacts can be used in a restarted build.
    * Some **TODO**s:
- [x] Finish fleshing out test coverage - see `RestartDeclarativePipelineActionTest` for list.
- [x] Need to apply `NotExecutedNodeAction` to the stages we skipped due to restart.
- [x] Need an icon for `RestartDeclarativePipelineAction`. Ha.
- [x] Help blurb for `index.jelly`
- [x] Help html for `stageName`.
- [x] Form validation for `stageName` - having some interesting challenges getting this set up and testable, since `Action` isn't `Describable` so the normal magic form tags don't work the same way. (EDIT: Figured it out yaaaay) (EDIT AGAIN: Wait, why the hell was I doing this as a `textarea` in the first place? That's dumb. I'm dumb. It's a `select` now.)
- [x] jenkins.io documentation
- [x] Javadoc all the things.
- [x] Eventually work on Blue Ocean side.
- [x] Switch merge target back to `master` once #227 is merged.
* Documentation changes:
    * Uh, yeah, that'll be needed.
* Users/aliases to notify:
    * ...


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
```diff
@@ -179,9 +179,17 @@ public void simpleRestart() throws Exception {
...
         text.setText("restart");
-        HtmlPage redirect = j.submit(form);
-
-        assertEquals(p.getAbsoluteUrl(), redirect.getUrl().toString());
+        try {
+            HtmlPage redirect = j.submit(form);
+
+            assertEquals(p.getAbsoluteUrl(), redirect.getUrl().toString());
+        } catch (Exception we) {
+            // Annoying 404 shows up for http://localhost:60117/jenkins/static/05fb3e5f/plugin/pipeline-stage-view/jsmodules/stageview.js
```
abayer: @svanoort - So for some reason calling `j.submit(form)` results in a 404 for the above URL (port and `static/.../` path variable, of course). It doesn't actually break the submission, it just throws an exception. I'm stuck with stage view getting sucked in via jira-steps's dependency on workflow-aggregator (siiiiiigh) because I need to be able to test the dementedness jira-steps 1.3.0 did with metasteps, so I'm not sure if there's a way around this. Any ideas? If not, I'll probably sadly remove the jira-steps test and dependency.

abayer:
(Yes, I'm rebasing on top of the latest from #227 whenever that changes - I want to keep the history for this PR distinct from that for #227)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartFlowFactoryAction.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public CpsFlowExecution create(FlowDefinition def, FlowExecutionOwner owner, List<? extends Action> actions) throws IOException {
+        Run original = Run.fromExternalizableId(originRunId);
+        String origScript = null;
+        boolean origSandbox = true;
+        if (original instanceof FlowExecutionOwner.Executable) {
+            FlowExecutionOwner originalOwner = ((FlowExecutionOwner.Executable) original).asFlowExecutionOwner();
+            try {
+                // TODO: Need to add logic for stashes and a copier for environment, maybe.
```
jglick: `StashManager.CopyStashesAndArtifacts` and `EnvActionImpl.Copier` FWIW

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    @Extension
+    public static final class SaveStashes extends StashManager.StashBehavior {
+
+        @Override
+        public boolean shouldClearAll(@Nonnull Run<?,?> build) {
+            if (build instanceof WorkflowRun) {
+                WorkflowRun r = (WorkflowRun)build;
```
jglick: You do not actually need to cast at all AFAICT.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
```diff
@@ -0,0 +1,122 @@
...
+    @Extension
+    public static final class StashClearingListener extends RunListener<Run<?,?>> {
+        @Override
+        public void onCompleted(Run<?,?> r, TaskListener listener) {
+            if (r instanceof WorkflowRun) {
+                WorkflowJob j = ((WorkflowRun) r).getParent();
+                PreserveStashesJobProperty prop = j.getProperty(PreserveStashesJobProperty.class);
+                if (prop != null && prop.getBuildCount() > 0) {
+                    // TODO: Determine if this is overzealous. Might need to ape log rotation more closely.
+                    for (WorkflowRun build : j.getBuilds().completedOnly()) {
```
jglick: This will be super inefficient for jobs with a lot of builds. You need to use `RunLoadCounter` to verify that you are not destroying performance.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
```diff
@@ -0,0 +1,122 @@
...
+        public boolean shouldClearAll(@Nonnull Run<?,?> build) {
+            if (build instanceof WorkflowRun) {
+                WorkflowRun r = (WorkflowRun)build;
+                WorkflowJob j = r.getParent();
+                PreserveStashesJobProperty prop = j.getProperty(PreserveStashesJobProperty.class);
+                if (prop != null) {
+                    int bc = prop.getBuildCount();
+                    // Being particularly defensive just to be safe.
+                    if (bc > 0) {
+                        for (WorkflowRun recentRun : j.getBuilds().completedOnly().limit(bc)) {
```
jglick: Test with `RunLoadCounter`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
```diff
@@ -0,0 +1,122 @@
...
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class PreserveStashesJobProperty extends OptionalJobProperty<WorkflowJob> {
+
+    private static final Logger LOGGER = Logger.getLogger(PreserveStashesJobProperty.class.getName());
+
+    private int buildCount = 10;
```
jglick: Strange default. 1 would make some sense, perhaps. Consider the normal case that you want to permit resumption of only the most recent build, and tune the default accordingly.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/causes/Messages.properties
```diff
@@ -0,0 +1,25 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+RestartedDeclarativePipelineCause.ShortDescription=Restarted from build #{0}, stage {1}
```
jglick: Create a real `description` view so that you can use model hyperlinks.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty/config.jelly
```diff
@@ -0,0 +1,31 @@
...
+  ~ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="buildCount" title="Keep this many of the most recent builds' stashes">
+        <f:number/>
```
jglick: use CSS classes to force this to be nonnegative or whatever

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartFlowFactoryAction.java
abayer: Actually, I just need to get rid of the `TODO`. =) `CopyStashesAndArtifacts` should just work on its own, and I opted to not copy the environment variables, since we're re-evaluating the `environment` sections.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: Will do.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: Will do.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: Sounds reasonable.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/causes/Messages.properties
abayer: Okiedokie.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty/config.jelly
abayer: Yeah, I literally just hacked this in yesterday to fix something. Need to flesh it out more.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
ljohnston: Sorry, I don't really have the context of the code or the details of this discussion, but from this related issue, https://issues.jenkins-ci.org/browse/JENKINS-45455...

*Use cases*
* Restarting a successful stage so to rollback or redeploy code

Sounds like this PR discussion is about limiting the restart to the most recent build. Seems like that's at odds with the above use case?

abayer: @ljohnston You'd want to specify a higher number to keep in that case - we're just talking about setting the default value, not the max.

ljohnston: Ah, got it... thx!

comscience15:
@abayer I have a question. Is it correct that this PR will restart a specific stage on the next build, not from the same build?

My understand from PR topic is seems to be Build A is done and Build B will restart at the specific stage. If it's true, can we also get the restart stage in the same build? 

For example,
1) Executed Stage 1
2) Executed Stage 2
3) Re-executed Stage 2
4) Executed Stage 3 ....... and so on

abayer:
Once a build has completed, it's basically immutable. So what this does is kick off a new build, starting from the stage you're restarting.

comscience15:
@abayer Thanks for clarifying.  Do you or someone have a plan in the future for my scenario too?

preet1309:
Hey... Really exited and eagerly waiting for this feature to be out there.
Can we get a rough timeline on this

Thanks in advance

abayer:
We should have a fairly concrete timeframe for this and #227 within a few weeks - my *rough* guess is sometime in June. Sorry for the delays - getting all the pieces in line has been a hassle. =)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: Blech. Finally got around to looking into this and realized `RunLoadCounter` is `AbstractProject`-only. Need to think about how to rework it to be Pipeline-friendly.

abayer: and got it.

abayer:
@jglick - see `PipelineRunLoadCounter` and `PreserveStashesJobPropertyLoadTest` - I *think* I got the right testing in there.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -0,0 +1,207 @@
...
+        JSONObject form = req.getSubmittedForm();
+        String stageName = Util.fixEmpty(form.getString("stageName"));
+
+        try {
+            run(stageName);
+        } catch (IllegalStateException ise) {
+            throw HttpResponses.error(SC_CONFLICT, ise);
+        }
+
+        rsp.sendRedirect("../.."); // back to WorkflowJob; new build might not start instantly so cannot redirect to it
```
jglick: Make a mention of this case in [JENKINS-49138](https://issues.jenkins-ci.org/browse/JENKINS-49138).

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -0,0 +1,207 @@
...
+        if (!isRestartEnabled()) {
+            throw new AccessDeniedException("not allowed to restart"); // AccessDeniedException2 requires us to look up the specific Permission
+        }
+        JSONObject form = req.getSubmittedForm();
+        String stageName = Util.fixEmpty(form.getString("stageName"));
+
+        try {
+            run(stageName);
+        } catch (IllegalStateException ise) {
+            throw HttpResponses.error(SC_CONFLICT, ise);
```
jglick: Have `run` throw `Failure`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartFlowFactoryAction.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public boolean shouldSchedule(List<Action> actions) {
+        return true;
+    }
+
+    @Override
+    public CpsFlowExecution create(FlowDefinition def, FlowExecutionOwner owner, List<? extends Action> actions) throws IOException {
+        Run original = Run.fromExternalizableId(originRunId);
+        String origScript = null;
+        boolean origSandbox = true;
```
jglick: Is non-sandboxed mode even an option for Declarative? I am not totally sure it is safe to support this. Probably better to just always run in the sandbox unless you have thought this through completely.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartFlowFactoryAction.java
jglick: Status?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
jglick: Status?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
jglick: So what was the result? I feel like the test may be wrong, since I do not see how this could _not_ load too much.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -331,7 +314,7 @@ class ModelInterpreter implements Serializable {
     }
 
     def stageInput(StageInput input, Closure body) {
-        if (input != null)  {
+        if (input != null) {
```
jglick: can revert

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction/index.jelly
```diff
@@ -0,0 +1,48 @@
...
+        <l:main-panel>
+            <j:if test="${it.restartEnabled}">
+                <h1>Restart ${it.owner.displayName} from Stage</h1>
+                <p>
+                    <j:out value="Help text to go here"/>
+                </p>
+                <f:form action="restart" method="POST" name="restart">
+                    <f:entry field="stageName" title="Stage Name">
+                        <!-- TODO: Help! -->
+                        <f:textarea checkUrl="'${it.checkUrl}?value='+encodeURIComponent(this.value)"/>
```
jglick: Do you not just want an `f:select` here? Or even a list of stages, each one with a **Restart from here** button?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/WhenStageTest.java
```diff
@@ -191,19 +191,72 @@ public void whenChangesetMoreCommits() throws Exception {
...
                 .logNotContains(
-                        "Stage 'Two' skipped due to when conditional",
+                        "Stage \"Two\" skipped due to when conditional",
                         "Warning, empty changelog.",
                         "Examining changelog from all builds of this change request.")
                 .go();
     }
 
+    @Test
+    public void whenChangesetPR() throws Exception {
```
jglick: Related?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
```diff
@@ -0,0 +1,709 @@
...
+
+    @Issue("JENKINS-45455")
+    @Test
+    public void restartDisabled() throws Exception {
+        SecurityRealm originalRealm = j.jenkins.getSecurityRealm();
+        AuthorizationStrategy originalStrategy = j.jenkins.getAuthorizationStrategy();
+
+        try {
+            j.jenkins.setSecurityRealm(j.createDummySecurityRealm());
+            GlobalMatrixAuthorizationStrategy gmas = new GlobalMatrixAuthorizationStrategy();
```
jglick: use `MockAuthorizationStrategy`

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
```diff
@@ -0,0 +1,709 @@
...
+            assertTrue(action.isRestartEnabled());
+            assertTrue(canRestart(b, "admin"));
+            assertTrue(canRestart(b, "dev1"));
+            assertFalse(canRestart(b, "dev2"));
+
+            p.setDisabled(true);
+            assertFalse(canRestart(b, "admin1"));
+            assertFalse(canRestart(b, "dev1"));
+            assertFalse(canRestart(b, "dev2"));
+        } finally {
```
jglick: Are you using a `ClassRule` or something? With normal `@Rule JenkinsRule` there is no need to reset anything.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
```diff
@@ -0,0 +1,709 @@
...
+            j.jenkins.setSecurityRealm(originalRealm);
+            j.jenkins.setAuthorizationStrategy(originalStrategy);
+        }
+    }
+
+    private static boolean canRestart(WorkflowRun b, String user) {
+        final RestartDeclarativePipelineAction a = b.getAction(RestartDeclarativePipelineAction.class);
+        return ACL.impersonate(User.get(user).impersonate(), new NotReallyRoleSensitiveCallable<Boolean,RuntimeException>() {
```
jglick: `ACL.as`

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
```diff
@@ -0,0 +1,709 @@
...
+
+        RestartDeclarativePipelineAction action = original.getAction(RestartDeclarativePipelineAction.class);
+        assertNotNull(action);
+        assertTrue(action.isRestartEnabled());
+
+        HtmlPage page = j.createWebClient().getPage(original, action.getUrlName());
+        HtmlForm form = page.getFormByName("restart");
+        HtmlTextArea text = form.getTextAreaByName("_.stageName");
+        text.setText("not-present");
+        Thread.sleep(500);
```
jglick: !

pipeline-model-definition/src/test/resources/restart/sameCheckoutMultibranch.groovy
```diff
@@ -0,0 +1,56 @@
...
+                    } else {
+                        echo "Even numbered build, success"
+                    }
+                }
+            }
+        }
+        stage('post-restart') {
+            when {
+                expression {
+                    !fileExists("newFile")
```
jglick: Confused by this. We expect this always to be true, right?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PipelineRunLoadCounter.java
```diff
@@ -0,0 +1,190 @@
...
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@SuppressWarnings({"unchecked", "rawtypes"}) // API design mistakes
+public final class PipelineRunLoadCounter {
```
jglick: Better to just fix the `RunLoadCounter` in `jenkins-test-harness` to work on any `Run`. Rather than `MarkerAdder extends Notifier`, I think you could simply have a `RunListener.onCompleted` which calls `Marker.add` for newly created builds.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
```diff
@@ -0,0 +1,113 @@
...
+    @Rule
+    public RestartableJenkinsRule story = new RestartableJenkinsRule();
+
+    @Rule
+    public LoggerRule logger = new LoggerRule();
+
+    @Issue("JENKINS-45455")
+    @Test
+    public void stashWithLoadCount() throws Exception {
+        story.addStep(new Statement() {
```
jglick: `.then` + lambda

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
```diff
@@ -0,0 +1,113 @@
...
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "stashWithLoadCount");
+                p.setDefinition(new CpsFlowDefinition(pipelineSourceFromResources("properties/stashWithLoadCount"), true));
+
+                // Build 10 times to make sure get 10 total builds.
+                for (int i = 0; i < 10; i++) {
+                    story.j.waitForCompletion(p.scheduleBuild2(0).waitForStart());
```
jglick: `r.buildAndAssertSuccess(p)`

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
```diff
@@ -0,0 +1,113 @@
...
+                PipelineRunLoadCounter.prepare(p);
+
+                // PreserveStashesJobProperty.StashClearingListener and .SaveStashes shouldn't trigger a full load.
+                assertEquals(0, PipelineRunLoadCounter.countLoads(p, new Runnable() {
+                    @Override
+                    public void run() {
+                        try {
+                            story.j.buildAndAssertSuccess(p);
+                        } catch (Exception e) {
+                            // This will never happen.
```
jglick: orly? how about an `assert`, just in case

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = (WorkflowJob)story.j.jenkins.getItem("stashWithLoadCount");
+                assertNotNull(p);
+                PipelineRunLoadCounter.prepare(p);
+
+                // PreserveStashesJobProperty.StashClearingListener and .SaveStashes shouldn't trigger a full load.
+                assertEquals(0, PipelineRunLoadCounter.countLoads(p, new Runnable() {
```
jglick: can use lambdas

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyTest.java
```diff
@@ -0,0 +1,112 @@
...
+        assertFalse(StashManager.stashesOf(p.getBuildByNumber(5)).isEmpty());
+    }
+
+    @Issue("JENKINS-45455")
+    @Test
+    public void stashWithNegativePropertyValue() throws Exception {
+        WorkflowRun r = expect("properties", "stashWithNegativePropertyValue")
+                .go();
+        // We shouldn't be keeping any stashes, since the build count isn't greater than 0.
+        assertTrue(StashManager.stashesOf(r).isEmpty());
```
jglick: I would just expect an `IllegalArgumentException` if a negative value were passed…

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: Yeah, I had the same suspicion. Which is why I put the `lastFailed` check in the test too - so it looks, at the very least, like the `completedOnly()` check doesn't actually lead to loading actions. I'm still wary.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: Still considering.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -0,0 +1,207 @@
...
+        @Override
+        public Class<WorkflowRun> type() {
+            return WorkflowRun.class;
+        }
+
+        @Override
+        @Nonnull
+        public Collection<? extends Action> createFor(@Nonnull WorkflowRun run) {
+            // TODO: Can't actually check if any other actions are present, on the run or on its parent, so I can't verify
+            // this is a Declarative run here. If I try, I end up in an infinite loop. Need to think on this.
```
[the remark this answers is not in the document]
abayer: Rightieo.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -0,0 +1,207 @@
...
+        actions.add(new CauseAction(new Cause.UserIdCause(), new RestartDeclarativePipelineCause(run, stageName)));
+        return ParameterizedJobMixIn.scheduleBuild2(run.getParent(), 0, actions.toArray(new Action[actions.size()]));
+    }
+
+    @Extension
+    public static class Factory extends TransientActionFactory<WorkflowRun> {
+
+        @Override
+        public Class<WorkflowRun> type() {
+            return WorkflowRun.class;
```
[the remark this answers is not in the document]
abayer: Given how many other Declarative things are specific to `WorkflowRun`, I don't see a problem in keeping the same requirements here.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
abayer: `Failure`?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartFlowFactoryAction.java
abayer: It's theoretically possible to be outside the sandbox - i.e., a Declarative Pipeline defined in a `CpsFlowDefinition` directly.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Nah, spacing fixes aren't bad things.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction/index.jelly
abayer: I'm basically tabling the UI aspects until I huddle with someone who actually knows UI. =)

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/WhenStageTest.java
abayer: Merge/rebase fun. Ignore for now, won't be in the diff for the real PR once #227 has merged.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
abayer: Yup, I'm using `@ClassRule`.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
abayer: hrm?

pipeline-model-definition/src/test/resources/restart/sameCheckoutMultibranch.groovy
abayer: Ah, yeah, lurking bit from copy-paste.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PipelineRunLoadCounter.java
abayer: Ok. I'll prototype it here first to make sure it does the trick and then open a PR over in `jenkins-test-harness`.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
abayer: Nope, deliberately not checking result - I have build 7 failing so that I can do the `lastFailing` load check later.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
abayer: Sure thing.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
abayer: Yeah, I should get around to switching to lambdas. =)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
jglick: [Javadoc](http://javadoc.jenkins.io/hudson/model/Failure.html)

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
jglick: Generally a bad code smell.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
jglick: `buildAndAssertStatus` then?

jglick: Never mind, no such method.

jglick: Though you _could_ use

```java
story.j.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0));
```

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
abayer: Danke.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyLoadTest.java
abayer: Then I'd have to special-case it for runs 1-6,8-10 vs run 7. I really don't care about the build result overall - I just have that one failure to verify load count, like I said.

jglick:
I guess you rebased—I can no longer look at changes since my last review.

abayer:
Yup, I did. https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/228/commits/c280fd21fee8058929183b981b89e1b3476b4d74 is probably what you're looking for.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -121,7 +118,7 @@ public void doRestart(StaplerRequest req, StaplerResponse rsp) throws ServletExc
         try {
             run(stageName);
         } catch (IllegalStateException ise) {
-            throw HttpResponses.error(SC_CONFLICT, ise);
+            throw new Failure("Failure restarting from stage: " + ise);
```
jglick: Why not just make `run` itself throw `Failure`?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -75,7 +77,7 @@ private RestartDeclarativePipelineAction(Run run) {
 
     @Override public String getIconFileName() {
         // TODO: Needs an icon.
```
jglick: obsolete comment

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
jglick: …and remove the `matrix-auth` test dep

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
jglick: I am fairly confident this is still wrong, and if your test is not catching it, you got the test wrong somehow.

Worse, for `artifact-manager-s3`, by default this call will do nothing, so loading build records to do nothing is a pure waste.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PipelineRunLoadCounter.java
jglick: …

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PipelineRunLoadCounter.java
```diff
@@ -0,0 +1,190 @@
...
+                assert currCount.get().incrementAndGet() != 0 : "too many build records loaded from " + project;
+            }
+        }
+
+        @Override public void onAttached(Run r) {}
+
+        @Override public void onBuildComplete() {}
+
+        @Override public String getIconFileName() {
+            return null;
```
jglick: just extend `InvisibleAction`

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PipelineRunLoadCounter.java
```diff
@@ -0,0 +1,190 @@
...
+            return new MarkerAdderStepExecution(context);
+        }
+
+
+        @Extension
+        public static final class MarkerAdderStepDescriptorImpl extends StepDescriptor {
+
+            @Override
+            public String getFunctionName() {
+                return "markerAdder";
```
jglick: See previous comment, this is too much work; you just need a `RunListener`.

jglick: try https://github.com/jenkinsci/jenkins-test-harness/pull/106

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: fwiw, in debug testing, the test only calls `WorkflowRun#onLoad` once total, for build 10, the most recent build as of the restart.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
abayer: No, we're still waiting on the real one - this is just a placeholder.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobPropertyTest.java
abayer: I prefer to be safer here - I'd rather keep builds from barfing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: Yeah, 1 makes sense for the default. Updating.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction/index.jelly
```diff
@@ -31,7 +31,8 @@
             <j:if test="${it.restartEnabled}">
                 <h1>Restart ${it.owner.displayName} from Stage</h1>
                 <p>
-                    <j:out value="Help text to go here"/>
+                    <j:out value="${%A completed Declarative Pipeline run can be restarted, starting at the beginning of any top-level stage which ran in the original
+    build.}"/>
```
jglick: Weird place to line-wrap? Anyway long text like this should go into an `index.properties`.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/causes/RestartDeclarativePipelineCause/description.jelly
```diff
@@ -0,0 +1,31 @@
...
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core">
+    <j:set var="b" value="${it.originalBuild}"/>
+    <j:if test="${b != null}">
+        <span>${%restarted_from_build(it.originStage, rootURL, b.url, b.displayName)}</span>
```
jglick: Possible stored-XSS in the `displayName`. (Messages are definitely exempted from automatic escaping. I cannot recall if that applies to message parameters as well.) I think you need to wrap in `h.escape`.

pom.xml
```diff
@@ -197,6 +197,11 @@
         <artifactId>jenkins-test-harness-tools</artifactId>
         <version>2.2</version>
       </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.main</groupId>
+        <artifactId>jenkins-test-harness</artifactId>
+        <version>2.39</version>
```
jglick: Rather override `jenkins-test-harness.version`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
jglick: > You mean `StashManager.maybeClearAll(build)` won't do anything for `artifact-manager-s3`?

By default, yes.

> something is definitely different between those cases

Sounds like something is wrong with the test. Debug it.

> I don't see how this'd be any worse than, say, `LogRotator`

Because `LogRotator` _deletes_ builds that are too old, so the next time it runs, it does not need to iterate them.

> `LogRotator` is looking at more data for each run than this is, since it's looking at `r.getTimestamp()` and `r.isKeepLog()`

Mostly irrelevant. The expensive part is loading the `Run` object from `build.xml`.

> the stash clearing stuff literally just looks for completed builds

Of which there could be a million. Every one of which you will iterate, forcing it to be loaded into heap, every time a build completes.

`SaveStashes#shouldClearAll` looks OK, because it uses a `limit`.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction/index.jelly
abayer: Good point.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/causes/RestartDeclarativePipelineCause/description.jelly
abayer: Okiedokie.

pom.xml
abayer: I forgot that was a thing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: I did debug it - as I said, `onLoad` is only called once. 

But ok, I'll throw a limit on the maximum number of builds-with-stashes to keep, and then set a limit in `StashClearingListener#onCompleted` to go with it.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/cli/RestartFromStageCommand.java
```diff
@@ -0,0 +1,87 @@
...
+    public int number;
+
+    @Override
+    public String getShortDescription() {
+        return Messages.RestartFromStageCommand_ShortDescription();
+    }
+
+    @Override
+    protected int run() throws Exception {
+        Run<?,?> run = number == 0 ? job.getLastBuild() : job.getBuildByNumber(number);
```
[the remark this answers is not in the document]
abayer: Seems reasonable.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -0,0 +1,210 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
```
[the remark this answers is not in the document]
abayer: ...and this file predates this year. =)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -0,0 +1,210 @@
...
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import javax.servlet.ServletException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+public class RestartDeclarativePipelineAction implements Action {
```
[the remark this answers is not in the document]
abayer: Not sure if we want that - it feels like a bunch of noise given that every single Declarative run will have this.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartFlowFactoryAction.java
```diff
@@ -0,0 +1,92 @@
...
+        if (original instanceof FlowExecutionOwner.Executable) {
+            FlowExecutionOwner originalOwner = ((FlowExecutionOwner.Executable) original).asFlowExecutionOwner();
```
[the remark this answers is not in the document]
abayer: Reasonable.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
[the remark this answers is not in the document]
rsandell: badge.jelly could potentially only render something if it is `lastBuildNumber - 5 && result != SUCCESS` ?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/Messages.properties
```diff
@@ -0,0 +1,26 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+#
+
+PreserveStashesJobProperty.ValidatorImpl.InvalidBuildCount=The number of builds to preserve must be between 1 and 50.
```
rsandell: That constant again.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty/config-details.jelly
```diff
@@ -0,0 +1,31 @@
...
+  ~ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="buildCount" title="Keep this many of the most recent builds' stashes">
+        <f:number value="1"/>
```
rsandell: `value="1"`? you mean `default`?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty/help-buildCount.html
```diff
@@ -0,0 +1,28 @@
...
+  ~ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  ~ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  ~ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+
+<div>
+    The number of most recent builds whose <code>stash</code>es should be preserved. Must be between 1 and 50.
```
rsandell: The constant again

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommandTest.java
```diff
@@ -29,6 +29,7 @@
 import hudson.model.User;
 import jenkins.model.Jenkins;
 import org.apache.commons.io.FileUtils;
+import org.jenkinsci.plugins.pipeline.modeldefinition.cli.DeclarativeLinterCommand;
```
rsandell: Maybe move the actual test into the new package as well.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
```diff
@@ -0,0 +1,649 @@
...
+        RestartDeclarativePipelineAction action = original.getAction(RestartDeclarativePipelineAction.class);
+        assertNotNull(action);
+        assertTrue(action.isRestartEnabled());
+
+        HtmlPage page = j.createWebClient().getPage(original, action.getUrlName());
+        HtmlForm form = page.getFormByName("restart");
+        HtmlSelect select = form.getSelectByName("stageName");
+        select.getOptionByValue(stageName).setSelected(true);
+        return j.submit(form);
+    }
```
rsandell: restart with new Jenkinsfile where stage names has changed?
See my thoughts before.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
```diff
@@ -0,0 +1,167 @@
...
+import org.kohsuke.stapler.DataBoundSetter;
+import org.kohsuke.stapler.QueryParameter;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class PreserveStashesJobProperty extends OptionalJobProperty<WorkflowJob> {
```
rsandell: Might be good to have this as a folder property as well?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/SkippedStageReason.java
```diff
@@ -0,0 +1,172 @@
...
+ */
+
+package org.jenkinsci.plugins.pipeline.modeldefinition.model;
+
+import org.jenkinsci.plugins.pipeline.StageStatus;
+
+import javax.annotation.Nonnull;
+import java.io.Serializable;
+
+public abstract class SkippedStageReason implements Serializable {
```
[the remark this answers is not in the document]
abayer: Crap, now I have to remember exactly what the point was. =)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
```diff
@@ -0,0 +1,167 @@
...
+        @Override
+        @Nonnull
+        public String getDisplayName() {
+            return "Preserve stashes from completed builds";
+        }
+
+        public FormValidation doCheckBuildCount(@QueryParameter int value) {
+            if (value < 0) {
+                return FormValidation.error("Must be greater than or equal to 0");
+            } else if (value > 50) {
```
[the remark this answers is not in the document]
abayer: d'oh. Fixing all of those.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Extension
+    public static final class StashClearingListener extends RunListener<Run<?,?>> {
+        @Override
+        public void onCompleted(Run<?,?> r, TaskListener listener) {
+            if (r instanceof WorkflowRun) {
+                WorkflowJob j = ((WorkflowRun) r).getParent();
+                PreserveStashesJobProperty prop = j.getProperty(PreserveStashesJobProperty.class);
+                if (prop != null && prop.getBuildCount() > 0) {
+                    // The "+1" is to ensure that we look at one more than the maximum possible number of builds with saved stashes.
```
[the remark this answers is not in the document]
abayer: Hmm...maybe? They'd all have to complete at basically the exact same time for that to even theoretically happen. Yeah, I don't think it can plausibly happen.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -135,13 +138,31 @@ class ModelInterpreter implements Serializable {
...
      */
-    def evaluateSequentialStages(Root root, Stages stages, Throwable firstError, Stage parent) {
+    def evaluateSequentialStages(Root root, Stages stages, Throwable firstError, Stage parent, String restartedStageName,
+                                 SkippedStageReason skippedReason) {
         return {
+            boolean skippedForRestart = restartedStageName != null
             stages.stages.each { thisStage ->
+                if (skippedForRestart) {
+                    // Check if we're skipping for restart but are now on the stage we're supposed to restart on.
+                    if (thisStage.name == restartedStageName) {
```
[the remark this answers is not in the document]
abayer: Restart will always use the original Jenkinsfile/shared libraries/etc, so that's not an actual possibility.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty/config-details.jelly
abayer: durrrr, yeah.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty/help-buildCount.html
abayer: Kinda stuck with it here since the help html is a static file...

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativeLinterCommandTest.java
abayer: Good call.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: Interesting. Lemme think on that for a little.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineActionTest.java
abayer: As mentioned above, that isn't actually possible, so we're good. =)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
abayer: At least for the moment, I'm gonna say let's not have a folder property equivalent. It adds a fair amount of complexity, and I'm honestly not sure if it's functionality worth having. We can always revisit it post-release.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty/help-buildCount.html
rsandell: You can do a `help.jelly` or a `help.groovy` instead ;)

abayer: Oh riiiiight

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty.java
```diff
@@ -112,8 +126,8 @@ public void onCompleted(Run<?,?> r, TaskListener listener) {
                 WorkflowJob j = ((WorkflowRun) r).getParent();
                 PreserveStashesJobProperty prop = j.getProperty(PreserveStashesJobProperty.class);
                 if (prop != null && prop.getBuildCount() > 0) {
-                    // TODO: Determine if this is overzealous. Might need to ape log rotation more closely.
-                    for (WorkflowRun build : j.getBuilds().completedOnly()) {
+                    // The "+1" is to ensure that we look at one more than the maximum possible number of builds with saved stashes.
+                    for (WorkflowRun build : j.getBuilds().completedOnly().limit(MAX_SAVED_STASHES + 1)) {
```
jglick: I guess you never managed to fix the test to fail if you omit `limit`?

Will probably neglect to clear stashes under some conditions, such as if you start off with a high limit and then edit `properties` to specify a lower limit.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/properties/PreserveStashesJobProperty/help-buildCount.jelly
```diff
@@ -22,6 +23,11 @@
   ~ THE SOFTWARE.
   -->
 
-<div>
-    The number of most recent builds whose <code>stash</code>es should be preserved. Defaults to 1 if unspecified.
-</div>
\ No newline at end of file
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <j:getStatic var="maxCount" className="org.jenkinsci.plugins.pipeline.modeldefinition.properties.PreserveStashesJobProperty" field="MAX_SAVED_STASHES" />
```
jglick: BTW it is generally easier to handle such things as simple instance properties (getters) on the `Descriptor`:

```xml
between 1 and ${descriptor.maxCount}
```

PR #230, opened 2017-12-18
[FIXED JENKINS-48380] Add per-stage "options"

abayer:
* JENKINS issue(s):
    * [JENKINS-48380](https://issues.jenkins-ci.org/browse/JENKINS-48380)
* Description:
    * On top of #227 again, so...yeah. Look for commits starting with https://github.com/jenkinsci/pipeline-model-definition-plugin/commit/2691632f2e111a94c615ce1ea999ee20162fa5d9 - this'll get cleaned up once #227 is merged.
    * More specifically, allow block-scoped steps that don't require an agent, and DeclarativeOptions that explicitly state they're eligible to be used in a stage (skipDefaultCheckout being the only one at the moment).
* Documentation changes:
    * Will be needed.
* Users/aliases to notify:
    * @reviewbybees


i386:
This is really cool. Why is it on top of #227? Asking as that PR (at least in my mind) is longer term than this one.

abayer:
@i386 Mentioned offline, but the reason is that #227 makes some major changes to how stages are managed at runtime. It's easier to implement this on top of that change, first of all, and I wasn't looking forward to trying to merge things together as well. But I'll see if I can tweeze them apart.

abayer:
Replaced with #231.

PR #231, opened 2017-12-19
[FIXED JENKINS-48380, JENKINS-48379] Add per-stage "options" and input

abayer:
* JENKINS issue(s):
    * [JENKINS-48380](https://issues.jenkins-ci.org/browse/JENKINS-48380)
    * [JENKINS-48379](https://issues.jenkins-ci.org/browse/JENKINS-48379)
    * [JENKINS-44277](https://issues.jenkins-ci.org/browse/JENKINS-44277)
* Description:
    * Replaces #230 by not being on #227. =)
    * More specifically, allow block-scoped steps that don't require an agent, and DeclarativeOptions that explicitly state they're eligible to be used in a stage (skipDefaultCheckout being the only one at the moment).
    * Also adds `input` section to `stage`. Parameters from `input` are added to the stage's environment.
    * Order of stage internals execution, fwiw, is:
        * Skip if earlier failures etc first.
        * "wrappers" from `options` - i.e., `timeout` and friends.
        * `input`
        * `agent` (for stages without nested `parallel` or `stages`)
        * `when`
        * `environment`
    * So `input` parameters should be available (as environment variables, not `params.FOO`) for `when` evaluation...
    * Yup, more - added `checkoutToSubdirectory('foo')` option, since we're already messing around with `DeclarativeOption` guts in this PR, and I wanted to make sure I remembered to have `checkoutToSubdirectory('foo')` usable on stages.
    * TODO: Don't allow file parameters, since there's no guarantee we'll be in an agent.
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1288
* Users/aliases to notify:
    * @reviewbybees


abayer:
fwiw, this shouldn't be too hard to do in the editor as well (if we already have `options` support there) - the eligible content is slightly different, but not much.

abayer:
Added @bitwiseman because I thought he might be interested. =)

bitwiseman:
@abayer  The input part of this would obviate the need for a separate `agent none` stage to do input steps.  I like that.  
"input parameters should be available (as environment variables, not params.FOO)"  - This is confusing.  The `whatever { parameters { ... } }` should show up under `params`.  I would suggest adding a `parameter` condition to `when`, rather than shoehorning parameters into `environment`.
But, wait in `sh` steps `params.FOO` shows up a an environment variable anyway, right? 
I see what you're shooting for, but the naming is going to cause confusion.

abayer:
@bitwiseman Note that if you don't have `agent none` at the top level, the `input` is still gonna happen inside the top-level `agent`. Nothing I can do about that. =)

And speaking of "nothing I can do about that", I can't add to `params` once the build has started. That pulls directly from the `ParametersAction` on the run itself (and is over in `workflow-cps`, fwiw, not here), and since the `input` parameters aren't part of the overall build parameters, they're not in the `ParametersAction` and therefore not in `params`. There really is nothing I can do here - the best I can do is the `withEnv(["INPUT_PARAM=input_value"]) { ... }` that gets called behind the scenes to make sure that `env.INPUT_PARAM` and `INPUT_PARAM` available.

abayer:
@bitwiseman I should mention that I *tried* to find a way to do that - I could theoretically add the `input` parameters to the run's `ParametersAction`, but that could mess things up proper with parallel stages etc, and would generally be...wrong. So I'm not gonna. =)

michaelneale:
nice - this has some good stuff

bitwiseman:
@abayer - is `checkoutToSubdirectory` an option that can go on a stage?  If so, at what point does it run?  Could I put an `sh` step inside the parens and use the return value as my argument?

abayer:
Yes, it can go on a stage, but it needs to take a string/gstring.

bitwiseman:
@abayer - my question was when does it execute? Obviously before checkout, but also before agent/workspace assignment?  After input or before?  

I'm loving all these new cool behaviors but I'm getting a bit nervous about potential order-of-execution interactions between these directives.

abayer:
@bitwiseman I’m not 100% certain - I think the value would resolve at the beginning of the build, but tests will be needed.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
```diff
@@ -76,16 +78,29 @@ class Agent extends MappedClosure<Object,Agent> implements Serializable {
...
-                SkipDefaultCheckout skip = (SkipDefaultCheckout) root?.options?.options?.get("skipDefaultCheckout")
-                if (!skip?.isSkipDefaultCheckout()) {
-                    doCheckout = true
+            if (!options.isEmpty()) {
+                SkipDefaultCheckout skip = (SkipDefaultCheckout) options.get("skipDefaultCheckout")
+                if (skip?.isSkipDefaultCheckout()) {
+                    doCheckout = false
+                }
+
+                CheckoutToSubdirectory subdir = (CheckoutToSubdirectory) options.get("checkoutToSubdirectory")
```
rsandell: Is there validation in place to make sure someone doesn't do `skipDefaultCheckout` and `checkoutToSubdirectory` at the same time?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/StageInput.groovy
```diff
@@ -0,0 +1,89 @@
...
+import javax.annotation.Nonnull
+
+/**
+ * A container for lists of parameter definitions.
+ *
+ * @author Andrew Bayer
+ */
+@ToString
+@EqualsAndHashCode
+@SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
```
rsandell: why not just put one in?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -180,49 +181,55 @@ class ModelInterpreter implements Serializable {
...
-                        if (thisStage.parallel != null) {
-                            script.parallel(getParallelStages(root, parentAgent, thisStage, firstError, parentStage, false, true, false))
-                        }
-                    } else {
-                        if (thisStage.parallel != null) {
-                            if (evaluateWhen(thisStage.when)) {
-                                withCredentialsBlock(thisStage.environment) {
-                                    withEnvBlock(thisStage.getEnvVars(script)) {
-                                        script.parallel(getParallelStages(root, parentAgent, thisStage, firstError, parentStage, false, false, false))
+                    inWrappers(thisStage.options?.wrappers) {
```
rsandell: Would be nice if we could have a two stage wrappers. One here and one inside the agent (for those wrappers requiring a workspace) Then people could use file credentials etc.

rsandell: And should we really not check for `firstError` etc. before entering the wrappers? they could produce quite a lot of computation for no use since the stage will be "skipped" anyway.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -432,16 +477,21 @@ class ModelInterpreter implements Serializable {
...
     /**
      * Executes the given closure inside 0 or more wrapper blocks if appropriate
-     * @param options The options configuration we're executing in
+     * @param wrappers A map of wrapper names to wrappers
      * @param body The closure to execute
      * @return The return of the resulting executed closure
      */
-    def inWrappers(Options options, Closure body) {
-        if (options?.wrappers != null) {
+    def inWrappers(Map<String,Object> wrappers, Closure body) {
```
rsandell: I don't see the reason for this change. You've just replaced the calls with `something.options.wrappers` instead.

rsandell:
could you add a test for using `withCredentials` in stage options? could be good to have it verified it can be used as a workaround for credential types supported by `withCredentials` but not (yet) by 'credentials' in 'environment'. :)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
abayer: No - not sure I want to add that level of validation, honestly. In practice, that'd just make the `checkoutToSubdirectory` meaningless, since the `skipDefaultCheckout` would override it.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/StageInput.groovy
abayer: Because Groovy ends up creating anonymous classes for closures that don't have 'em, so I just get in the habit of throwing the suppression on. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Lemme think on this.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: This may be an artifact of an earlier iteration where I wasn't supporting per-stage options...I'll revisit.

abayer:
Will add that!

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Oh, no, duh on me - it's because `root.options` is `Options.class` while `stage.options` is `StageOptions.class` and I couldn't figure out a good way to make them the same class, so hey. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: Yeah, agreed on the error case - tweaking.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
rsandell: There is this thing in Java called `interface` ;) although usually discouraged in Jenkins plugin development due to it being hard to handle binary compatibility going forward. But iirc we've used that quite extensively in this plugin :)

PR #232, opened 2017-12-20
[FIXED JENKINS-44461] Add beforeAgent flag for when

abayer:
* JENKINS issue(s):
    * [JENKINS-44461](https://issues.jenkins-ci.org/browse/JENKINS-44461)
* Description:
    * If true, the when condition will be evaluated before entering the stage's agent (if defined). The default is false - better safe than sorry there.
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1290
* Users/aliases to notify:
    * @reviewbybees


bitwiseman:
@abayer  Is the design final for this?  I think I'd rather see the revers of this as the default.  I know that would be a breaking change but I think we'd rather have _non_-allocation of an agent be the default whenever possible.   The name could be `requiresAgent`.

abayer:
Design is definitely not complete. I just thought of this today, after all. =)

I’d say one flag to control before both input and agent, nothing more granular. The code gets too complex at that point for marginal gain.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -209,22 +209,38 @@ class ModelInterpreter implements Serializable {
...
-                                } else {
-                                    Utils.logToTaskListener("Stage '${thisStage.name}' skipped due to when conditional")
-                                    Utils.markStageSkippedForConditional(thisStage.name)
-                                    isSkipped = true
                                 }
                             }
+                            // If beforeAgent is true, evaluate the when before entering the agent.
+                            boolean whenPassed = false
+                            if (thisStage.when?.beforeAgent != null && thisStage.when?.beforeAgent) {
+                                whenPassed = evaluateWhen(thisStage.when)
```
rsandell: `skipDueToParent`?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -209,22 +209,38 @@ class ModelInterpreter implements Serializable {
...
+                                if (whenPassed) {
+                                    inDeclarativeAgent(thisStage, root, thisStage.agent) {
+                                        stageBody.call()
+                                    }
+                                }
+                            } else {
+                                inDeclarativeAgent(thisStage, root, thisStage.agent) {
```
rsandell: Don't remember the exact flow here, but where is the "don't allocate agent if previous stage failed" check?

abayer: At the top of the method.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -209,22 +209,38 @@ class ModelInterpreter implements Serializable {
...
                                     }
-                                } else {
-                                    Utils.logToTaskListener("Stage '${thisStage.name}' skipped due to when conditional")
-                                    Utils.markStageSkippedForConditional(thisStage.name)
-                                    isSkipped = true
                                 }
                             }
+                            // If beforeAgent is true, evaluate the when before entering the agent.
+                            boolean whenPassed = false
+                            if (thisStage.when?.beforeAgent != null && thisStage.when?.beforeAgent) {
```
rsandell: I think this can be shortened to `thisStage.when?.beforeAgent` as groovy evaluates `null` as `false`.

PR #233, opened 2018-01-08
Add validation hook for ModelASTValue

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * No-op currently, but this'll allow a DeclarativeValidatorContributor to do more specific validation of a ModelASTValue in the future.

* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
@@ -55,6 +55,9 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


steven-foster:
Multiple errors will be useful for me, thanks!

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
```diff
@@ -74,121 +78,376 @@ public final String validateElement(@Nonnull ModelASTElement element, @CheckForN
         return null;
     }
     
+    @Nonnull
+    public final List<String> validateElementAll(@Nonnull ModelASTElement element, @CheckForNull FlowExecution execution) {
+        String r = validateElement(element, execution);
+        List<String> result = new ArrayList<>();
```
rsandell: No need to allocate a full new `ArrayList`:

`return r != null ? Collections.singeltonList(r) : Collections.emptyList()`
or
`return r != null ? Arrays.asList(r) : Collections.emptyList()`

PR #234, opened 2018-01-09
Start testing on a newer core as well.

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * Let's start running PR builds against a newer core version as well as the one we depend on.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...

PR #235, opened 2018-01-12
Revert "Start testing on a newer core as well."

abayer:
Reverts jenkinsci/pipeline-model-definition-plugin#234

Causes too much weirdness with test hangs. Reverting.


abayer:
Didn't actually help. yay.

PR #237, opened 2018-01-18
Add an extension point for runtime transformation of steps.

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * This'll end up getting used in `jx-pipelines-plugin` to avoid the same kind of closure->model transformation we got rid of when we switched to `RuntimeASTTransformer` here.
    * Also unrestricted `Utils#createStepsBlock(Closure)` because that'll probably end up used in implementations of this.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @rsandell


abayer:
@rsandell fyi, I may do a lightning-fast 1.2.7 release once this is approved and merged, but I can probably wait for your when tag stuff. =)

PR #240, opened 2018-01-23
 [JENKINS-48523] when conditions for tags and change requests

rsandell:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-48523 
* Description:
    * Two new when conditions that makes it easier to check the `TAG_NAME` and `CHANGE_*` environment variables.
    * Getting those environment vars for use in testing requires at least branch-api 2.0.11, which got me down the rabbit hell hole of upper dependencies requirements, where in the end I had to pumb as many versions up as possible without having things blow up.
* Documentation changes:
    * `help.html` is written for all, not sure if that is generated yet or not.
   * jenkins-infra/jenkins.io/pull/1430
* Users/aliases to notify:
    * @reviewbybees 
    * @abayer


abayer:
Initial build blew up with weird node problems, but did have legit test failures - https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/PR-240/1/testReport/

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BuildingChangeConditional.java
```diff
@@ -0,0 +1,347 @@
...
+    /**
+     * CHANGE_URL regular expression matching.
+     *
+     * Ant style matching doesn't work on url strings due to it being meant for regular paths
+     * and all those extra characters conflicts.
+     *
+     * @return urlX
+     * @see ObjectMetadataAction#getObjectUrl()
+     */
+    public String getUrlX() {
```
abayer: I don't like the `X` suffix on regex fields - `urlPattern` instead perhaps?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BuildingChangeConditional.java
```diff
@@ -0,0 +1,347 @@
...
+    private void check(String pattern, Comparator comparator, Names varName, EnvVars vars) {
+        if (isNotEmpty(pattern)) {
+            if (!comparator.compare(pattern, varName.get(vars))) {
+                throw new AssertionError(varName.name() + "("+varName.get(vars)+")" + " does not match " + pattern);
+            }
+        }
+    }
+
+    @Extension
+    @Symbol("buildingChange")
```
abayer: I don't love the symbol, but can't think of a better one yet.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BuildingChangeConditional.java
rsandell: `pattern` feels more like ant to me, maybe just because I call it that in `branch` and `tag`.
But I needed something simple to distinguish the various comparator types.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BuildingChangeConditional.java
rsandell: I needed something that was a bit further away from `changeset` and `changelog` than just `change`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BuildingChangeConditional.java
rsandell: Another approach could be to have another parameter to specify the comparator to use for all parameters (defaults to ant) but not all things works with ant style matching; url for example.

abayer: `urlRegexp` maybe, then? `urlX` is definitely not gonna work - short/simple is good, but that's just confusing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BuildingChangeConditional.java
abayer: I definitely understand - I'm just saying we need more to explain what this means than just `(something)Change` or `change`. After all, any SCM-triggered build on `master` has a *change*, just not a *change request*.

rsandell: I was thinking of `cr` at first but feared it was too short. And then when we decided on `buildingTag` I though I would follow that naming scheme :) But I agree, it's not very beautiful.

pipeline-model-definition/src/test/resources/when/whenBuildingChange.groovy
```diff
@@ -0,0 +1,159 @@
...
+
+pipeline {
+    agent {
+        label "here"
+    }
+    stages {
+        stage("One") {
+            steps {
+                echo "Hello"
+                sh "env | sort"
```
abayer: So ideally, change any use of `sh` to be like https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/test/resources/envBackslashes.groovy#L42-L53 (using `bat` as well), or if you can't replicate this in `bat`, put `onAllowedOS(PossibleOS.LINUX, PossibleOS.MAC);` as the first line of the actual `@Test` method.

rsandell: ah, this is just a debug step that I forgot to remove :/

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BuildingChangeConditional.java
abayer: Yeah, this is gonna need more thought and discussion. @michaelneale, @bitwiseman - thoughts?

michaelneale: is changeRequest the generalisation of pull request for all SCMs?

rsandell: Yes, branch-api and scm-api generalize all of them as change request. Hence the environment variable names.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeRequestConditional.java
```diff
@@ -0,0 +1,347 @@
...
+     * CHANGE_AUTHOR regular expression matching.
+     * @return authorDisplayName
+     * @see ContributorMetadataAction#getContributor()
+     */
+    public String getAuthorX() {
+        return authorX;
+    }
+
+    @DataBoundSetter
+    public void setAuthorX(String authorX) {
```
abayer: I am not a fan of the `X` suffix here. What does it mean to the user? If we're going with different fields for pattern matches vs string matches (quite reasonably), why not `authorPattern`?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeRequestConditional.java
```diff
@@ -0,0 +1,347 @@
...
+    @Extension
+    @Symbol("changeRequest")
+    public static class DescriptorImpl extends DeclarativeStageConditionalDescriptor<ChangeRequestConditional> {
+        @Override
+        public Expression transformToRuntimeAST(@CheckForNull ModelASTWhenContent original) {
+            return ASTParserUtils.transformWhenContentToRuntimeAST(original);
+        }
+    }
+
+    enum Comparator { //TODO this should be in a utility class for all when conditions to use
```
abayer: So do so? =) No need to rewrite the other conditions to use it right away, but might as well start out with this in the place you want it to end up.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeRequestConditional.java
```diff
@@ -0,0 +1,347 @@
...
+            boolean compare(String pattern, String actual) {
+                //TODO validation for pattern compile
+                return actual.matches(pattern);
+            }
+        };
+
+        abstract boolean compare(String pattern, String actual);
+    }
+
+    enum Names { //TODO this should be in a utility class for all when conditions to use
```
abayer: Same as above. If it's gonna be in a utility class, put it in a utility class.

rsandell:
Hmm, the build seems to have failed because it couldn't read a test report.

abayer:
Yeah, that's https://issues.jenkins-ci.org/browse/INFRA-1516

rsandell:
It doesn't look like that issue. The JUnit plugin failed reading an xml file. The build succeeded just fine.

abayer:
Nope, I dug into this earlier - the git failures are blowing up in such a way that `TEST-AgentTest.xml` or whatever is malformed. Hooray chains of pain.

rsandell:
:man_facepalming:

rsandell:
config.jelly added and help refined. But I don't know how to verify it without the declarative snippet generator :)

abayer:
Good enough for now. =)

PR #241, opened 2018-01-25
[JENKINS-49135] Add support for passing Options var/method to options

abayer:
* JENKINS issue(s):
    * [JENKINS-49135](https://issues.jenkins-ci.org/browse/JENKINS-49135)
* Description:
    * This is very preliminary (and even the minimal approach still needs tests for `StageOptions` as well as `Options`) and may not actually land. The idea is that you'd be able to either declare a variable of type `Options` or have a method that returns an instance of `Options`, and use that variable or method call as the argument to `options`, like this:

```
def opts = new Options([buildDiscarder(logRotator(numToKeepStr:'1'))], [:], [:])

pipeline {
  ...
  options opts
  ...
}

// or

def genOpts() {
    return new Options([buildDiscarder(logRotator(numToKeepStr: '1'))], [:], [:])
}

pipeline {
  ...
  options getOpts()
  ...
}
```
* Description continued
    * This would *not* allow for validation of the `options` contents, it *would* require actually instantiating an `Options` instance, rather than defining a closure like you'd normally specify for `options`, and it would not work in the editor (i.e., translating to JSON would just result in an empty `options` directive). This is necessary because the transformation behind the scenes from `options { ... }` to an `Options` instance requires runtime AST transformation, which means we need to treat any `options` method call anywhere in the Jenkinsfile or a shared library specially. Which is not optimal, to say the least.
    * Note that this would require declaring the *whole* `options` directive - no combining with explicitly specified entries as laid out in the JIRA.
    * If this ever actually lands, it'll be a power-user feature anyway, so I'd be ok with this requirement. That said, this is a good reason why I'm unsure about this landing.
    * Oh, and if we decide to move forward for this, I'll add support for some other directives (i.e., `triggers`, `parameters`, probably `agent`, `tools`), but not *all* (i.e., definitely *not* `stage`, `stages`, `when`, `post`, `steps`...).
* Documentation changes:
    * Good question. This is the kind of power-user feature that I wouldn't want documented on jenkins.io...which of course is another suggestion that this probably shouldn't land.
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTOptions.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTOptions.java
@@ -4,6 +4,8 @@ import java.util.ArrayList;
 import java.util.List;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+import org.codehaus.groovy.ast.expr.MethodCallExpression;
+import org.codehaus.groovy.ast.expr.VariableExpression;
 import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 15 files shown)


abayer:
Yeah, not gonna go in this direction. Closing.

PR #243, opened 2018-01-29
[FIXED JENKINS-49226] Add equals condition.

abayer:
* JENKINS issue(s):
    * [JENKINS-49226](https://issues.jenkins-ci.org/browse/JENKINS-49226)
* Description:
    * This adds a new `equals(expected:..., actual:...)` `when` condition. It basically does what you'd expect: checks if the two values are equal. This is intended to replace some of the most common usages of `when` `expression`s. I'm still trying to think of other common `expression` use cases that should have their own specific conditions...
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1425
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/EqualsConditional.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/EqualsConditional.java
@@ -1,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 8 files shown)


rsandell:
Thinking of some of what we discussed in #240 (perhaps using a `comparator` parameter, and my todo comment [here](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/240/files#diff-8283e844f00ebd795630679a036efa30R301)
Could do a similar thing here with a `comparator` parameter so this could also do regexp and ant path matching (but default to string equals).
Could potentially be added after #240 is merged.

abishek3876:
Is there a reason why scripts are encouraged less and custom DSL is preferred more in pipelines? If it's for security or some other reason it makes sense, but otherwise people trying to write pipelines will be more comfortable with groovy scripting syntax rather than learn a new DSL just for this. People are not allowed a learning curve for setting up CI in most organizations.

PR #244, opened 2018-01-30
Add the ability to make validator contributors optional.

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * I've got a potential use case for this in some awfully speculative stuff, but this won't cause any harm and lets us be more flexible in some cases for validation down the road.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
@@ -26,22 +26,28 @@ package org.jenkinsci.plugins.pipeline.modeldefinition;
 import hudson.slaves.DumbSlave;
 import jenkins.model.OptionalJobProperty;
 import org.apache.commons.lang.StringUtils;
+import org.codehaus.groovy.control.MultipleCompilationErrorsException;
+import org.hamcrest.Matchers;
 import org.jenkinsci.Symbol;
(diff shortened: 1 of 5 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java
```diff
@@ -882,4 +895,34 @@ public void bigIntegerFailure() throws Exception {
...
+
+    @Test
+    public void enableOptionalValidator() throws Exception {
+        String script = pipelineSourceFromResources("simplePipeline");
+
+        assertNotNull(Converter.scriptToPipelineDef(script));
+
+        thrown.expect(MultipleCompilationErrorsException.class);
+        thrown.expectMessage(Matchers.containsString("Echo is banned"));
+        Converter.scriptToPipelineDef(script, Collections.<Class<? extends DeclarativeValidatorContributor>>singletonList(RejectEchoStep.class));
```
rsandell: So no user facing way of enabling this? Only code?

abayer: Correct.

abayer:
I honestly think this can go in prior to the next minor release - the changes are only relevant at the beginning of a Pipeline run (i.e., when validation is run), so running builds will be just fine. I don't believe the parsing/validation/transformation phase is durable anyway - it depends on some transient fields.

PR #245, opened 2018-02-07
[JENKINS-42688] More powerful conditions in post

abayer:
* JENKINS issue(s):
    * [JENKINS-42688](https://issues.jenkins-ci.org/browse/JENKINS-42688)
* Description:
    * Specifically, allow reusing `when` conditions in `post`. The syntax is still up for negotiation, but we do need to be able to specify both the `when` condition and the `steps`, so I think the syntax I've got is as good as it gets. See below for an example.
   * Old-school `success`, `always`, etc will still work as well. To retain backwards compatibility with some...idiosyncratic behavior, we'll first check to see if *any* conditions (traditional `post` or the new `when`-based ones) are satisified, then process the satisfied `when`-based conditions, and *then* iterate over the traditional `post` conditions, checking if it's satisified, and then executing. This is for cases, like, say, doing `junit` in `always` and wanting to fire something in `unstable`.
    * Might be worth considering adding an ordinal of some sort to the `when`-based `conditional`s, so that you can control order of execution and evaluation there too?
    * Still needs testing of the new conditions - right now I've just got backwards compat testing.

```
post {
  condition {
    when {
      allOf {
        branch "master"
        buildStatus status:"SUCCESS"
      }
    }
    steps {
      echo "tada"
    }
  }
}
```

* Documentation changes:
    * Oh my yes.
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ pipeline {
             parallel {
                 stage("linux") {
                     agent {
-                        label "java"
+                        label "highmem"
                     }
                     steps {
                         sh 'mvn clean install -Dmaven.test.failure.ignore=true'
@@ -66,7 +66,7 @@ pipeline {
                 }
                 stage("linux-newer-core") {
                     agent {
-                        label "java"
+                        label "highmem"
                     }
                     steps {
                         sh "mvn clean install -Dmaven.test.failure.ignore=true -Djava.level=8 -Djenkins.version=${NEWER_CORE_VERSION}"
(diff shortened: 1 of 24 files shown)


batmat:
@abayer is this ready for review? This shows up under my https://github.com/pulls/review-requested URL :). 

Thanks!

bitwiseman:
Do the `condition` directives run in the order they appear (unlike status conditions)?  `always` should run before any other conditions.  

I'm sorry for not giving feedback on this earlier.  Mostly this feels to me like one more step toward Scripted pipeline.  

What is the win here?  Once a user hits this level of complexity, it makes more sense for them to move the functionality into a shared library where they can use if statements.  Strong -1 for this feature as whole, sorry. 

Assuming we've already decided this feature is going to happen, the name `condition` is  too generic/overloaded.  We already call the status-based directives "conditions" in the documentation.  `conditional`?  `complex`?  `complex-condition`?  The fact that I can't think of a good term for this is further indication that this is not the place for this.

abayer:
This is still very very WIP and I am at this point far more likely to abandon it than try to pick it up. It was worth an experiment but that's probably it.

abayer:
This is a failed effort - closing.

PR #246, opened 2018-02-20
[FIXED JENKINS-49558] Add containerPerStage option

abayer:
* JENKINS issue(s):
    * [JENKINS-49558](https://issues.jenkins-ci.org/browse/JENKINS-49558)
* Description:
    * If given with a top-level Docker or Dockerfile agent, the top-level
will magically switch to just a label agent, while each stage without
an explicit agent specified will end up using the root agent
definition with reuseNode enabled. The result will be that each stage
gets a new container. Tada.
    * This could theoretically cause problems in resuming builds after
upgrade, but not likely, since we actually instantiate the
DeclarativeAgent instance just in time, not at the beginning of the
build. So the new field AbstractDockerAgent#containerPerStageRoot
should get populated correctly.
    - [x] Doc PR created.
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1443
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -486,7 +486,7 @@ class ModelInterpreter implements Serializable {
     def inDeclarativeAgent(Object context, Root root, Agent agent, Closure body) {
         if (agent == null
             && root.agent.getDeclarativeAgent(root, root) instanceof AbstractDockerAgent
-            && root.options?.options?.get("containerPerStage") != null) {
+            && root.options?.options?.get("newContainerPerStage") != null) {
             agent = root.agent
         }
         if (agent == null) {
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -28,6 +28,7 @@ import com.cloudbees.groovy.cps.impl.CpsClosure
 import hudson.FilePath
 import hudson.Launcher
 import hudson.model.Result
+import org.jenkinsci.plugins.pipeline.modeldefinition.agent.AbstractDockerAgent
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.*
 import org.jenkinsci.plugins.pipeline.modeldefinition.options.DeclarativeOption
 import org.jenkinsci.plugins.pipeline.modeldefinition.steps.CredentialWrapper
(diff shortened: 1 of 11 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -483,6 +484,11 @@ class ModelInterpreter implements Serializable {
      * @return The return of the resulting executed closure
      */
     def inDeclarativeAgent(Object context, Root root, Agent agent, Closure body) {
+        if (agent == null
```
rsandell: oooo, here we go with the hard coding shei* again :) We were so close at cleaning all that up :D

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/options/impl/ContainerPerStage/help.html
```diff
@@ -0,0 +1,27 @@
...
+  ~ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  ~ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  ~ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+
+<p>
+    If specified, any stage using the top-level Docker or Dockerfile agent will get a new container on the same node.
```
rsandell: new container __instance__?

rsandell: Yes technically a container is an instance of an image, so saying a new container means a new instance. But I thought the extra word might help docker newbs understand more of what it means.

Could also add an essay below describing it more than just the one sentence. Or maybe that's a topic for the jenkins.io documentation?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/options/impl/ContainerPerStage/help.html
abayer: Yeah, that's for jenkins.io.

abayer:
I'm fine with the longer name - renamed to `newContainerPerStage`.

PR #248, opened 2018-02-26
[FIXED JENKINS-41060] Add new fixed/regression post conditions

abayer:
* JENKINS issue(s):
    * [JENKINS-41060](https://issues.jenkins-ci.org/browse/JENKINS-41060)
* Description:
    * Rather than trying to make `changed` take an additional parameter and break a bunch of parsing, let's make it simpler: here's two new `post` conditions: `fixed` and `regression`. 
    * `fixed` runs if the previous build exists and wasn't `SUCCESS`, and the current build is `SUCCESS`.
    * `regression` runs if the previous build exists and the current build's status is worse than the previous build - i.e., `currentResult.isWorseThan(previousResult)`.
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1444
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
@@ -1,0 +1,65 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.plugins.pipeline.modeldefinition.model.conditions
+
+import hudson.Extension
+import hudson.model.Result
+import org.jenkinsci.Symbol
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
+import org.jenkinsci.plugins.workflow.job.WorkflowRun
+
+/**
+ * A {@link BuildCondition} for matching builds where the previous build was not SUCCESS but the current build is.
+ *
+ * @author Andrew Bayer
+ */
+@Extension(ordinal=900d) @Symbol("fixed")
+class Fixed extends BuildCondition {
+    @Override
+    boolean meetsCondition(WorkflowRun r) {
+        Result execResult = getExecutionResult(r)
+        // Only look at the previous completed build.
+        WorkflowRun prev = r.getPreviousCompletedBuild()
+
+        // Get the *worst* result of either the execution or the run. If the run's result is null, that's effectively
+        // SUCCESS.
+        Result runResult = execResult.combine(r.getResult() ?: Result.SUCCESS)
+
+        // If there's no previous build, we can't exactly be fixed, can we?
+        if (prev == null) {
+            return false
+        } else {
+            return runResult == Result.SUCCESS && prev.getResult() != Result.SUCCESS
+        }
+    }
+
+    @Override
+    String getDescription() {
+        return Messages.Fixed_Description()
+    }
+
+
+    static final long serialVersionUID = 1L
+
+}
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
@@ -29,6 +29,8 @@ import org.jenkinsci.Symbol
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
 import org.jenkinsci.plugins.workflow.job.WorkflowRun
 
+import javax.annotation.Nonnull
+
 /**
  * A {@link BuildCondition} for matching builds where the previous build was not SUCCESS but the current build is.
  *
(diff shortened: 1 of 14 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+        // Get the *worst* result of either the execution or the run. If the run's result is null, that's effectively
+        // SUCCESS.
+        Result runResult = execResult.combine(r.getResult() ?: Result.SUCCESS)
+
+        // If there's no previous build, we can't exactly be fixed, can we?
+        if (prev == null) {
+            return false
+        } else {
+            return runResult == Result.SUCCESS && prev.getResult() != Result.SUCCESS
```
rsandell: Should `ABORTED` and `NOT_RUN` really be a part of this?

batmat: +1, I think it's worth making those not trigger the block IMO. Again discussable, but aborted is not a failure/unstable IMO, and this is probably the only two states I care about when using a `fixed` post condition.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Regression.groovy
```diff
@@ -0,0 +1,65 @@
...
+import org.jenkinsci.Symbol
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
+import org.jenkinsci.plugins.workflow.job.WorkflowRun
+
+/**
+ * A {@link BuildCondition} for matching builds where the previous build was better than the current build.
+ *
+ * @author Andrew Bayer
+ */
+@Extension(ordinal=880d) @Symbol("regression")
```
batmat: same :)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Regression.groovy
```diff
@@ -0,0 +1,65 @@
...
+class Regression extends BuildCondition {
+    @Override
+    boolean meetsCondition(WorkflowRun r) {
```
batmat: NIT: `r` /me is not at east with one-char var names :)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Regression.groovy
```diff
@@ -0,0 +1,65 @@
...
+        Result execResult = getExecutionResult(r)
+        // Only look at the previous completed build.
+        WorkflowRun prev = r.getPreviousCompletedBuild()
+
+        // Get the *worst* result of either the execution or the run. If the run's result is null, that's effectively
+        // SUCCESS.
+        Result runResult = execResult.combine(r.getResult() ?: Result.SUCCESS)
```
batmat: Can we not factor out that code ^, this looks duplicated with `fixed.groovy` isn't it?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Regression.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+        // If there's no previous build, we can't exactly be regressing, can we?
+        if (prev == null) {
+            return false
+        } else {
+            return runResult.isWorseThan(prev.getResult())
```
batmat: :+1: (though same, I suspect this will be enabled for `ABORTED` when one probably does not expect it)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Messages.properties
```diff
@@ -28,4 +28,6 @@ Changed.Description=Run if the current build's status is different than the prev
 Failure.Description=Run if the build status is "Failure"
 NotBuilt.Description=Run if the build status is "Not Built"
 Success.Description=Run if the build status is "Success" or hasn't been set yet
-Unstable.Description=Run if the build status is "Unstable"
\ No newline at end of file
+Unstable.Description=Run if the build status is "Unstable"
+Fixed.Description=Run if the current build is "Success" and the previous build is not
+Regression.Description=Run if the current build's status is worse than the previous build's status
```
batmat: Shouldn't we double the single quotes as we normally should in `.properties`?  Not sure where those messages are exposed to users to check how the line 30 above is rendered already in the existing releases.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
abayer: Okiedokie.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
```diff
@@ -0,0 +1,65 @@
...
+import org.jenkinsci.Symbol
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
+import org.jenkinsci.plugins.workflow.job.WorkflowRun
+
+/**
+ * A {@link BuildCondition} for matching builds where the previous build was not SUCCESS but the current build is.
+ *
+ * @author Andrew Bayer
+ */
+@Extension(ordinal=890d) @Symbol("fixed")
```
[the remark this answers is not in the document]
abayer: Check the other ones - `Changed` is 900, so I wanted to have `Fixed` and `Regression` run after `Changed`, so slightly lower numbers. This is why I used 800, 900, 1000, etc as the original ordinals: room to expand. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
```diff
@@ -0,0 +1,65 @@
...
+class Fixed extends BuildCondition {
+    @Override
+    boolean meetsCondition(WorkflowRun r) {
```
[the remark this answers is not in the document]
abayer: Yeah, I can annotate.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Messages.properties
abayer: Good catch.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Regression.groovy
abayer: Eh, legacy. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Regression.groovy
abayer: Yeah, I can do that.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Regression.groovy
abayer: I can live with that. It *is* a regression by every Jenkins standard. =)

PR #249, opened 2018-03-02
[WiP] When condition checking if the current agent is a windows agent

rsandell:
* JENKINS issue(s):
    * N/A
* Description:
    * When condition checking if the current agent is a windows agent
* Documentation changes:
    * needed
* Users/aliases to notify:
    * @abayer


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/WindowsConditionalScript.groovy
```diff
@@ -0,0 +1,44 @@
...
+class WindowsConditionalScript extends DeclarativeStageConditionalScript<WindowsConditional> {
+
+    WindowsConditionalScript(CpsScript s, WindowsConditional c) {
+        super(s, c)
+    }
+
+    @Override
+    boolean evaluate() {
+        CpsThread t = CpsThread.current()
+        return !t?.getContextVariables()?.get(Launcher.class)?.isUnix()
```
rsandell: I don't think this works, but needed to upload to check

pipeline-model-definition/src/test/resources/when/whenWindows.groovy
```diff
@@ -0,0 +1,50 @@
...
+
+pipeline {
+    agent any
+    stages {
+        stage("On Windows") {
+            when {
+                windows()
+            }
+            steps {
+                echo "Hello Windows"
```
MarkEWaite: I think it would be better to use a Windows specific step to illustrate
```suggestion
                bat "echo Hello Windows World"
```

pipeline-model-definition/src/test/resources/when/whenWindows.groovy
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+        stage("Not Windows") {
+            when {
+                not {
+                    windows()
+                }
+            }
+            steps {
+                echo "Hello rest of world"
```
MarkEWaite: I think it would be better to use a step which does not run on Windows in this case
```suggestion
                sh "echo Hello non-Windows world"
```

bitwiseman:
@rsandell 
I'm going to close this for now.  If we get a chance to do more on it, please feel free to reopen.

PR #250, opened 2018-03-02
[JENKINS-49893] Add a Declarative directive snippet generator

abayer:
* JENKINS issue(s):
    * [JENKINS-49893](https://issues.jenkins-ci.org/browse/JENKINS-49893)
* Description:
    * Adds a snippet generator specifically for Declarative directives (initially included: `agent`, `parameters`, `triggers`, and `options`), with only valid options included in the dropdowns, proper syntax generation, useful descriptions, etc.
    * You can take a look by checking out this PR, running `mvn hpi:run`, creating a Pipeline job, and going to the `Declarative Directive Generator` link in the side panel.
    * Downstream of https://github.com/jenkinsci/workflow-cps-plugin/pull/208
- [x] Tests that we actually generate the code we're expecting.
- [x] `when` directive support
- [x] `input` directive support (a bit janky, but it works)
- [x] Help HTML and explanatory text for each directive.
- [x] Figure out exactly what to call it in the sidepanel.
- [x] Improve `config.jelly` for various `Describable`s provided by Declarative.
- [x] Maybe tests for making sure the generated code is actually *right*? As in toss it into a `Jenkinsfile` with other minimal stuff, run it, make sure the right things happened? Probably makes sense, probably takes some effort. Will work on it. (NOTE: Gonna punt this for now. May revisit later)
- [x] `tools` directive
- [x] `libraries` directive
- [x] `environment` directive in some form (NOTE: Need to figure out credentials support somehow)
- [x] `post` directive in some form
- [x] `parallel` directive (as a placeholder) (NOTE: Ended up deciding to just have you choose whether the stage is gonna have parallel or steps)
- [x] `stage` directive
- [x] https://github.com/jenkinsci/workflow-cps-plugin/pull/208 gets merged and released
- [x] Doc PR
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1472
* Users/aliases to notify:
    * @reviewbybees


michaelneale:
this would be very much appreciated.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @Nonnull
+    public List<DirectiveDescriptor> getDirectives() {
+        return DirectiveDescriptor.all();
+    }
+
+    @Restricted(DoNotUse.class) // for stapler
+    public @CheckForNull
+    Item getItem(StaplerRequest req) {
```
rsandell: If it's for stapler I don't think package level access works. IIRC it needs to be public

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
```diff
@@ -0,0 +1,191 @@
...
+    @Restricted(DoNotUse.class)  // For stapler
+    public static String getSymbolForDescriptor(Descriptor d) {
+        if (d instanceof StepDescriptor) {
+            return ((StepDescriptor) d).getFunctionName();
+        } else {
+            Set<String> symbols = SymbolLookup.getSymbolValue(d);
+            if (!symbols.isEmpty()) {
+                return symbols.iterator().next();
+            } else {
+                return "(unknown)";
```
rsandell: It could maybe be a normal `Step`? In that case it would be the step's `descriptor.getFunctionName()` iirc.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
abayer: That's a weird copypasta - the `public` is on the line above. Fixing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
abayer: Yeah, look above. =)

bitwiseman:
@abayer Checked out, ran `mvn hpi:run`, there's  no option to create a pipeline in the UI.

bitwiseman:
@abayer 
This is great!

Can you make this a sibling of snippet generator (under "pipeline syntax") instead a whole new top level item? 

Will you be adding `stage`?

abayer:
@bitwiseman I can probably do that but it’ll involve a fair amount of work - we’d need an extension point in `workflow-cps` for contributing links to the `Pipeline Syntax` sidepanel. At least to start, I wasn’t sure it was worth it.

Not sure if I’ll add `stage` - the problem there being that at that point, we’ve basically got to implement a full Declarative editor. `steps` would be an obvious pain in the ass, and `post` is, for various reasons, a challenge to represent in the snippet generator approach.

bitwiseman:
@abayer 
Ah, yeah, I can see how adding an extension point would be more work.  On the other hand, adding a job-level sidebar item is less than great UX.  Declarative is Pipeline too, so it will be confusing to not have it under the Pipeline sidebar item.   But I'm not the boss of you and I'll still proclaim the awesomeness of this feature either way. 

Stage is pri 2, but ... Could `steps` be generated as `steps { /* pipeline steps here */ }`?

abayer:
Another couple reasons for no `stage` - no `environment` or `tools` (nor do I expect to add `environment`, since it doesn't actually make sense as a web form, though I may add `tools`), and the fun of nested `parallel` (and eventually `stages`) creating some really unpleasant forms/UI/code.

Thinking on the sidepanel stuff - I'm trying to figure out what the right way to do it would be.

abayer:
Ok, this is now downstream of https://github.com/jenkinsci/workflow-cps-plugin/pull/209 - which changes the sidepanel links for the `Pipeline Syntax` page(s) to come from an extension point. And now we implement that extension point twice here (for the directive generator and a link to the Declarative online docs). So the job-level sidepanel link directly to the directive generator is gone - you get there via `Pipeline Syntax` now.

abayer:
Back in work-in-progress - I think I see how I can do the other directives after all, albeit with placeholders for some things.

kshultzCB:
I realize this is a "if it hurts when you do that, don't do that" sort of thing, but...is there a way to stop users from endlessly nesting the same directive repeatedly? Here's an example with the `when` generator and a bunch of nested `allOf`s.:

image

abayer:
@kshultzCB - argh! I thought I'd already beaten that. I mean, it's theoretically allowed, but yeah. Lemme see.

abayer:
Ok, we're basically feature complete now. Still gotta figure out what the docs will look like, and I'd love UI/text critiques/improvements. =)

kshultzCB:
Since you've implemented `stage`, will `stages` make the cut once the time comes? Same question applies for the restart PR - e.g., will `isRestartedRun` show up in the options for `when`. That kind of thing.

I was about to ask if you could add maven and git to the `tools` snippet generator, because I didn't see it when running from `hpi:run`. Then I thought to myself, "self, why don't you go into tools config, add a maven, and see if it shows up by magic." I did that, and look - magic! Neat.

abayer:
No, don't think I'll add `stages` - it's too simplistic. =) New `when` conditions should show up automatically if they have a config page - I need to make sure to have checks for that, actually.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
```diff
@@ -0,0 +1,193 @@
...
+
+    @Restricted(DoNotUse.class) // accessed via REST API
+    public HttpResponse doGenerateDirective(StaplerRequest req, @QueryParameter String json) throws Exception {
+        // TODO is there not an easier way to do this? Maybe Descriptor.newInstancesFromHeteroList on a one-element JSONArray?
+        JSONObject jsonO = JSONObject.fromObject(json);
+        Jenkins j = Jenkins.getActiveInstance();
+        Class<?> c = j.getPluginManager().uberClassLoader.loadClass(jsonO.getString("stapler-class"));
+        DirectiveDescriptor descriptor = (DirectiveDescriptor)j.getDescriptor(c.asSubclass(AbstractDirective.class));
+        if (descriptor == null) {
+            return HttpResponses.plainText("<could not find " + c.getName() + ">");
```
rsandell: Shouldn't this be more of a 404 or something?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
rsandell: :man_facepalming:

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
```diff
@@ -0,0 +1,193 @@
...
+            if (!symbols.isEmpty()) {
+                return symbols.iterator().next();
+            } else {
+                return "(unknown)";
+            }
+        }
+    }
+
+    @Restricted(DoNotUse.class)
+    @Extension public static class PerJobAdder extends TransientActionFactory<Job> {
```
rsandell: Shouldn't this be removed?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/EnvironmentDirective.java
```diff
@@ -0,0 +1,114 @@
...
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class EnvironmentDirective extends AbstractDirective<EnvironmentDirective> {
```
rsandell: BTW shouldn't these be in the api plugin?

abayer: Nah, they're specific to the Directive Generator, so I want 'em in the same place as `DirectiveGenerator`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
abayer: Annoyingly, no - get rid of that and `someJob/directive-generator` stops working.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator.java
abayer: This is trying to fit the behavior we already have with Snippet Generator - we just need to get a string to show in the result box, and a 404 would break things.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfile/config.jelly
```diff
@@ -0,0 +1,55 @@
...
+    <f:entry field="args" title="Additional arguments">
+        <f:textbox/>
+    </f:entry>
+    <f:entry field="label" title="Label">
+        <f:textbox/>
+    </f:entry>
+    <f:entry field="reuseNode" title="Reuse Node">
+        <f:checkbox/>
+    </f:entry>
+    <f:entry field="customWorkspace" title="Custom Workspace">
```
rsandell: Perhaps some of these should be behind an advanced section?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipeline/config.jelly
```diff
@@ -0,0 +1,52 @@
...
+    <f:entry field="label" title="Label">
+        <f:textbox/>
+    </f:entry>
+    <f:entry field="reuseNode" title="Reuse Node">
+        <f:checkbox/>
+    </f:entry>
+    <f:entry field="customWorkspace" title="Custom Workspace">
+        <f:textbox/>
+    </f:entry>
+    <f:entry field="registryUrl" title="Registry URL">
```
rsandell: Perhaps some of these should be behind an advanced section?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/AgentDirective/config.jelly
```diff
@@ -0,0 +1,42 @@
...
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form" xmlns:l="/lib/layout">
+    <f:block>
+        See <a href="https://jenkins.io/doc/book/pipeline/syntax/#agent" target="_blank">the online documentation</a>
+        for more information on the <code>agent</code> directive.
+    </f:block>
+    <f:dropdownList name="agent" title="${descriptor.displayName}" help="${descriptor.getHelpFile('agent')}">
```
rsandell: So why doesn't a hetero-list work here?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
```diff
@@ -0,0 +1,102 @@
...
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:l="/lib/layout" xmlns:f="/lib/form" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:local="local">
+    <l:layout title="${%Pipeline Syntax: Directive Generator}" norefresh="true">
+        <st:include page="sidepanel"/>
+        <l:main-panel>
+            <f:form name="config">
+                <f:section title="${%Overview}"/>
+                <f:block>
+                    The <b>Directive Generator</b> allows you to generate the Pipeline code for a Declarative Pipeline
```
rsandell: This block should be a "${%blurb}" so that it can be translated.
But then again all the `getDisplayName`s are missing I18n as well.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
```diff
@@ -0,0 +1,102 @@
...
+                    directive, such as <code>agent</code>, <code>options</code>, <code>when</code>, and more. Choose the
+                    directive you're interested in from the dropdown, and then choose the contents of the directive from
+                    the new form. Once you've filled out the form with the choices and values you want for your directive,
+                    click <b>Generate Declarative Directive</b> and the Pipeline code will appear in the box below. You
+                    can copy that code directly into the <code>pipeline</code> block in your <code>Jenkinsfile</code>,
+                    for top-level directives, or into a <code>stage</code> block for stage directives. See
+                    <a href="https://jenkins.io/doc/pipeline/syntax/" target="_blank">the online syntax documentation</a>
+                    for more information on directives.
```
rsandell: How many directives can a directive direct... :)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
```diff
@@ -0,0 +1,102 @@
...
+                </d:taglib>
+                <f:dropdownList name="prototype" title="${%Sample Directive}">
+                    <local:listDirectives directives="${it.getDirectives()}"/>
+                </f:dropdownList>
+                <j:set var="id" value="${h.generateId()}"/>
+                <f:block>
+                    <input type="button" value="${%Generate Declarative Directive}" onclick="handlePrototype_${id}(); return false" class="submit-button primary"/>
+                    <f:textarea id="prototypeText_${id}" readonly="true" style="margin-top: 10px" />
+                    <script>
+                        function handlePrototype_${id}() {
```
rsandell: would be nice if thid was in it's own `.js` file so it can be linted during build. Maybe overkill.

rsandell: also, whitespaces please :)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/symbol-hetero-list.jelly
```diff
@@ -0,0 +1,146 @@
...
+    </j:forEach>
+
+    <div class="repeatable-insertion-point" />
+
+    <div class="prototypes to-be-removed">
+      <!-- render one prototype for each type -->
+      <j:set var="instance" value="${null}" />
+      <j:set var="descriptors" value="${h.filterDescriptors(it,attrs.descriptors)}" />
+      <j:forEach var="descriptor" items="${descriptors}" varStatus="loop">
+        <j:invokeStatic var="symbol" className="org.jenkinsci.plugins.pipeline.modeldefinition.generator.DirectiveGenerator" method="getSymbolForDescriptor">
```
rsandell: is `invokeStatic` working now? It used to be broken, or maybe that was just for static fields?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/EnvironmentDirective/config.jelly
```diff
@@ -0,0 +1,49 @@
...
+    <f:block>
+        See <a href="https://jenkins.io/doc/book/pipeline/syntax/#environment" target="_blank">the online documentation</a>
+        for more information on the <code>environment</code> directive.
+        <f:entry title="${%Environment Variables}" help="${descriptor.getHelpPage()}">
+            <f:repeatable var="env" items="${instance.getEnv()}">
+                <table width="100%">
+                    <f:entry title="${%Variable Name}" help="${descriptor.getEnvHelp('name')}">
+                        <f:textbox name="env.name" value="${env.name}" />
+                    </f:entry>
+                    <!-- TODO: Figure out how to support credentials here. -->
```
rsandell: Two different add buttons? Or a dropdown add button maybe.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/StageDirective/config.jelly
```diff
@@ -0,0 +1,45 @@
...
+  -->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:dg="/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator">
+    <f:block>
+        See <a href="https://jenkins.io/doc/book/pipeline/syntax/#stage" target="_blank">the online documentation</a>
+        for more information on the <code>stage</code> directive.
+        <f:entry field="name" title="${%Stage Name}">
+            <f:textbox/>
+        </f:entry>
+        <f:entry title="">
```
rsandell: radioblock or enumSet instead?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AnyOfConditional/config.jelly
```diff
@@ -0,0 +1,35 @@
...
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:dg="/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator">
+    <f:section title="${descriptor.displayName}">
+        <f:block>
+            <j:invokeStatic var="conditionals" className="org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditionalDescriptor" method="forGenerator" />
```
rsandell: This is usually when I go for a config.groovy instead of a config.jelly ;)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/NotConditional/help-child.html
```diff
@@ -0,0 +1,24 @@
...
+  ~
+  ~ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  ~ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  ~ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  ~ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+Get you some parameters.
```
rsandell: huh?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/OptionsDirective.java
```diff
@@ -0,0 +1,154 @@
...
+        @Nonnull
+        public List<Descriptor> getDescriptorsForContext(boolean inStage) {
+            List<Descriptor> descriptors = new ArrayList<>();
+
+            for (Descriptor d : ExtensionList.lookup(JobPropertyDescriptor.class)) {
+                Set<String> symbolValue = SymbolLookup.getSymbolValue(d);
+                if (!symbolValue.isEmpty()) {
+                    boolean blockedSymbol = false;
+                    for (String symbol : symbolValue) {
+                        if (Options.getBLOCKED_PROPERTIES().contains(symbol)) {
```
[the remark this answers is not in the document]
abayer: The warning bell is that this is how Groovy ends up exposing a method to get at a static variable in a class from Java. Is lame, but it works.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/OptionsDirective.java
```diff
@@ -0,0 +1,154 @@
...
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class OptionsDirective extends AbstractDirective<OptionsDirective> {
+    public static final List<String> ADDITIONAL_BLOCKED_STEPS = ImmutableList.of("script", "ws", "withEnv", "withCredentials",
```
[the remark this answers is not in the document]
abayer: Had to do some additional filtering on the available wrapper steps list just for UI purposes.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/OptionsDirective.java
```diff
@@ -0,0 +1,154 @@
...
+                    }
+                    if (!blockedSymbol && !inStage) {
+                        descriptors.add(d);
+                    }
+                }
+            }
+
+            for (DeclarativeOptionDescriptor d : ExtensionList.lookup(DeclarativeOptionDescriptor.class)) {
+                if (!SymbolLookup.getSymbolValue(d).isEmpty()) {
+                    if (!inStage || d.canUseInStage()) {
```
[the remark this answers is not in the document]
abayer: So this gets tweaked in the stage options PR, but what we're doing here is saying "Ok, if we're not in a stage, we assume all options are valid, and if we *are* in a stage, only those that are explicitly marked as usable in a stage are valid".

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/StageDirective.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+
+        @Override
+        @Nonnull
+        public String toGroovy(@Nonnull StageDirective directive) {
+            StringBuilder result = new StringBuilder("stage(");
+            result.append("'").append(directive.name).append("') {\n");
+            switch (directive.contentType) {
+                case 0:
```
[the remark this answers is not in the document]
abayer: I have a weird little hatred for enums.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineFromDockerfile/config.jelly
abayer: I decided not to bother with advanced blocks generally. May tweak later.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/AgentDirective/config.jelly
abayer: Hell if I know. I just poked things with sticks until they worked, I don't remember why they didn't in earlier versions. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
abayer: Yeah, on my longer term TODO.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
abayer: Yeah, this is all copy-pasted/tweaked from Snippetizer's stuff, so I didn't make any more changes than I had to in case that resulted in the world exploding.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/symbol-hetero-list.jelly
abayer: Well, it is working - there are other usages of it elsewhere in Jenkins, and it definitely works here.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/EnvironmentDirective/config.jelly
abayer: Yeah, the problems I had with credentials were...weird. Something to revisit in the future.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/StageDirective/config.jelly
abayer: *shrug* I have no strong feelings, but this does seem to work for now.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AnyOfConditional/config.jelly
abayer: I'm trying to avoid `config.groovy` these days. =)

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/NotConditional/help-child.html
abayer: ha, placeholder text. Fixing.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/StageDirective.java
```diff
@@ -0,0 +1,152 @@
...
+                    result.append("steps {\n");
+                    result.append("// One or more steps need to be included within the steps block.\n");
```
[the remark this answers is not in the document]
abayer: Everything gets auto-indented magically via the same method we use to indent the Jenkinsfiles translated from the editor - which wants everything to be left-aligned when passed to it so it can figure out the right indent properly.

abayer:
@rsandell - switched to an `enum` for `StageDirective#contentType` and corresponding change to the `config.jelly`.

PR #251, opened 2018-03-05
Matrix rewrite proof of concept

JeremyMarshall:
* JENKINS issue(s):
    * n/a
* Description:
    * Adds a new `axes` section in a `pipeline` which will multiply out the stages and 
      add environment variables for the axes. This does no loops in the declarative syntax 
      just rewrites before running
     Is a POC so only does serial stages.
    The DSL allows for `axes` in a stage itself (just for the stage its in) but this has not been 
    implemented in the code.
    Things to do are make the parallel stage work, add an order to the stages and maybe a
    way to remove various combinations
    e.g
 ```
pipeline {
    agent any
    options { buildDiscarder(logRotator(numToKeepStr: '1')) }
    axes {
       axis1('a','b','c')
       axis2('x','y','z')
    }
    stages {
                stage('Example 1') {
                   environment {
                       CC = 'clang'
                   }
                   agent any
                   steps {
                        sh ("env|sort|grep -i axis")
                   }
                }
                stage('Example 2') {
                   agent any
                    axes {
                       //not implemented
                       axis3('1','2','3')
                    }
                    steps {
                        sh ("env|sort|grep -i axis")
                    }
                }
    }
    post {
        always {
            echo 'I will always say Hello again!'
        }
    }
}
```
* Documentation changes:
    * None as yet, wait for feedback
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAxes.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAxes.java
@@ -1,0 +1,121 @@
+package org.jenkinsci.plugins.pipeline.modeldefinition.ast;
+
+import net.sf.json.JSONArray;
+import net.sf.json.JSONObject;
+import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator;
+
+import javax.annotation.Nonnull;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.List;
+
+/**
+ * Represents a map of axis types to axis names (i.e., the name of the configured installation). Corresponds to
+ * {@code axes}.
+ *
+ * @author Andrew Bayer
+ */
+public final class ModelASTAxes extends ModelASTElement {
+    private Map<ModelASTKey, List<ModelASTValue>> axes = new LinkedHashMap<>();
+
+    public ModelASTAxes(Object sourceLocation) {
+        super(sourceLocation);
+    }
+
+    @Override
+    public JSONArray toJSON() {
+        final JSONArray a = new JSONArray();
+        for (Map.Entry<ModelASTKey, List<ModelASTValue>> entry: axes.entrySet()) {
+            JSONObject o = new JSONObject();
+            o.accumulate("key", entry.getKey().toJSON());
+
+            for(ModelASTValue value: entry.getValue()) {
+                o.accumulate("value", value.toJSON());
+            }
+
+            a.add(o);
+        }
+        return a;
+    }
+
+    @Override
+    public void validate(@Nonnull final ModelValidator validator) {
+        validator.validateElement(this);
+        for (Map.Entry<ModelASTKey, List<ModelASTValue>> entry : axes.entrySet()) {
+            entry.getKey().validate(validator);
+
+            for(ModelASTValue value: entry.getValue()) {
+                value.validate(validator);
+            }
+        }
+    }
+
+    @Override
+    public String toGroovy() {
+        StringBuilder result = new StringBuilder("axes {\n");
+        for (Map.Entry<ModelASTKey, List<ModelASTValue>> entry : axes.entrySet()) {
+            result.append(entry.getKey().toGroovy()).append(" (");
+            String comma = " ";
+
+            for(ModelASTValue value: entry.getValue()) {
+                result.append(comma).append(value.toGroovy());
+                comma = ", ";
+            }
+            result.append(") \n");
+        }
+        result.append("}\n");
+        return result.toString();
+    }
+
+    @Override
+    public void removeSourceLocation() {
+        super.removeSourceLocation();
+        for (Map.Entry<ModelASTKey, List<ModelASTValue>> entry : axes.entrySet()) {
+            entry.getKey().removeSourceLocation();
+
+            for(ModelASTValue value: entry.getValue()) {
+                value.removeSourceLocation();
+            }
+        }
+    }
+
+    public Map<ModelASTKey, List<ModelASTValue>> getaxes() {
+        return axes;
+    }
+
+    public void setaxes(Map<ModelASTKey, List<ModelASTValue>> axes) {
+        this.axes = axes;
+    }
+
+    @Override
+    public String toString() {
+        return "ModelASTaxes{" +
+                "axes=" + axes +
+                "}";
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        if (!super.equals(o)) {
+            return false;
+        }
+
+        ModelASTAxes that = (ModelASTAxes) o;
+
+        return getaxes() != null ? getaxes().equals(that.getaxes()) : that.getaxes() == null;
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = super.hashCode();
+        result = 31 * result + (getaxes() != null ? getaxes().hashCode() : 0);
+        return result;
+    }
+}
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
@@ -111,14 +111,10 @@ public abstract class ModelASTElement implements Cloneable {
     }
 
     @Override
-    public Object clone()
-    {
-        try
-        {
+    public Object clone() {
+        try {
             return super.clone();
-        }
-        catch( CloneNotSupportedException e )
-        {
+        } catch( CloneNotSupportedException e ) {
             return null;
         }
     }
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
@@ -29,7 +29,7 @@ import net.sf.json.JSONObject;
 import org.codehaus.groovy.ast.ASTNode;
 import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator;
 
-public abstract class ModelASTElement {
+public abstract class ModelASTElement implements Cloneable {
     /**
      * The sourceLocation is a reference to whatever section of the original source we're parsed from corresponds to this
      * element. When parsed from Pipeline Script, it's an {@link ASTNode}, and when parsed from JSON, it's a {@link JSONObject}.
@@ -108,6 +108,19 @@ public abstract class ModelASTElement {
     @Override
     public int hashCode() {
         return ModelASTElement.class.hashCode();
+    }
+
+    @Override
+    public Object clone()
+    {
+        try
+        {
+            return super.clone();
+        }
+        catch( CloneNotSupportedException e )
+        {
+            return null;
+        }
     }
 }
 
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTPipelineDef.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTPipelineDef.java
@@ -22,6 +22,7 @@ public final class ModelASTPipelineDef extends ModelASTElement {
     private ModelASTBuildParameters parameters;
     private ModelASTTriggers triggers;
     private ModelASTLibraries libraries;
+    private ModelASTAxes axes;
 
     public ModelASTPipelineDef(Object sourceLocation) {
         super(sourceLocation);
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

**Comment by abayer** [comment]
Very interesting! Probably worth implementing on top of #227 (once I fix its merge conflict) - I was intending to do matrix using the sequential stage groups there. I'll dig into this more later - I'm ratholing on some config stuff at the moment. =)

File: pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAxes.java [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+    }
+
+    @Override
+    public String toGroovy() {
+        StringBuilder result = new StringBuilder("axes {\n");
+        for (Map.Entry<ModelASTKey, List<ModelASTValue>> entry : axes.entrySet()) {
+            result.append(entry.getKey().toGroovy()).append(" (");
+            String comma = " ";
```
abayer: This is a little mind-bending - `comma` equals `" "`? =) I get it, but there's probably a better way to do this - especially with this moved on top of #227 and therefore being Java 8. I just started playing with streams for the first time (I know, pathetically late, but hey) and that's pretty much perfect for this sort of thing - like, say:

```
result.append(entry.getValue().stream().map(ModelASTValue::toGroovy).collect(Collectors.joining(", ")));
```

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
```diff
@@ -29,7 +29,7 @@
 import org.codehaus.groovy.ast.ASTNode;
 import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator;
 
-public abstract class ModelASTElement {
+public abstract class ModelASTElement implements Cloneable {
```
abayer: Why do we want this `Cloneable`?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
```diff
@@ -109,5 +109,18 @@ public boolean equals(Object o) {
     public int hashCode() {
         return ModelASTElement.class.hashCode();
     }
+
+    @Override
+    public Object clone()
```
abayer: Formatting nitpick - I'd prefer consistent styling, i.e., `{` on the same line, `catch (CloneNotSupportedException e)` instead of `catch( CloneNotSupportedException e )`, etc.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTPipelineDef.java
```diff
@@ -22,6 +22,7 @@
     private ModelASTBuildParameters parameters;
     private ModelASTTriggers triggers;
     private ModelASTLibraries libraries;
+    private ModelASTAxes axes;
```
abayer: Like I mentioned in my first comment, I think the way to go here is to actually build this on top of the sequential stages in #227, as a `stage`-level construct only.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStages.java
```diff
@@ -14,7 +14,7 @@
  *
  * @author Andrew Bayer
  */
-public final class ModelASTStages extends ModelASTElement {
+public final class ModelASTStages extends ModelASTElement{
```
abayer: formatting nitpick. =) If you could revert it back to `extends ModelASTElement {`, that'd be great.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidator.java
```diff
@@ -25,29 +25,7 @@
...
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTPostStage;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStage;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStages;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStep;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTTools;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTTrigger;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTTriggers;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTWhen;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTWhenCondition;
+import org.jenkinsci.plugins.pipeline.modeldefinition.ast.*;
```
abayer: ah, import reformat fun! I don't particularly care that much about that - I have IntelliJ set up to never do star imports personally, but this isn't going to cause my IDE to reformat this again in the future, so no harm, no foul overall. But I do know other people can be more resistant to "formatting only" changes like this. Just an FYI to watch out for in the future.

pipeline-model-definition/pom.xml
```diff
@@ -66,7 +66,43 @@
...
+              <goal>report-only</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>findbugs-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
abayer: Please remove this - I want the findbugs errors. =)

pipeline-model-definition/pom.xml
```diff
@@ -66,7 +66,43 @@
...
+          <execution>
+            <phase>none</phase>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
abayer: Oh yeah, definitely remove this. The enforcer is wise and must be obeyed.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Axes.groovy
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * A map of axis types to axis name (i.e., specific installation's configured name) to install and add to the path and
+ * environment for the build.
+ *
+ * @author Andrew Bayer
+ */
+@SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
+class Axes extends MappedClosure<Closure,Axes> implements Serializable {
+
+//    private static final Object CACHE_KEY = new Object()
```
abayer: I'm assuming this is all copy-pasta that you just haven't cleaned out yet?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
```diff
@@ -57,11 +57,13 @@ class Root implements Serializable {
...
     Libraries libraries
 
+    Axes axes
+
     final String astUUID
 
     @Whitelisted
     Root(Agent agent, Stages stages, PostBuild post, Environment environment, Tools tools, Options options,
-         Triggers triggers, Parameters parameters, Libraries libraries, String astUUID) {
+         Triggers triggers, Parameters parameters, Libraries libraries, Axes axes, String astUUID) {
```
abayer: As I've mentioned, I think this should just be a `stage` level thing.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -482,6 +485,43 @@ class ModelParser implements Parser {
...
+                    def axsKey = parseKey(mc.method)
+
+                    List<Expression> args = ((TupleExpression) mc.arguments).expressions
+                    if (args.isEmpty()) {
+                        errorCollector.error(axsKey, Messages.ModelParser_NoArgForAxis(axsKey.key))
+                    } else {
+                        r.axes[axsKey] = []
+
+                        args.each { a ->
+                            if (!(a instanceof ConstantExpression) && !(a instanceof GStringExpression)) {
```
abayer: So this brings up an interesting question: do we want to be able to support other expressions in `axes`? I think we probably will - at the very least, we're going to want some way of specifying multiple axis values as a parameter (though probably not with [Extended Choice plugin](https://wiki.jenkins.io/display/JENKINS/Extended+Choice+Parameter+plugin) because it's kinda problematic), which would probably be referred to with something like `axisX(params.SOME_PARAM_NAME)`, which would end up being a list in practice. And we might well want to support letting users call a shared library function to get the axis list...

Hmm. As I think about it, maybe something like this as a syntax?
```
axes {
  axis(name: "LITERAL_AXIS", values: [1, 2, 3])
  axis(name: "PARAM_AXIS", values: params.SOME_MULTI_PARAM)
  axis(name: "FUNCTION_AXIS", values: someSharedLibMethod())
}
```

Your thoughts? =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -186,6 +186,7 @@ class RuntimeASTTransformer {
             // Filter for only the desired value type - ModelASTValue for env vars, ModelASTInternalFunctionCall for
             // credentials.
             if (v instanceof ModelASTElement && valueType.isInstance(v) && v.sourceLocation != null) {
+                println(v.sourceLocation.class)
```
abayer: Ah, yes, lurking `println`s. I miss them so often myself. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -604,6 +605,47 @@ class RuntimeASTTransformer {
...
 
+
+    /**
+     * Generates the axes combinations
+     *
+     * @param axes The incoming axes
+     * @param ret The resulting combinations
+     * @return The resulting combinations.
+     */
+    List GeneratePermutations(Map axes){
```
abayer: `List generatePermutations(Map axes)` - and in practice we can hopefully end up getting it typed better.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
JeremyMarshall: When duplicating stuff later I just cloned it. I was adding thing into the environment and it seemed easier to add it into the base class

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidator.java
JeremyMarshall: fixed in a rebase to master, I don't want to lose the commentary

pipeline-model-definition/pom.xml
JeremyMarshall: reverted - I didn't realise I needed to build the parent directory first

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Root.groovy
JeremyMarshall: I'll wait for the other PR and move it to nested stages

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
JeremyMarshall: I would think. I do like things declarative so calling out and not having repeatable. I suppose that argument has been had before.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
JeremyMarshall: I like this one as it dumps the finished product out to retry. but yes I'll get rid of it

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
JeremyMarshall: It has been better typed before but I took it out

bitwiseman:
@JeremyMarshall @abayer This is really interesting and exciting.  I suspect there would be quite a bit of interest in this feature.  It would be great if there were a JIRA for this.  Even better would be to start up a [Jenkins Enhancement Proposal](https://github.com/jenkinsci/jep) and collect feedback.  You've already started a prototype implementation. 😄

bitwiseman:
@JeremyMarshall 
Matrix is now out in 1.4.0-beta1.  I'd love your feedback.

PR #252, opened 2018-03-07
[FIXED JENKINS-49976] Get rid of ModelASTValue anonymous classes

abayer:
* JENKINS issue(s):
    * [JENKINS-49976](https://issues.jenkins-ci.org/browse/JENKINS-49976)
* Description:
    * Based on other PRs I've seen, getting rid of the anonymous classes seems to do the trick to get rid of the warnings. So.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTValue.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTValue.java
@@ -115,47 +115,60 @@ public abstract class ModelASTValue extends ModelASTElement implements ModelASTM
     }
 
     public static ModelASTValue fromConstant(final Object o, Object sourceLocation) {
-        return new ModelASTValue(sourceLocation, o) {
-            @Override
-            public boolean isLiteral() {
-                return true;
-            }
-
-            @Override
-            public String toGroovy() {
-                if (getValue() instanceof String) {
-                    String str = (String) getValue();
-                    str = str.replace("\\", "\\\\");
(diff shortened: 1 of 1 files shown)


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTValue.java
```diff
@@ -115,47 +115,60 @@ public String toString() {
...
-                    String str = (String) getValue();
-                    str = str.replace("\\", "\\\\");
-                    if (str.indexOf('\n') == -1) {
-                        return "'" + (str.replace("'", "\\'")) + "'";
-                    } else {
-                        return "'''" + (str.replace("'", "\\'")) + "'''";
-                    }
-                } else if (getValue() != null) {
-                    return getValue().toString();
+    public static final class ConstantValue extends ModelASTValue {
```
jglick: May as well be `private`.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTValue.java
```diff
@@ -115,47 +115,60 @@ public String toString() {
...
+        }
     }
 
-    public static ModelASTValue fromGString(final String gstring, Object sourceLocation) {
-        return new ModelASTValue(sourceLocation, gstring) {
-            @Override
-            public boolean isLiteral() {
-                return false;
-            }
+    public static final class GStringValue extends ModelASTValue {
```
jglick: ditto

PR #253, opened 2018-03-08
Move CredentialsBindingHandler extension point to extensions plugin

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * This way, plugins will be able to just depend on pipeline-model-extensions if they want to implement a handler. I also moved the actual implementations out of CredentialsBindingHandler and into a better package - I couldn't move CredentialsBindingHandler itself into a better package because I didn't want to break compatibility.
* Documentation changes:
    * n/a - internal relocation only
* Users/aliases to notify:
    * @reviewbybees


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/credentials/impl/FileCredentialsHandler.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018, CloudBees, Inc.
```
rsandell: 2016 - 2018

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/CredentialsBindingHandler.java
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016, CloudBees, Inc.
+ * Copyright (c) 2018, CloudBees, Inc.
```
rsandell: 2016 - 2018

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/credentials/impl/StringCredentialsHandler.java
```diff
@@ -0,0 +1,64 @@
...
+    public Class<? extends StandardCredentials> type() {
+        return StringCredentials.class;
+    }
+
+    @Nonnull
+    @Override
+    public List<Map<String, Object>> getWithCredentialsParameters(String credentialsId) {
+        Map<String, Object> map = new HashMap<>();
+        map.put("$class", StringBinding.class.getName());
+        map.put("variable", new EnvVarResolver());
```
rsandell: What `EnvVarResolver` are these classes picking up? The original was a static internal class. I didn't think those were inherited?

abayer: They are inherited, apparently. I guess we could move `EnvVarResolver` to its own file as well for clarity.

rsandell: Unless someone has started using it already?

abayer: Yeah, better safe than sorry - let's leave it as is.

PR #254, opened 2018-03-09
[FIXED JENKINS-47106] Support alwaysPull in the top level agent section

segevfiner:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-47106
* Description:
    * Support `alwaysPull` in the top level agent section, was broken by https://github.com/jenkinsci/pipeline-model-definition-plugin/commit/a367db7c2a727e5ea87381777b72b4c84b52f1a2#diff-f6e51fe76a49913b4ff00606e7d4d28aL43
* Documentation changes:
    * This is a bug fix.
* Users/aliases to notify:
    * @abayer

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
@@ -39,7 +39,7 @@ class DockerPipelineScript extends AbstractDockerPipelineScript<DockerPipeline> 
     @Override
     Closure runImage(Closure body) {
         return {
-            if (!Utils.withinAStage()) {
+            if (!Utils.withinAStage() && describable.alwaysPull) {
                 script.stage(SyntheticStageNames.agentSetup()) {
                     try {
                         script.getProperty("docker").image(describable.image).pull()


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerPipelineScript.groovy
```diff
@@ -39,7 +39,7 @@ class DockerPipelineScript extends AbstractDockerPipelineScript<DockerPipeline>
     @Override
     Closure runImage(Closure body) {
         return {
-            if (!Utils.withinAStage()) {
+            if (!Utils.withinAStage() && describable.alwaysPull) {
```
abayer: Hrrrrm. So I'm assuming what you want here is the ability to *not* always pull? Because the existing behavior for a top-level `docker` agent is to always pull.

segevfiner: Yes. That's what the issue is about. `alwaysPull` is ignored in the top level agent directive.

I think that users often do want to "always pull" and not pulling is mostly used for local/quick testing. So you might want to reconsider the default of `alwaysPull` which I think is `false`.

abayer: So that makes me want to think on this for a while. Right now the effective behavior is "always explicitly pull for a top-level agent, do not explicitly pull for a stage agent unless `alwaysPull` is set". So if we add `alwaysPull` checking to the top level and turn the default for `alwaysPull` to false, the new effective behavior will be "always pull for either top level or stage agent, unless `alwaysPull` is explicitly set to `false`".

That concerns me a bit - we deliberately set up the current behavior for stage-level agents and I don't think we want to change that. The top-level behavior for `docker` is different than at the stage level for probably no longer relevant historical reasons - we wanted to keep the `docker pull ...` logs inside a stage that Blue Ocean knows to not display unless there's a problem. But we're seeing a lot more stage-level `docker` agent usage than we expected at the time, so whether or not `alwaysPull` is set, any time someone's using a Docker image that isn't already cached locally, the Docker pull logs are showing up in that stage's log anyway. That is a little different than at the top level, because if we get rid of the explicit `docker pull` call, the pull logs will show up when `docker run` is called and won't be in any stage at all.

Hrrrrrm.

segevfiner: Also keep in mind that not using `docker pull` might end up using stale images and failing builds that require an up-to-date image. But sometimes it's useful to not `docker pull`, e.g. testing a local image.

abayer: Ok, let's live a little - merging, but if there are reports of issues, we may revert later. Thanks!

PR #255, opened 2018-03-13
[JENKINS-48050] introduce "container" agent type

ndeloof:
* JENKINS issue(s): 
  https://issues.jenkins-ci.org/browse/JENKINS-48050

* Description:
   introduce "container" agent type with plugable implementation, + default docker-plugin based implementation

* Users/aliases to notify:
   @abayer


abayer:
Needs a time stamped SNAPSHOT of docker-plugin deployed and used as the dependency.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeDockerUtils.java
```diff
@@ -120,4 +121,15 @@ public static String getRegistryCredentialsId(@Nullable String override) {
             return null;
         }
     }
+
+    public static ContainerAgentProvider getProvider() {
```
abayer: Could we get some javadoc here and elsewhere to explain what these are doing?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/ContainerAgentProvider.java
```diff
@@ -0,0 +1,7 @@
+package org.jenkinsci.plugins.pipeline.modeldefinition.agent.impl;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class ContainerAgentProvider {
```
abayer: Definitely needs javadoc since there’s nothing actually in here to help figure it out.

pipeline-model-definition/pom.xml
```diff
@@ -158,6 +164,12 @@
       <version>1.1.1</version>
     </dependency>
 
+    <dependency>
+      <groupId>io.jenkins.docker</groupId>
+      <artifactId>docker-plugin</artifactId>
+      <version>1.2-SNAPSHOT</version>
```
abayer: Use the time stamped snapshot version here.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/ContainerAgentProvider.java
abayer: Ah, I see this is incomplete - this looks, from other code, to actually be an extension point?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerAgent/help.html
```diff
@@ -0,0 +1,27 @@
...
+  ~ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  ~ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  ~ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+
+<p>
+    The job or stage will run within a Docker container.
```
abayer: Please add a config.jelly for supporting the changes coming in #250

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/ContainerAgentProvider.java
ndeloof: would "Provides ContainerAgent" javadoc be really helpful here ?
isn't class name obvious ?

pipeline-model-definition/pom.xml
ndeloof: correct me if I'm wrong, but we don't publish timestamped snapshots for plugins' pull-request builds
https://ci.jenkins.io/job/Plugins/job/docker-plugin/job/JENKINS-48050/8/console

abayer: Correct. You need to do a mvn deploy by hand. Also, make sure these new dependencies are in dependencyManagement in the parent pom.

ndeloof:
> I know this is work in progress, but it really, really needs tests.

see https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/255/commits/6bd544434e24ee03b2ec9f1f459e764b64a235bb

abayer:
@ndeloof Thanks - still needs more tests for making sure things like `environment` interact properly, stage-level agents...it would be great to see if https://issues.jenkins-ci.org/browse/JENKINS-48319 and https://issues.jenkins-ci.org/browse/JENKINS-46831 are issues with `container` or not, etc.

ndeloof:
Those are already covered by https://github.com/jenkinsci/docker-plugin/blob/master/src/test/java/io/jenkins/docker/pipeline/DockerNodeStepTest.java#L181

abayer:
@ndeloof I’d still like to see tests in Declarative itself.

ndeloof:
JENKINS-48319, JENKINS-46831: those are irrelevant with container agents are those are just plain agents, without anything shared between a "top level" and "stage" agent. `container` probably won't address the use case those issues have been reported for. We could write a test case for them with plain label agent as well, this would not demonstrate anything has been fixed.

abayer:
I’d still like tests demonstrating that the issues are not relevant for the container implementation. It’s always better to overdo it with tests than to miss something.

abayer:
And again, please do a `mvn deploy` of docker-plugin yourself, and use the timestamp SNAPSHOT version as the dependency here.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/ContainerAgent.java
```diff
@@ -0,0 +1,80 @@
...
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class ContainerAgent extends DeclarativeAgent<ContainerAgent> {
+
+    private final String image;
+
+    private String registryCredentials;
+
+    @DataBoundConstructor
+    public ContainerAgent(@Nonnull String image) {
```
abayer: We need some more setters/fields - `String type` for overriding the default `container` implementation, for certain, and probably some way to specify sidecar containers and a way to specify implementation-specific configuration.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/DockerContainerAgentProvider.java
```diff
@@ -0,0 +1,40 @@
...
+        return SCRIPT;
+    }
+
+    @Override
+    public Descriptor<ContainerAgentProvider> getDescriptor() {
+        return Jenkins.getInstance().getDescriptor(DockerContainerAgentProvider.class);
+    }
+
+    @Extension
+    public static class DescriptorImpl extends Descriptor<ContainerAgentProvider> {
```
abayer: Needs a `@Symbol` for use in specifying an overriding implementation.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/config/GlobalConfig.java
```diff
@@ -72,6 +79,19 @@ public void setRegistry(DockerRegistryEndpoint registry) {
         this.registry = registry;
     }
 
+    public ContainerAgentProvider getProvider() {
```
abayer: We need this same logic in `FolderConfig` too.

abayer:
Please fix the findbugs error and the failing unit tests. Thanks.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/ContainerAgent.java
ndeloof: sidecar containers is out of scope for JENKINS-48050 phase 1
(but easy to add later)
I don't understand the need for "type". All implementation should create the exact same container deployment, just using distinct API (i.e docker or kubectl)

bitwiseman:
@ndeloof @abayer @jglick 
It's been over a year since last update.  Should we close or move this forward?

jglick:
@bitwiseman FWIW my counterproposal is https://github.com/jenkinsci/docker-plugin/pull/681. Anything to put a nail in the coffin of `docker-workflow` (https://github.com/jenkinsci/docker-workflow-plugin/pull/148#issuecomment-438820089).

ndeloof:
inactive, feel free to take over

PR #256, opened 2018-03-27
Allow stage-only DeclarativeOptions

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * No concrete use case for this right now, but I've got some possibilities I'm brewing that would want this elsewhere, so let's get it in so that it'll be available in 1.2.8.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...


abayer:
@rsandell fwiw, https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/256/files/51f25e2d69d9662dcdd8b9792739c337b03ef67c..e60068f48aa6dbb42189696620d613ff3791da45 is the relevant changes here.

rsandell:
@abayer yepp, that part I think looks good, but this can't be merged until the snippetizer is approved and merged, so I can't really approve this either :)

abayer:
Fair enough!

abayer:
That's a good point - will revisit.

PR #257, opened 2018-04-02
WIP: Create our own per-test timeout rule.

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * JenkinsRule's timeout logic goes batty when JenkinsRule is used as a ClassRule - it ends up applying to the *entire* test class, not just the individual test. So that's annoying. Here, I copy-pasted some stuff around to create a per-test Rule that still does the nice thread-dump stuff on timeout. This is just a hack, mind you. Still trying to think how I'd actually do this in JenkinsRule...
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,7 +33,7 @@ pipeline {
                         label "highmem"
                     }
                     steps {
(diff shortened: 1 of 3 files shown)


abayer:
Well, that didn't work.

PR #258, opened 2018-04-03
[FIXED JENKINS-41239] Add new cleanup post condition

abayer:
* JENKINS issue(s):
    * [JENKINS-41239](https://issues.jenkins-ci.org/browse/JENKINS-41239)
* Description:
    * The new `cleanup` condition will always run regardless of build status, like `always`, but runs *after* all other `post` conditions have been evaluated, rather than before.
    * I wanted to call this `finally` but that actually breaks Groovy parsing because `finally` is a reserved word. Dang. So for now, I'm calling it `cleanup`, but am open to suggestions.
* Documentation changes:
    * TODO
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Cleanup.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Cleanup.groovy
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2017, CloudBees, Inc.
+ * Copyright (c) 2018, CloudBees, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


rpocase:
Love that this is coming, but is there a better word? Perhaps `last`? Using `cleanup` implies a specific effect to me (e.g. it would run `deleteDir()` every time).

abayer:
@rpocase True Story: trying to think of a name is part of why I haven’t done this earlier. =) Not being able to use `finally` really blocked out the best name. I thought about `last` but...I dunno, something doesn’t quite feel right. Not ruling it out, mind you - hoping for more feedback and suggestions.

deiga:
Possible names: `eventually`, `ultimately`, `lastly`, `atLast`, `aftermath`, `conclusion`, `conclusively`

Moritz90:
I personally think cleanup is a great name. I'm sure I'm not the only one who wants to use it for `cleanWs()`.

deiga:
@Moritz90 Is there a need for calling `cleanWs()`? Doesn't jenkins automatically delete the temporary workspace it created?

steven-foster:
@deiga not in my experience. Jenkins does have a periodic cleanup thread that runs every 24 hours and deletes workspaces over 30 days old. I'm not actually sure if that affects pipeline workspaces. For me that's too long/infrequent so I make sure to `deleteDir()` in the pipeline

bitwiseman:
@deiga - Nope. This allows workspaces to be reused.

bitwiseman:
@abayer
I'm mostly okay with `cleanup` - it communicates the intended use.  "Generally, use `always` but if you need to do some last clean up use `cleanup`."   

I'd also vote for `lastly`.  It's close to `finally`.  

Another option: `afterAll`.

abayer:
Hmmmm. I might be able to talk myself into `lastly`. My only real concern with `cleanup` is that people might *assume* it would automatically do a `cleanWs()` or the like - I mean, the whole point of the condition basically *is* to do clean-up after the other conditions have run, so calling it `cleanup` isn't a problem in that regard.

rsandell:
My first idea was `lastly` as well, as a synonym to `finally`.
http://www.thesaurus.com/browse/finally

rsandell:
But given the hope that this would be called `finally` and going through #259 I would expect this post condition to always run even though some earlier post condition had failed. Can't remember if that is the case though.

abayer:
@rsandell It will run regardless of status of earlier post blocks, yes.

PR #259, opened 2018-04-03
[FIXED JENKINS-44993] Don't swallow errors in post

abayer:
* JENKINS issue(s):
    * [JENKINS-44993](https://issues.jenkins-ci.org/browse/JENKINS-44993)
* Description:
    * We only throw one exception for a failure in a stage, whether that failure is in the `steps` block or in a `post` condition's execution. Any subsequent exceptions get swallowed and never logged. This will change that, so that we'll start logging any exceptions encountered in executing `post` blocks even if the stage has already hit an error before getting to the `post` blocks.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
@@ -138,7 +138,7 @@ public class BuildConditionResponderTest extends AbstractModelDefTest {
     public void buildConditionOrdering() throws Exception {
         expect("buildConditionOrdering")
                 .logContains("[Pipeline] { (foo)", "hello")
-                .inLogInOrder("I AM ALWAYS", "I CHANGED", "I SUCCEEDED")
+                .logContainsInOrder("I AM ALWAYS", "I CHANGED", "I SUCCEEDED")
                 .go();
     }
 
@@ -312,6 +312,7 @@ public class BuildConditionResponderTest extends AbstractModelDefTest {
                         "hello",
                         "[Pipeline] { (" + SyntheticStageNames.postBuild() + ")",
                         "I FAILED",
+                        "CLEANUP RAN",
                         "Error when executing failure post condition",
                         "Error when executing always post condition",
                         "No such property: otherUndefined for class",
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
@@ -303,4 +303,20 @@ public class BuildConditionResponderTest extends AbstractModelDefTest {
         j.assertLogNotContains("I FAILED", run1);
 
     }
+
+    @Issue("JENKINS-44993")
+    @Test
+    public void failureInPostBlock() throws Exception {
+        expect(Result.FAILURE, "failureInPostBlock")
+                .logContains("[Pipeline] { (foo)",
+                        "hello",
+                        "[Pipeline] { (" + SyntheticStageNames.postBuild() + ")",
+                        "I FAILED",
+                        "Error when executing failure post condition",
+                        "Error when executing always post condition",
+                        "No such property: otherUndefined for class",
+                        "No such property: undefined for class")
+                .logNotContains("MOST DEFINITELY FINISHED")
+                .go();
+    }
 }
--- a/pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
+++ b/pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
@@ -42,6 +42,9 @@ pipeline {
             echo "I FAILED"
             echo "SOME ${otherUndefined} VAR"
         }
+        cleanup {
+            echo "CLEANUP RAN"
+        }
     }
 }
 
--- a/pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
+++ b/pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
@@ -1,0 +1,49 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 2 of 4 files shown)


pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
```diff
@@ -0,0 +1,49 @@
...
+                echo "hello"
+            }
+        }
+    }
+    post {
+        always {
+            echo "SOME ${undefined} VAR"
+        }
+        success {
+            echo "MOST DEFINITELY FINISHED"
```
bitwiseman: Add a test where the failure occurs here.

pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
```diff
@@ -0,0 +1,49 @@
...
+    stages {
+        stage("foo") {
+            steps {
+                echo "hello"
+            }
+        }
+    }
+    post {
+        always {
+            echo "SOME ${undefined} VAR"
```
bitwiseman: Add an echo that doesn't fail before this this one..

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
```diff
@@ -242,4 +242,21 @@ public void changedAndNotSuccess() throws Exception {
...
+    @Issue("JENKINS-44993")
+    @Test
+    public void failureInPostBlock() throws Exception {
+        expect(Result.FAILURE, "failureInPostBlock")
+                .logContains("[Pipeline] { (foo)",
+                        "hello",
+                        "[Pipeline] { (" + SyntheticStageNames.postBuild() + ")",
+                        "I FAILED",
+                        "Error when executing failure post condition",
+                        "Error when executing always post condition",
```
bitwiseman: I would expect this line to occur before "I FAILED".

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
```diff
@@ -242,4 +242,21 @@ public void changedAndNotSuccess() throws Exception {
...
+                        "No such property: otherUndefined for class",
+                        "No such property: undefined for class")
```
bitwiseman: I'd expect this line to occur immediately after "Error when executing always post condition",

bitwiseman:
This is a really cool improvement.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
abayer: This isn't ordered - it's just "are each of these strings present in the log somewhere?"

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
abayer: See above.

pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
abayer: Not sure what that'd give us? The point of the test is just to make sure that we actually log all the exceptions thrown in the `post` conditions - I can add that test, just not sure there's a reason for it.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
bitwiseman: @abayer 
Ah, okay.  Well, we'll need to add a `logContainsInOrder` or equivalent.
The order of these lines matters and depending on implementation they might occur in a different order, including the one you shown here is not.

bitwiseman: Oh, looks like you already have it implemented: 
[inLogInOrder](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java#L390)

pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
bitwiseman: It's to show that things behave as they should even in the exception throwing cases.  
For  example, we want to verify that the same behavior occurs in the `success` condition as in the `always`.  

I don't know if it matters but `failure` has a higher ordinal than `success`.  Does that mean failure will have already been considered when `success` throws an exception and as such will not execute?  Same question for `unstable` and etc.

pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
```diff
@@ -0,0 +1,49 @@
...
+        }
+        success {
+            echo "MOST DEFINITELY FINISHED"
+        }
+        failure {
+            echo "I FAILED"
+            echo "SOME ${otherUndefined} VAR"
+        }
```
rsandell: should add `cleanup` _(or whatever we decide to call it)_ here as well when #258 is merged.

abayer: Good point.

pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
abayer: Gotcha. Will do. And yes, that does mean that an error in `success` will not result in `failure` being called - which has been the case all along. All we're doing here is surfacing the errors that could previously have been swallowed silently, not changing the flow behavior.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
abayer: I completely forgot that!

bitwiseman: @abayer Maybe rename it to `logContainsInOrder` so it is easier to find/remember?  `inLog...` is quite odd.

pipeline-model-definition/src/test/resources/failureInPostBlock.groovy
bitwiseman: I can see the logic of it - the `failure` condition should only fire if the stage failed not if one of the post condition steps failed - but that is not consistent.  By this same logic, the failure condition shouldn't fire if the steps all passed but there is a failure in the `always` condition (which does happen currently). 
Filed https://issues.jenkins-ci.org/browse/JENKINS-50645.

abayer:
@bitwiseman Looking to merge this shortly - any chance you could re-review and let me know what, if anything, needs to be changed?

PR #260, opened 2018-04-04
Add DeclarativeDirective extension point for adding directives

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * New directives must provide a method for parsing from Jenkinsfile syntax, a JSON schema, a runtime model class, validation for the parsed AST model, and a method for transforming from the parsed AST model to the runtime model.
    * Initially, will only provide support for preprocessing directives - which will provide a way to transform the full AST model after parsing and before validation and transformation. This will enable plugin authors to contribute directives that can be replaced by predetermined content before validation starts.
* Documentation changes:
    * Actually, this makes me think that developer docs for *all* the Declarative extension points would be a good idea...
* Users/aliases to notify:
    * ...


abayer:
Note - this is a work-in-progress. I need to add tests and do some prototyping for a use case I have in mind to make sure this implementation will actually do the trick. I also need to decide if it's actually worth it to keep this as general as it is, or if it makes more sense to just provide a pre-processing directive extension point...

rsandell:
By just reading the title I fear this could lead to the "declarative model" getting out of hand if there is free reign for other plugins to add to the structure.

bitwiseman:
@abayer  This seems like it could _really_ use a JEP.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/DeclarativeDirectiveDescriptor.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public abstract class DeclarativeDirectiveDescriptor<D extends DeclarativeDirective<D>> extends Descriptor<D> {
+    private static Logger LOGGER = Logger.getLogger(DeclarativeDirectiveDescriptor.class.getName());
+
+    @Nonnull
+    public abstract String getName();
```
rsandell: You seem to be using this in quite a lot of places that confuses me on the distinction between name and displayName.
But I can't come up with a better word for this either :)

bitwiseman:
@abayer  Any chance you can resolve the merge conflicts here?  Should this remain open?

PR #261, opened 2018-04-09
[FIXED JENKINS-50682] Actually pass descriptor to when condition forms

abayer:
* JENKINS issue(s):
    * [JENKINS-50682](https://issues.jenkins-ci.org/browse/JENKINS-50682)
* Description:
    * Also add display name for a few `when` conditions, move all of `ChangeRequestConditional`'s fields into "advanced", since they're optional and all, and smack things around a bit internally to not display `when` directive parameters when they're not required and they've got an empty string value.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/WhenDirective.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/WhenDirective.java
@@ -29,13 +29,16 @@ import hudson.model.Descriptor;
 import org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditional;
 import org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditionalDescriptor;
 import org.jenkinsci.plugins.structs.describable.DescribableModel;
+import org.jenkinsci.plugins.structs.describable.DescribableParameter;
 import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
 import org.jenkinsci.plugins.workflow.cps.Snippetizer;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import javax.annotation.Nonnull;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public class WhenDirective extends AbstractDirective<WhenDirective> {
     private DeclarativeStageConditional conditional;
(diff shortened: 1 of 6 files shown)


rtyler:
Not sure what value I can add as a reviewer, so here's a helpful screenshot:

selection_388

abayer:
@rtyler I just like to make you feel involved.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/WhenDirective.java
```diff
@@ -81,12 +84,29 @@ public String getDisplayName() {
...
                 if (model != null) {
+                    Map<String,Object> args = new HashMap<>();
+                    // Hack to null out non-required fields for UI prettiness.
+                    for (String argKey : ud.getArguments().keySet()) {
+                        DescribableParameter p = model.getParameter(argKey);
+                        if (p == null ||
+                                p.isRequired() ||
+                                !(String.class.equals(p.getErasedType())) ||
+                                !ud.getArguments().get(argKey).equals("")) {
+                            args.put(argKey, ud.getArguments().get(argKey));
```
rsandell: Quite inefficient, you should iterate over `ud.getArguments().entrySet()` instead of doing map lookups all the time.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/WhenDirective.java
```diff
@@ -81,12 +84,29 @@ public String getDisplayName() {
...
-                DescribableModel model = ud.getModel();
+                DescribableModel<? extends DeclarativeStageConditional> model = ud.getModel();
                 if (model != null) {
+                    Map<String,Object> args = new HashMap<>();
+                    // Hack to null out non-required fields for UI prettiness.
+                    for (String argKey : ud.getArguments().keySet()) {
+                        DescribableParameter p = model.getParameter(argKey);
+                        if (p == null ||
+                                p.isRequired() ||
+                                !(String.class.equals(p.getErasedType())) ||
```
rsandell: I don't think I've seen the use of `String.class.equals` before, looks weird. `String` is final so I don't think you need anything complicated to check if something is a string or not.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/WhenDirective.java
rsandell: So you merged without addressing this?

PR #262, opened 2018-04-11
[FIXED JENKINS-50652] Don't fire post failure for aborted durable task

abayer:
* JENKINS issue(s):
    * [JENKINS-50652](https://issues.jenkins-ci.org/browse/JENKINS-50652)
* Description:
    * It appears that an aborted durable task like sh results in `CpsFlowExecution#result` being `ABORTED`, but `WorkflowRun#result` being `FAILURE`, until the whole build actually finishes and the run gets updated accordingly. So the `failure` `post` condition shouldn't fire if `CpsFlowExecution#result` is `ABORTED`.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
@@ -41,7 +41,8 @@ class Failure extends BuildCondition {
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r) {
         Result execResult = getExecutionResult(r)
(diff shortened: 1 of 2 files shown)


abayer:
Joy, hanging tests I can't reproduce locally! Sigh. Working on it.

svanoort:
> Joy, hanging tests I can't reproduce locally! Sigh. Working on it.

🖖

abayer:
wooo, it was yet more fun with `@ClassRule` and various `JenkinsRule`/test framework stuff that...doesn't quite work as expected when `JenkinsRule` is used as a `@ClassRule`. In this case, it was that `semaphore 'wait'` in two different tests in the same class would end up with multiple hits, so I had to make the `semaphore` names unique across the whole class, not just a single test. Fun!

PR #264, opened 2018-04-17
[JENKINS-43016] Convert empty string label agent to agent any in JSON

abayer:
* JENKINS issue(s):
    * [JENKINS-43016](https://issues.jenkins-ci.org/browse/JENKINS-43016)
* Description:
    * This'll make the editor happier - the label agent type does require an argument, but an empty string *is* a valid (but weird) use case. The editor, however, has no way of distinguishing between "you didn't specify a label" and "you specified an empty string for the label". There still should be work in the editor to figure out how to actually handle this situation, but this fix at least will unbreak existing Jenkinsfiles with empty string label agents being opened in the editor, by converting the JSON in that case to be `agent any`, which is functionally equivalent if no additional options are specified.
    * However, if additional options are specified (such as `customWorkspace`), we have to punt and keep using the empty string label agent.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @reviewbybees

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
@@ -51,9 +51,9 @@ public final class ModelASTAgent extends ModelASTElement {
                 Map<ModelASTKey, ModelASTMethodArg> vars = ((ModelASTClosureMap) variables).getVariables();
                 // Don't actually switch to "agent any" if there are any additional options besides the label.
                 if (vars.size() == 1) {
-                    for (ModelASTKey argKey : vars.keySet()) {
-                        if (argKey.getKey().equals("label")) {
-                            ModelASTMethodArg argValue = vars.get(argKey);
+                    for (Map.Entry<ModelASTKey,ModelASTMethodArg> entry : vars.entrySet()) {
+                        if (entry.getKey().getKey().equals("label")) {
+                            ModelASTMethodArg argValue = entry.getValue();
                             if (argValue instanceof ModelASTValue && ((ModelASTValue) argValue).getValue().equals("")) {
                                 return true;
                             }
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
@@ -4,6 +4,7 @@ import net.sf.json.JSONObject;
 import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator;
 
 import javax.annotation.Nonnull;
+import java.util.Map;
 
 /**
  * Represents what context in which to run the build - i.e., which label to run on, what Docker agent to run in, etc.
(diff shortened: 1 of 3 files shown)


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
```diff
@@ -22,19 +23,49 @@ public ModelASTAgent(Object sourceLocation) {
...
+                return true;
+            }
+            if (variables instanceof ModelASTClosureMap) {
+                Map<ModelASTKey, ModelASTMethodArg> vars = ((ModelASTClosureMap) variables).getVariables();
+                // Don't actually switch to "agent any" if there are any additional options besides the label.
+                if (vars.size() == 1) {
+                    for (Map.Entry<ModelASTKey,ModelASTMethodArg> entry : vars.entrySet()) {
+                        if (entry.getKey().getKey().equals("label")) {
+                            ModelASTMethodArg argValue = entry.getValue();
+                            if (argValue instanceof ModelASTValue && ((ModelASTValue) argValue).getValue().equals("")) {
```
rsandell: Is there a risk of `getValue()` returning `null`? If so maybe turn around the equals to `"".equals(((ModelASTValue) argValue).getValue())` or `StringUtils.isEmpty(((ModelASTValue) argValue).getValue())` to safeguard against NPE?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
```diff
@@ -22,19 +23,49 @@ public ModelASTAgent(Object sourceLocation) {
...
+                } else if (variables instanceof ModelASTValue) {
+                    j.accumulate("argument", variables.toJSON());
+                }
             }
         }
         return j;
     }
 
+    private boolean isEmptyStringLabelAgent() {
+        if (agentType.getKey().equals("label") || agentType.getKey().equals("node")) {
```
rsandell: I feel like I've missed some past large change, when was `agent node` introduced?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParserTest.java
```diff
@@ -63,6 +63,35 @@ public void labelWithOptionsBecomesNode() throws Exception {
...
+        assertEquals(getJSONErrorReport(jp, "agentLabelEmptyString"), 0, jp.getErrorCollector().getErrorCount());
+        assertNotNull("Pipeline null for agentLabelEmptyString", newRoot);
+
+        JSONObject anyJson = JSONObject.fromObject(fileContentsFromResources("json/agentAny.json"));
+
+        JSONParser anyParser = new JSONParser(Converter.jsonTreeFromJSONObject(anyJson));
+        ModelASTPipelineDef anyRoot = anyParser.parse();
+
+        assertEquals(getJSONErrorReport(anyParser, "agentAny"),
+                0, anyParser.getErrorCollector().getErrorCount());
```
rsandell: I see you checking that there are no errors, and that might be the actual thing to test. But shouldn't you check that it's actually `agent any` that it has been converted to and just not something that doesn't have errors? I mean `agent label ' '` might not have errors, but actually not empty?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAgent.java
abayer: Yeesh, it was a while ago. I can't honestly remember. It's just an alias for `label`, really.

rsandell: Yeah, I have some slither of memory about that now that you mention it :)

abayer:
Waiting for feedback from @kzantow =)

PR #265, opened 2018-04-17
[FIXED JENKINS-50815] Fallback to GIT_LOCAL_BRANCH for when branch

abayer:
* JENKINS issue(s):
    * [JENKINS-50815](https://issues.jenkins-ci.org/browse/JENKINS-50815)
* Description:
    * If we're not on multibranch (meaning BRANCH_NAME is null), fallback to GIT_LOCAL_BRANCH for when branch condition. We can theoretically add other SCMs here too if needed. We're using GIT_LOCAL_BRANCH because it most closely matches BRANCH_NAME - i.e., GIT_BRANCH includes the remote, GIT_LOCAL_BRANCH does not.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BranchConditionalScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BranchConditionalScript.groovy
@@ -37,7 +37,7 @@ class BranchConditionalScript extends DeclarativeStageConditionalScript<BranchCo
     @Override
      boolean evaluate() {
         String branchName = (String)script.getProperty("env").getProperty("BRANCH_NAME")
-        if (branchName == null) {
+        if (branchName == null || branchName == "") {
             // fall back to GIT_LOCAL_BRANCH - note that GIT_BRANCH will have the remote on it, while BRANCH_NAME doesn't.
             branchName = (String)script.getProperty("env").getProperty("GIT_LOCAL_BRANCH")
         }
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
@@ -1051,13 +1051,9 @@ public class BasicModelDefTest extends AbstractModelDefTest {
                 .go();
     }
 
-    @Ignore("This breaks on PCT, so re-enable when we depend on newer core than 2.60")
     @Issue("JENKINS-45198")
     @Test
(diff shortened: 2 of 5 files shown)


abayer:
Oh joy, I can't reproduce the failures locally.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BranchConditionalScript.groovy
```diff
@@ -36,7 +36,11 @@ class BranchConditionalScript extends DeclarativeStageConditionalScript<BranchCo
 
     @Override
      boolean evaluate() {
-        return describable.branchMatches(describable.compare,
-            (String)script.getProperty("env").getProperty("BRANCH_NAME"))
+        String branchName = (String)script.getProperty("env").getProperty("BRANCH_NAME")
+        if (branchName == null || branchName == "") {
+            // fall back to GIT_LOCAL_BRANCH - note that GIT_BRANCH will have the remote on it, while BRANCH_NAME doesn't.
```
imod: typo: ...while `GIT_LOCAL_BRANCH` doesn't.

abayer:
@bitwiseman - it seems to check out a branch for a `CpsScmFlowDefinition`, at least. The test passes (or, well, it will now that I fixed the `environment` syntax to correctly override `BRANCH_NAME` from the PR build itself so that the test can actually be relevant!).

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BranchConditionalScript.groovy
```diff
@@ -36,7 +36,11 @@ class BranchConditionalScript extends DeclarativeStageConditionalScript<BranchCo
 
     @Override
      boolean evaluate() {
-        return describable.branchMatches(describable.compare,
-            (String)script.getProperty("env").getProperty("BRANCH_NAME"))
+        String branchName = (String)script.getProperty("env").getProperty("BRANCH_NAME")
+        if (branchName == null || branchName == "") {
```
rsandell: `StringUtils.isEmpty(branchName)`

rsandell: well actually this is groovy so:
```groovy
 if(!branchName?.trim())
```
maybe? I'm a bit unsure about the `trim()` part :)

rsandell: Or maybe not :) https://gist.github.com/kdabir/2867339

PR #266, opened 2018-04-17
[FIXED JENKINS-50645] Move failure to run after other status checks

abayer:
* JENKINS issue(s):
    * [JENKINS-50645](https://issues.jenkins-ci.org/browse/JENKINS-50645)
* Description:
    * This way failure will still evaluate even if the actual failure doesn't occur until in the success or unstable blocks.
* Documentation changes:
    * Docs will need to reflect the changed order.
* Users/aliases to notify:
    * @reviewbybees


abayer:
Note that I'm not touching `aborted` here - it already runs before `success` et al, and figuring out what exactly we'd want to do differently with `aborted` isn't really clear to me at the moment. So if there are requests for that, please open a new JIRA etc. =)

PR #270, opened 2018-05-16
Apostrophes get eaten

jsoref:
We created a broken pipeline {} and it yielded this:
```
WorkflowScript: 41: Stage doesnt have a block @ line 41, column 9.
```

`'` is a fairly hard to use character.

* In general for most messages, using contractions is a bad idea, so the first commit here expands them.
* @jglick prefers using fancy quotes over `'`. And one nice thing about them is that one can actually see start and end points w/ them, whereas `'` is ambiguous.


bitwiseman:
This works around a bug in the Blue Ocean by replacing all `'` with a other characters? 
It looks like this expected behavior in the MessageFormatter class, with the workaround being to [escape single-quote (`'`) using two single-quotes (`''`)](https://stackoverflow.com/questions/4449639/apostrophe-doesnt-get-translated-properly-when-placed-in-a-resource-bundle).

jsoref:
Yes, a2dd51b as part of #250 tried to "escape" things but used the wrong escape character (I've been asked to write a test for this (https://github.com/jenkinsci/jenkins-test-harness) and I hope to at some point.

Since jenkins (at least @jglick) prefers using fancy quotes, this basically avoids the issue, and reserves `'` for the rare case where it, itself is escaping something: https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/a9fbdce7fe51171b2106ec183cd1381b3266b1ef/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/Messages.properties#L60

bitwiseman:
@jsoref 
I've approved this change as it is a valid fix for an issue. Thank you!

Overall I get your reasoning, but I want to point out that @jglick also prefers to use open double-quotes (`/u201C`) and close double-quotes (`/u201D`) instead of simple double-quotes (`"`), but we don't use those in messages.  😞 The lack of a consistent stated style guide means all we can do is muddle through, I guess.

jsoref:
For reference:
> [ERROR] Failed to execute goal org.jvnet.localizer:maven-localizer-plugin:1.24:generate (default) on project pipeline-model-definition: Execution default of goal org.jvnet.localizer:maven-localizer-plugin:1.24:generate failed: can't parse argument number:  ... : For input string: " ... " -> [Help 1]

https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html
> Within a String, a pair of single quotes can be used to quote any arbitrary characters except single quotes. For example, pattern string "'{0}'" represents string "{0}", not a FormatElement. A single quote itself must be represented by doubled single quotes '' throughout a String. For example, pattern string "'{''}'" is interpreted as a sequence of '{ (start of quoting and a left curly brace), '' (a single quote), and }' (a right curly brace and end of quoting), not '{' and '}' (quoted left and right curly braces): representing string "{'}", not "{}".

What happened here was the single quotes were actually performing a function:
```diff
@@ -155,1 +155,1 @@ WhenConditionalValidator.changelog.badPa
-ModelInterpreter.NoNodeContext=Attempted to execute a step that requires a node context while \'agent none\' was specified. Be sure to specify your own \'node { ... }\' blocks when using \'agent none\'.
+ModelInterpreter.NoNodeContext=Attempted to execute a step that requires a node context while \u2018agent none\u2019 was specified. Be sure to specify your own \u2018node { ... }\u2019 blocks when using \u2018agent none\u2019.
```
The first set of single quotes and the last set of single quotes were just someone trying to talk to the reader. But the middle set of single quotes were being honored by the MessageFormatter. Note: in all three cases, the MessageFormatter ate the quotes, the reader never saw any of them. But, only in the middle case did the quotes actually perform a useful function....

Thus, the fix (relative to the above) is:
```diff
@@ -155,1 +155,1 @@ WhenConditionalValidator.changelog.badPa
-ModelInterpreter.NoNodeContext=Attempted to execute a step that requires a node context while \u2018agent none\u2019 was specified. Be sure to specify your own \u2018node { ... }\u2019 blocks when using \u2018agent none\u2019.
+ModelInterpreter.NoNodeContext=Attempted to execute a step that requires a node context while \u2018agent none\u2019 was specified. Be sure to specify your own \u2018node '{ ... }'\u2019 blocks when using \u2018agent none\u2019.
```
By tightening the scope to the `{ ... }` which are the MessageFormatter characters which cause issues, we also clarify what we're guarding.

I can't figure out where the actual `generate failed` and `[Help 1]` messages come from...
I think localizer is https://github.com/kohsuke/localizer/tree/master/lib/src/main/java/org/jvnet/localizer
I think maven-localizer-plugin is: https://github.com/kohsuke/localizer/tree/master/maven-plugin/src/main/java/org/jvnet/localizer
(I know that the `can't parse argument number` part comes from `MessageFormat.format` and friends.)

It would be nice to be able to improve the error reporting.
It should be possible for the error report to identify the file name, line number, and message id to help people figure out what's going on in the future.

abayer:
Thanks, @jsoref!

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/Messages.properties
```diff
@@ -152,7 +152,7 @@ ModelValidatorImpl.MissingInputMessage=No message specified for input
 WhenConditionalValidator.changelog.missingParameter=Changelog is missing required parameter "pattern".
 WhenConditionalValidator.changelog.badPattern="{0}" is not a valid regular expression. {1}
 
-ModelInterpreter.NoNodeContext=Attempted to execute a step that requires a node context while \'agent none\' was specified. Be sure to specify your own \'node { ... }\' blocks when using \'agent none\'.
+ModelInterpreter.NoNodeContext=Attempted to execute a step that requires a node context while \u2018agent none\u2019 was specified. Be sure to specify your own \u2018node '{ ... }'\u2019 blocks when using \u2018agent none\u2019.
```
jglick: I do not get it. What would

```groovy
node '{ ... }'
```

possibly mean? I think the `'`s should be deleted here just as they were in the other two cases in this line.

jsoref: @jglick: they're talking to `MessageFormatter.format`, and they tell it "This `{` and its contents including the `...` and the closing brace `}` are not to be treated like `{0}`"  -- something which it would try to substitute with an argument.

PR #273, opened 2018-06-29
[FIXED JENKINS-52261] Mark end nodes for skipped-for-restart stages properly

abayer:
* JENKINS issue(s):
    * [JENKINS-52261](https://issues.jenkins-ci.org/browse/JENKINS-52261)
* Description:
    * We were marking the *start* node for stages skipped for restart with `NotExecutedNodeAction`, but not the end nodes, and it turns out, `StatusAndTiming#computeChunkStatus2` looks at the end node. So...whoops.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -75,6 +75,7 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecution
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode
 import org.jenkinsci.plugins.workflow.graph.BlockStartNode
 import org.jenkinsci.plugins.workflow.graph.FlowNode
+import org.jenkinsci.plugins.workflow.graphanalysis.DepthFirstScanner
 import org.jenkinsci.plugins.workflow.graphanalysis.Filterator
 import org.jenkinsci.plugins.workflow.graphanalysis.FlowScanningUtils
 import org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner
@@ -306,7 +307,7 @@ class Utils {
         }
     }
 
-    static void markStageWithTag(String stageName, String tagName, String tagValue, boolean isRestart = false) {
+    static void markStageWithTag(String stageName, String tagName, String tagValue) {
         List<FlowNode> matched = findStageFlowNodes(stageName)
 
         matched.each { currentNode ->
@@ -320,10 +321,43 @@ class Utils {
                     tagsAction.addTag(tagName, tagValue)
                     currentNode.save()
                 }
+            }
+        }
+    }
 
-                if (isRestart) {
-                    currentNode.addAction(new NotExecutedNodeAction())
+    static void markStartAndEndNodesInStageAsNotExecuted(String stageName, FlowExecution execution = null) {
+        if (execution == null) {
+            CpsThread thread = CpsThread.current()
+            execution = thread.execution
+        }
+
+        List<BlockStartNode> nodes = []
+
+        ForkScanner scanner = new ForkScanner()
+
+        FlowNode stage = scanner.findFirstMatch(execution.currentHeads, null, isStageWithOptionalName(stageName))
+
+        if (stage != null && stage instanceof BlockStartNode) {
+            nodes.add(stage)
+
+            // Additional check needed to get the possible enclosing parallel branch for a nested stage.
+            Filterator<FlowNode> filtered = FlowScanningUtils.fetchEnclosingBlocks(stage)
+                .filter(isParallelBranchFlowNode(stageName))
+
+            filtered.each { f ->
+                if (f != null && f instanceof BlockStartNode) {
+                    nodes.add(f)
                 }
+            }
+        }
+
+        DepthFirstScanner depthFirstScanner = new DepthFirstScanner()
+
+        nodes.each { n ->
+            n.addAction(new NotExecutedNodeAction())
+            FlowNode endNode =  depthFirstScanner.findFirstMatch(execution.currentHeads, null, endNodeForStage(n))
+            if (endNode != null) {
+                endNode.addAction(new NotExecutedNodeAction())
             }
         }
     }
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -173,6 +173,9 @@ class ModelInterpreter implements Serializable {
                             firstError = e
                         }
                     }
+                    if (skippedForRestart) {
+                        Utils.markStartAndEndNodesInStageAsNotExecuted(thisStage.name)
+                    }
                 }
             } finally {
                 // And finally, run the post stage steps if this was a parallel parent.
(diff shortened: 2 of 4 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -320,12 +321,45 @@ class Utils {
...
+                    nodes.add(f)
                 }
             }
         }
+
+        DepthFirstScanner depthFirstScanner = new DepthFirstScanner()
+
+        nodes.each { n ->
+            n.addAction(new NotExecutedNodeAction())
+            FlowNode endNode =  depthFirstScanner.findFirstMatch(execution.currentHeads, null, endNodeForStage(n))
```
abayer: @svanoort - I'm particularly interested to hear whether there's a better way to do this (get the end node for a start node), so comments more than welcome. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -320,12 +321,45 @@ class Utils {
                     tagsAction.addTag(tagName, tagValue)
                     currentNode.save()
                 }
+            }
+        }
+    }
+
+    static void markStartAndEndNodesInStageAsNotExecuted(String stageName, FlowExecution execution = null) {
```
svanoort: So, if I understand correctly, we're hunting for the start/end nodes for ALL stages and and parallels labelled with stageName?  There may be a nuance here I'm missing but here's a simpler way I'd propose rather than doing the multi-pass search:

```java
public void markStageAndBranchNotExecuted(FlowExecution exec, String labelName) {
        DepthFirstScanner scan = new DepthFirstScanner();
        Predicate MATCH = new Predicate<FlowNode>() {
            @Override
            public boolean apply(@Nullable FlowNode x) {
                if (x instanceof BlockStartNode) {
                    LabelAction label = x.getAction(LabelAction.class);
                    // Works because ParallelLabelAction extends LabelAction implements ThreadNameAction
                    // And stages will have a StageAction too
                    if (label instanceof ThreadNameAction || x.getAction(StageAction.class) != null) {
                        return labelName.equals(label.getDisplayName());
                    }
                } else if (x instanceof BlockEndNode) {
                    FlowNode f = ((StepEndNode) x).getStartNode();
                    return this.apply(f);
                }
                return false;
            }
        };
        scan.setup(exec.getCurrentHeads());
        scan.filter(MATCH).forEachRemaining(x -> ((FlowNode)x).addAction(new NotExecutedNodeAction()));
    }
```
Yes, it will do an exhaustive scan (unfortunately) but I think it's somewhat unavoidable with this logic. 

Note: I haven't tested the above, so there may be some edge case I'm missing

svanoort: Note that only works with block-scoped stages -- IIRC we don't use any other kind with Declarative though, right?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -364,9 +367,13 @@ class ModelInterpreter implements Serializable {
                   Stage parentStage) {
         return {
             Utils.logToTaskListener(reason.message)
-            Utils.markStageWithTag(thisStage.name, StageStatus.TAG_NAME, reason.stageStatus, reason.isNotExecutedNode())
+            Utils.markStageWithTag(thisStage.name, StageStatus.TAG_NAME, reason.stageStatus)
```
svanoort: Any reason we couldn't do a bulk marking for both parallels and normal stages here, without checking for parallel content? Or maybe there's a subtlety I miss?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -320,12 +321,45 @@ class Utils {
...
+
+        FlowNode stage = scanner.findFirstMatch(execution.currentHeads, null, isStageWithOptionalName(stageName))
 
-                if (isRestart) {
-                    currentNode.addAction(new NotExecutedNodeAction())
+        if (stage != null && stage instanceof BlockStartNode) {
+            nodes.add(stage)
+
+            // Additional check needed to get the possible enclosing parallel branch for a nested stage.
+            Filterator<FlowNode> filtered = FlowScanningUtils.fetchEnclosingBlocks(stage)
```
svanoort: So my big uncertainty here is if we *do* need to limit to just the enclosing parallel branch or can do a more global search for matching branch names.  I *think* so but I haven't traced out all the legal combinations possible as of sequential stages (like we should check we can't get false positives from similarly named branches/stages that can occur elsewhere).

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -320,12 +321,45 @@ class Utils {
                     tagsAction.addTag(tagName, tagValue)
                     currentNode.save()
                 }
+            }
+        }
+    }
+
+    static void markStartAndEndNodesInStageAsNotExecuted(String stageName, FlowExecution execution = null) {
+        if (execution == null) {
+            CpsThread thread = CpsThread.current()
```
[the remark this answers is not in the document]
abayer: *shrug* I know, I know, but no other way to do about 80 things Declarative needs to do. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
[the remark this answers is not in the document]
abayer: Y'know, I'm not sure there is a valid reason why not. I'll play around with that on Friday.

abayer: Correct, no block-scoped. I'll try the `GraphListener` approach first, because that just does seem smarter if I can pull it off.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
abayer: I wanna make sure both the stage and the parallel branch it's in get marked correctly, 'cos I don't know for sure what `StatusAndTiming` will tell Blue Ocean otherwise. *shrug*

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: That is a possibility - stages are guaranteed to have unique names *in their context* but are not guaranteed to have unique names across other nested stage (parallel or sequential) contexts.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
svanoort: If you do GraphListener you should be able to use the predicate -- just replace the string equality with checking a `HashSet<String>` for the label -- note this relies on not being able to have collisions between executed and not executed names for any stage/parallel branch.

abayer:
@svanoort - whoops, I brainfarted and merged this because I saw the approved and forgot that there was other stuff. I'll cook up a followup.

PR #283, opened 2018-08-09
[FIXED JENKINS-51227] Add quietPeriod DeclarativeOption

abayer:
* JENKINS issue(s):
    * [JENKINS-51227](https://issues.jenkins-ci.org/browse/JENKINS-51227)
* Description:
    * This will let you set the quiet period for the job from the options block. As with job properties, triggers, and parameters, the quiet period will only be set to the default of 0 if the previous build actually had the QuietPeriod DeclarativeOption specified and the current build does not, so that UI-configured quiet periods don't get nuked.
* Documentation changes:
    * Will be needed.
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -39,6 +39,7 @@ import hudson.model.ParameterDefinition
 import hudson.model.ParametersDefinitionProperty
 import hudson.model.Result
 import hudson.triggers.Trigger
+import jenkins.model.Jenkins
 import org.apache.commons.codec.digest.DigestUtils
 import org.codehaus.groovy.ast.ASTNode
 import org.codehaus.groovy.ast.stmt.BlockStatement
@@ -58,6 +59,7 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.causes.RestartDeclarativeP
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.Environment
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.StepsBlock
 import org.jenkinsci.plugins.pipeline.modeldefinition.options.DeclarativeOption
+import org.jenkinsci.plugins.pipeline.modeldefinition.options.impl.QuietPeriod
 import org.jenkinsci.plugins.pipeline.modeldefinition.steps.CredentialWrapper
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted
 import org.jenkinsci.plugins.structs.SymbolLookup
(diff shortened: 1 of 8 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -610,6 +627,19 @@ class Utils {
...
+            // Check if QuietPeriod option is specified
+            QuietPeriod quietPeriod = (QuietPeriod) rawOptions.find { it instanceof QuietPeriod }
+            if (quietPeriod != null) {
+                j.setQuietPeriod(quietPeriod.quietPeriod)
+            } else {
+                String quietPeriodName = Jenkins.getActiveInstance().getDescriptorByType(QuietPeriod.DescriptorImpl.class)?.getName()
+                // If the quiet period was set by the previous build, wipe it out.
+                System.err.println("options: ${previousOptions}")
+                if (quietPeriodName != null && previousOptions.contains(quietPeriodName)) {
+                    j.setQuietPeriod(0)
```
dwnusbaum: Unless declarative previously set the quiet period to 0 in some other way, I think you should use `Jenkins.$GETTER.getQuietPeriod()` here to use the default quiet period in Jenkins (normally 5) instead. (WorkflowJob#getQuietPeriod does that as well if quietPeriod is null)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -610,6 +627,19 @@ class Utils {
...
         BulkChange bc = new BulkChange(j)
         try {
+            // Check if QuietPeriod option is specified
+            QuietPeriod quietPeriod = (QuietPeriod) rawOptions.find { it instanceof QuietPeriod }
+            if (quietPeriod != null) {
+                j.setQuietPeriod(quietPeriod.quietPeriod)
+            } else {
+                String quietPeriodName = Jenkins.getActiveInstance().getDescriptorByType(QuietPeriod.DescriptorImpl.class)?.getName()
+                // If the quiet period was set by the previous build, wipe it out.
+                System.err.println("options: ${previousOptions}")
```
dwnusbaum: Is this visible to users, or is it just for debugging? (Just curious)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: Good catch!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: That...is me missing a debug println.

jglick:
This was the wrong design. Should have been done as a `JobProperty` in `workflow-job`.

PR #285, opened 2018-08-23
[FIXED JENKINS-52114] Make success/failure use parallel stage result

abayer:
* JENKINS issue(s):
    * [JENKINS-52114](https://issues.jenkins-ci.org/browse/JENKINS-52114)
* Description:
    * This changes `success` and `failure` `post` conditions for stages to look at whether there's an explicit error passed to them, rather than just looking at build status. This doesn't apply for top-level `post`, and has some special logic to not fire `success` if the build status is anything but `FAILURE`, and to not fire `failure` if the error is a result of aborting the build.
    * There is one potentially problematic behavior change as a result - if you set `currentBuild.result` to `FAILURE` explicitly within a stage, `success` will still fire, and `failure` will not fire. There's no way to tell that this particular stage explicitly set `currentBuild.result` vs another stage running in parallel having an actual error.
    * Worth mentioning that this doesn't involve any change at all to any other conditions' behaviors - `unstable` still needs [JENKINS-39203](https://issues.jenkins-ci.org/browse/JENKINS-39203), `changed/fixed/regression` all apply to the whole build's status (at least until JENKINS-39203 and friends, at which point we could *theoretically* compare statuses for individual stages across builds, but probably won't), and with `aborted`, we're just treating it as meaning the whole build is aborted.
* Documentation changes:
    * ~~n/a - this just makes `failure` and `success` behave the way people seem to have expected. Shame we can't do anything about `unstable` here, though.~~
    * On second or third thought, yes, we will need to update the documentation to call out that `failure` and `success` on a `stage` are looking at that specific `stage`'s "result" rather than the whole build's status (unlike `unstable` and `aborted`), and that manually setting `currentBuild.result = 'FAILURE'` in a `stage` will no longer result in `failure` for that `stage` firing.
    * PR: https://github.com/jenkins-infra/jenkins.io/pull/1756
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/test/resources/postStage/localAll.groovy
+++ b/pipeline-model-definition/src/test/resources/postStage/localAll.groovy
@@ -35,6 +35,11 @@ pipeline {
                     if (res != null) {
                         echo "Setting build result ${res}"
                         currentBuild.result = res
+                        // JENKINS-52114 - can't tell the difference between setting currentBuild.result in this stage
+                        // and an error in a parallel stage, so let's error.
+                        if (res == "FAILURE") {
+                            error "Failing explicitly"
+                        }
                     } else {
                         echo "All is well"
                     }
--- a/pipeline-model-definition/src/test/resources/postStage/postFailureSuccessInParallel.groovy
+++ b/pipeline-model-definition/src/test/resources/postStage/postFailureSuccessInParallel.groovy
@@ -103,6 +103,12 @@ pipeline {
                 always {
                     echo 'Post ran'
                 }
+                success {
+                    echo "Parallel parent success"
+                }
+                failure {
+                    echo "Parallel parent failure"
+                }
             }
         }
     }
--- a/pipeline-model-definition/src/test/resources/postStage/postFailureSuccessInParallel.groovy
+++ b/pipeline-model-definition/src/test/resources/postStage/postFailureSuccessInParallel.groovy
@@ -1,0 +1,109 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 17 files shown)


pipeline-model-definition/src/test/resources/postStage/localAll.groovy
```diff
@@ -35,6 +35,11 @@ pipeline {
                     if (res != null) {
                         echo "Setting build result ${res}"
                         currentBuild.result = res
+                        // JENKINS-52114 - can't tell the difference between setting currentBuild.result in this stage
+                        // and an error in a parallel stage, so let's error.
+                        if (res == "FAILURE") {
+                            error "Failing explicitly"
```
dwnusbaum: Is this necessary to fix a failing test? If so, would this PR break existing Pipelines that are doing something similar by changing which post conditions are evaluated, or would it do something else?

dwnusbaum: I don't really have an informed opinion either way, it just seems a little scary to intentionally change behavior that we were testing against. It would be nice if we had telemetry to give us an idea of how often users rely on the previous behavior.

abayer: Oh, yeah, I'm wary of it. But I don't think it's hugely common - you can only set the build to `FAILURE` directly from within a `script` block (or I guess via some step that sets the build status to `FAILURE` without the step itself failing - the only examples I can think of for that would maybe be static analysis/code coverage publishers under certain configurations), and then you have to have `failure` in `post` for that stage.

This does merit further discussion/thought, for sure.

pipeline-model-definition/src/test/resources/postStage/postFailureSuccessInParallel.groovy
```diff
@@ -0,0 +1,109 @@
...
+                        success {
+                            echo "Found success in Child 3"
+                        }
+                        failure {
+                            echo "Found failure in Child 3"
+                        }
+                    }
+                }
+            }
+            post {
```
kshultzCB: Would there be any value in having `success` and `failure` here too? With `failure` in the `logContains` part of the test, and `success` in the `logNotContains` part of the test.

kshultzCB: To verify that the top-most stuff in`post` still behaves as expected.

abayer: Couldn't hurt - in practice, that's actually covered by a pile o' other tests, but hey, nothing wrong with testing it here too.

kshultzCB: :+1: . My thinking was, "make sure we still process these `success` and `failure` blocks after we've failed inside of the `parallel`. I'm sure that's being done elsewhere, but for cases like directly setting `currentBuild.result = 'FAILURE'` , I wasn't certain.

abayer:
@rsandell @kshultzCB @dwnusbaum @bitwiseman - I'm looking to do the Declarative 1.3.2 release in the next few days, and I'd like to get this in, so reviews/thoughts/concerns would be appreciated. Thanks!

abayer:
@dwnusbaum - yeah, my final thinking is that this being an issue in practice would require a combination of a number of uncommon-to-rare patterns in use together, most notably manually setting `currentBuild.result = 'FAILURE'` and having a `stage` `post` `failure`. Plus, if it turns out to be a more widely used behavior than we believe, we can always revert/rework/etc. =)

abayer:
@kshultzCB - yup, PR will be up for the docs before I merge this.

abayer:
Doc PR up at https://github.com/jenkins-infra/jenkins.io/pull/1756

PR #286, opened 2018-08-29
[FIXED JENKINS-53316] Switch to using jackson2-api dependency

abayer:
* JENKINS issue(s):
    * [JENKINS-53316](https://issues.jenkins-ci.org/browse/JENKINS-53316)
* Description:
    * And get rid of pipeline-model-json-shaded while we're here.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...


abayer:
I'm *fairly* sure this isn't going to break anything else - nobody else uses `json-schema-validator` so far as I can tell, so there shouldn't be a need to shade it.

pipeline-model-api/pom.xml
```diff
@@ -40,19 +40,18 @@
...
-          <artifactId>json-schema-validator</artifactId>
-        </exclusion>
-      </exclusions>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>jackson2-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-json-org</artifactId>
+      <version>2.8.11</version>
```
dwnusbaum: Can you omit the version to get it automatically from the parent's `dependencyManagement`?

pom.xml
```diff
@@ -300,6 +299,16 @@
         <artifactId>annotation-indexer</artifactId>
         <version>1.12</version>
       </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
+        <artifactId>jackson-datatype-json-org</artifactId>
+        <version>2.8.11</version>
```
dwnusbaum: Depending on this separately from jackson2-api could cause problems if the API plugin is updated and is no longer compatible with 2.8.11, but I don't think we have any great option since bundling that lib in jackson2-api would add an `org.apache.geronimo.bundles:json` dependency to all users, which also seems undesirable.

pipeline-model-api/pom.xml
abayer: Whoops, forgot to do that. Fixing.

pom.xml
abayer: Yeah, pretty much my thinking.

kshultzCB:
The shading stuff originally went in for [JENKINS-41911](https://issues.jenkins-ci.org/browse/JENKINS-41911), right?

dwnusbaum:
@abayer Based on your old comment in JENKINS-41911, does using json-schema-validator:2.0.4 cause issues with the newer version of Jackson?

> And I just remembered why we're on the 2.0.4 release of json-schema-validator (which uses old 2.2.x jackson): because the newer train of json-schema-validator (2.2.x, which uses 2.7.x jackson) also needs Guava 16.

dwnusbaum:
@abayer Ok, I checked `mvn dependency:tree` and there don't appear to be any weird Jackson/Guava versions being pulled in, so if the tests work then l think the change should be good.

PR #289, opened 2018-09-25
Trying to de-flake two tests

abayer:
These suckers are flaky for reasons I just don't understand - the
build result check will come up as failure, even though the build is
definitely aborted. I'm suspecting a timing issue, so am trying a sleep.

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BuildConditionResponderTest.java
@@ -296,7 +296,11 @@ public class BuildConditionResponderTest extends AbstractModelDefTest {
         Thread.sleep(1000);
         run1.doStop();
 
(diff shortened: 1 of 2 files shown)


abayer:
I can reproduce it like once in a couple dozen runs at best, so no, no debugger luck. And this hack didn't help either - I think I'm actually gonna just get rid of the result check, since for the purposes of this test I honestly don't care what the build status is *after* the build completes, just while it's going on.

abayer:
Worth mentioning that there are other cases of flakiness that pop up now and again, mainly with tests timing out very rarely with no pattern to them, but these two tests are the only ones that flake out with anything resembling consistency.

abayer:
I've now run the full test suite 15 times on a beefy test box without these two tests barfing, but I'm gonna let that keep running overnight to be sure.

abayer:
Narrowed down test scope even farther (so that other flaky tests didn't cause false negatives), and had a ~15% flake rate on `master` for the two tests in question, while I've got over 150 runs of each of the tests against this branch with no failures at all. So I'm gonna merge. =)

PR #291, opened 2018-10-15
[FIXED JENKINS-53804] Call post conditions as early as possible

abayer:
* JENKINS issue(s):
    * [JENKINS-53804](https://issues.jenkins-ci.org/browse/JENKINS-53804)
* Description:
    * Specifically, try to catch errors from directives and execute the post conditions for the build/stage right afterwards, while we're still inside any enclosing directives - i.e., still in top-level `agent` when a top-level `credentials` explodes.
    * Also reworked post handling in general for stages when I realized that an error in a directive before execution of a stage's steps or nested stages would actually result in `post` for the stage not being run at all, which is not exactly ideal.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -73,26 +73,26 @@ class ModelInterpreter implements Serializable {
 
                 // Entire build, including notifications, runs in the agent.
                 inDeclarativeAgent(root, root, root.agent) {
-                    withCredentialsBlock(root.environment) {
-                        withEnvBlock(root.getEnvVars(script)) {
-                            inWrappers(root.options?.wrappers) {
-                                toolsBlock(root.tools, root.agent, null) {
-                                    firstError = evaluateSequentialStages(root, root.stages, firstError, null, restartedStage, null).call()
+                    firstError = runDirectiveWithPostBuild(root, firstError) {
+                        withCredentialsBlock(root.environment) {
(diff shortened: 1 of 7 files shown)


abayer:
Note that I've put this in `work-in-progress` for the moment, because at a minimum I need to add javadoc where I've left `TODO`s, and in general...I'm just not sure yet about this one.

abayer:
...and there are a pile o' test failures, suggesting I screwed something else up. =)

abayer:
Yup, something's wrong with how stage errors are logged, at a minimum. I'll pick this up either tomorrow or after Jenkins World Nice next week.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -122,6 +117,63 @@ class ModelInterpreter implements Serializable {
...
+     * TODO
+     */
+    void runDirectiveWithPostBuild(Root root, Throwable firstErrorSeen, Closure body) {
+        try {
+            body.call()
+        } catch (Throwable e) {
+            if (firstErrorSeen == null) {
+                firstErrorSeen = e
+            }
+            // TODO: Ignoring errors in post because we already errored in a directive anyway so this would just mask things.
```
dwnusbaum: Would it be useful to track these exceptions as suppressed exceptions rather than just dropping them, or are exceptions after the first one generally useless to the end user? I am imagining something like:

    if (firstErrorSeen == null) {
        firstErrorSeen = e
    } else {
        firstErrorSeen.addSuppressed(e)
    }
    try { 
        executePostBuild(root)
    } catch (Throwable e2) {
        firstErrorSeen.addSuppressed(e2)
    }

PR #292, opened 2018-10-16
[FIXED JENKINS-53734] Allow parallel within top-level sequential

abayer:
* JENKINS issue(s):
    * [JENKINS-53734](https://issues.jenkins-ci.org/browse/JENKINS-53734)
* Description:
    * Note that we're not expecting any visualization change for this - Blue Ocean does not have the ability to visualize nested parallels, nor do we expect it to any time soon, so this'll just show up there as a sequential stage.
    * We'll still reject any attempt to use `parallel` within another `parallel`, but something like the below would now be allowed.
```
stages {
  stage("foo") {
    stages {
      stage("bar") {
        parallel {
          stage("a") {
            ...
          }
          stage("b") {
            ...
          }
        }
      }
    }
  }
}
```
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1848
* Users/aliases to notify:
    * ...


abayer:
@dwnusbaum - yup, we do - https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/adb3ff4acd2f09ea5d297aee45a9efed426ca3c9/pipeline-model-definition/src/test/resources/errors/parallelStagesDeepNesting.groovy for example.

jjj11x:
>   * We'll still reject any attempt to use `parallel` within another `parallel`

What's the rationale for rejecting any attempt to use `parallel` within another `parallel`? Is it a technical limitation (e.g. something not implemented) or a philosophical limitation (e.g. would be potentially too complicated for users)? It would be convenient to have both stages and parallel nested within another parallel.

bitwiseman:
@jjj11x 
Commenting on extremely old PRs is generally unlikely to get  any response. I just happened it see this roll by on my notifications. 

I believe the reasoning was related to visualization, I don't believe there is any direct technical limitation.  The underlying system supports nesting parallels.

PR #293, opened 2018-10-16
[FIXED JENKINS-53662] Add isRestartedStage() when condition

abayer:
* JENKINS issue(s):
    * [JENKINS-53662](https://issues.jenkins-ci.org/browse/JENKINS-53662)
* Description:
    * This will only fire when the run was restarted from a stage, and we're currently checking whether to execute the stage with the same name as the restarted stage.
* Documentation changes:
    * The new condition will need to be added to the docs.
* Users/aliases to notify:
    * ...


abayer:
Yeah, I'm probably going to table this and work on a more general way of providing information on the restarted build/stage as a whole, rather than just one part at a time.

abayer:
Moving this to WIP to make sure I don't inadvertently merge it before I am sure of the right approach.

bitwiseman:
@abayer  Any chance you can resolve the merge conflicts here?  Should this remain open?

PR #297, opened 2018-11-15
Jenkins 50880

joseblas:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-50880
* Description:
    * Adding beforeInput flag in StageWhen
* Documentation changes:
    * Not yet done

### DIFF
--- a/pipeline-model-definition/src/test/resources/json/whenBeforeInputTrue.json
+++ b/pipeline-model-definition/src/test/resources/json/whenBeforeInputTrue.json
@@ -1,0 +1,56 @@
+{"pipeline": {
+  "stages":   [
+    {
+      "name": "One",
+      "branches": [      {
+        "name": "default",
+        "steps": [        {
+          "name": "echo",
+          "arguments": [          {
+            "key": "message",
+            "value":             {
+              "isLiteral": true,
+              "value": "Hello"
+            }
+          }]
+        }]
+      }]
+    },
+    {
+      "name": "Two",
+      "branches": [      {
+        "name": "default",
+        "steps": [        {
+          "name": "script",
+          "arguments": [          {
+            "key": "scriptBlock",
+            "value":             {
+              "isLiteral": true,
+              "value": "echo \"World\"\n                    echo \"Heal it\""
+            }
+          }]
+        }]
+      }],
+      "when": {"conditions": [      {
+        "name": "branch",
+        "arguments":         {
+          "isLiteral": true,
+          "value": "SOME_OTHER_BRANCH"
+        },
+        "name": "beforeInput",
+        "arguments":         {
+          "isLiteral": true,
+          "value": "true"
+        }
+      }]}
+    }
+  ],
+  "environment": [  {
+    "key": "BRANCH_NAME",
+    "value":     {
+      "isLiteral": true,
+      "value": "master"
+    }
+  }],
+  "agent": {"type": "any"}
+}}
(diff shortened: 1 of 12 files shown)


abayer:
Not sure why the build blew up so comprehensively - I've kicked off a new run and am also building locally.

abayer:
Ah, the tests are real failures. Could you take a look at `SuccessfulEndpointOpsTest` and `ConvertRoundTripTest`, @joseblas?

abayer:
And I've pushed a fix to your branch. =)

abayer:
The `ExecuteConvertedTest` for `whenBeforeInputTrue` is hanging on Groovy->AST->Groovy execution, but passing fine on JSON->AST->Groovy execution. Something's awry there.

abayer:
aaaaaaah - `whenBeforeInputTrue.json` and `whenBeforeInputTrue.groovy` aren't actually the same thing. And actually, `whenBeforeInputTrue` and `whenBeforeInputFalse` are a little confusingly named - they're actually testing "what if the `when` condition resolves to true" and "what if the `when` condition resolves to false", not "what if `beforeInput` is true" and "what if `beforeInput` is false". I'd suggest renaming the tests and related files to something clearer.

And then, I'd say use `whenBeforeInputFalse` (or rather, its renamed equivalent) as the test case for the conversion stuff - since it won't execute the `input` and so `ExecuteConvertedTest` won't hang forever waiting for input.

You can get what the JSON should look like by running `mvn test -Dtest=WhenStageTest#whenBeforeInputFalse -Dorg.jenkinsci.plugins.pipeline.modeldefinition.parser.ModelParser.astDebugLogging=true`. This will print out a couple things to STDERR - the JSON that the Jenkinsfile converts to, and the Groovy AST for the runtime conversion into the `Root` etc objects. You can copy the JSON it prints out into `whenBeforeInputFalse.json` and know it'll be right.

pipeline-model-definition/src/test/resources/json/whenBeforeInputTrue.json
```diff
@@ -1,54 +1,61 @@
...
+  "stages": [  {
+    "name": "One",
+    "when":     {
+      "conditions": [      {
+        "name": "expression",
+        "arguments": [        {
+          "key": "scriptBlock",
+          "value":           {
+            "isLiteral": true,
+            "value": "false"
```
abayer: So this file should be renamed to `whenBeforeInputFalse.json` and `AbstractModelDefTest`'s list of test cases should be changed accordingly - since this is the JSON form of `whenBeforeInputFalse.groovy`, not `whenBeforeInputTrue.groovy`.

joseblas: This JSON has been changed, now is really beforeInputTrue

Jenkinsfile
```diff
@@ -13,7 +13,7 @@ pipeline {
     options {
         buildDiscarder(logRotator(numToKeepStr:'10'))
         timestamps()
-        timeout(time: 90, unit: 'MINUTES')
+        timeout(time: 180, unit: 'MINUTES')
```
abayer: This timeout change is not needed - the underlying problem here remains that `whenBeforeInputTrue.json` is still actually the JSON equivalent of `whenBeforeInputFalse.groovy`, meaning that tests are failing and/or hanging due to the mismatch between `whenBeforeInputTrue.json` and `whenBeforeInputTrue.groovy`.

pipeline-model-definition/src/test/resources/json/whenBeforeInputTrue.json
abayer: No, this is still wrong - the `when` condition here is returning false, as in `whenBeforeInputFalse.groovy`. As I said above, this file should be renamed to `whenBeforeInputFalse.json` and `whenBeforeInputFalse` should be used in `AbstractModelDefTest`, since `whenBeforeInputFalse` will not hang in the execution of converted tests, while `whenBeforeInputTrue` will hang.

abayer:
@joseblas - could you please do the docs PR I mentioned? Thanks!

joseblas:
@abayer The PR is already done: https://github.com/jenkins-infra/jenkins.io/pull/1900

PR #298, opened 2018-11-16
Fix description of "Unstable"

reinholdfuereder:
Seen in "Declarative Directive Generator": grafik

1. Remove old (?) description for "Fixed" that accidentally was appended to "Unstable"
2. Since the two descriptions for "Fixed" vary, which one is actually the correct one? I guess/hope for the latter/current, as "Fixed" is hopefully also run when the previous build was e.g. "Aborted"?
    - Run if the current build is "Success" and the previous build is "Failure" or "Unstable"
    - Run if the previous build was not successful and the current build's status is "Success"


reinholdfuereder:
Please note that https://jenkins.io/doc/book/pipeline/syntax/ still documents it (more or less) like the first/former one:

> fixed:
>   Only run the steps in post if the current Pipeline’s or stage’s run is successful and the previous run failed or was unstable.

reinholdfuereder:
(I am pretty sure the failing test has nothing to do with my commit, but is a rather general flakyness of the test bench...)

PR #301, opened 2018-11-28
[FIXED JENKINS-54919] Process agent configuration on the fly

abayer:
* JENKINS issue(s):
    * [JENKINS-54919](https://issues.jenkins-ci.org/browse/JENKINS-54919)
* Description:
    * This lets us reference environment variables (well, sorta - you can reference an environment variable defined at the parent level in a stage agent, but not an environment variable defined at the same level) and more immediately, this will allow referencing parameters that have just been added to the Pipeline in this build (well, as long as they have default values set).
    * Basically, this makes Agent work more like StageConditionals by wrapping the map that defines the agent in a closure, and evaluating that closure at the time we actually need to use it, rather than baking the values in when we instantiate the runtime model.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
@@ -160,4 +160,6 @@ class Agent extends MappedClosure<Object,Agent> implements Serializable {
         DeclarativeAgent a = getDeclarativeAgent(null, null)
         return a != null && !None.class.isInstance(a)
     }
+
+    private static final long serialVersionUID = -9134086848416921688L
 }
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
@@ -49,10 +49,17 @@ import javax.annotation.CheckForNull
 @ToString
 @EqualsAndHashCode
 @SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
-class Agent implements Serializable {
+class Agent extends MappedClosure<Object,Agent> implements Serializable {
 
     final Closure rawClosure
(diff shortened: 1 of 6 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Agent.groovy
```diff
@@ -49,11 +49,18 @@ import javax.annotation.CheckForNull
 @ToString
 @EqualsAndHashCode
 @SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
-class Agent extends MappedClosure<Object,Agent> implements Serializable {
+class Agent implements Serializable {
```
dwnusbaum: I'm not quite sure how the serialization is used here, but do we need to worry about compatibility with the old version of the class? If so, there will be issues because of the class hierarchy change and implicit `serialVersionUID`.

abayer: Hmmm...the only way we could have issues would be with currently running builds at upgrade time. I can probably work around that, actually. Lemme see what I can do.

abayer: Think I got it. =)

dwnusbaum: I think that adding fields and constructors will change the computed `serialVersionUID` so there will still be problems (unless the serialization library being used ignores `serialVersionUID`). You'd need to check what the old computed value was and add its value explicitly here to avoid that issue.

dwnusbaum: And FWIW I ran the following in the Jenkins script console with Declarative 1.3.2 installed and it looks like the `serialVersionUID` of the class before the change is `-9134086848416921688`:

    println(ObjectStreamClass.lookup(org.jenkinsci.plugins.pipeline.modeldefinition.model.Agent.class).getSerialVersionUID())

PR #303, opened 2018-11-28
[JENKINS-51792] Change ModelParser to error only once when there are multiple steps blocks

kshultzCB:
* JENKINS issue(s):
    * [JENKINS-51792](https://issues.jenkins-ci.org/browse/JENKINS-51792)
* Description:
    * Given a malformed Pipeline, where there was more than one `steps` section inside of a `stage`, we were reporting a `Duplicate parallel branch name: "default"`. This was a bit confusing if the user's Pipeline had no branch named `default` in it at all. This change checks to make sure that the stage is empty before adding a `default` branch.
* Documentation changes:
    * No documentation changes should be necessary. This is just cleanup of some unnecessary error messages, which a user would only see if they had a mistake in their Pipeline anyway.
* Users/aliases to notify:
    * N/A

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -550,8 +550,12 @@ class ModelParser implements Parser {
                                     }
                                     stage.failFast = parallel.failFast
                                 } else {
-                                    // otherwise it's a single line of execution
-                                    stage.branches.add(parseBranch("default", block))
+                                    if (stage.getBranches().isEmpty()) {
+                                        // Only add a 'default' branch here if we don't already have one.
+                                        stage.branches.add(parseBranch("default", block))
+                                    } else {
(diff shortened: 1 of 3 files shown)


kshultzCB:
[Build 2](https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/PR-303/2/) timed out on CI. I don't have the privileges on ci.jenkins.io to relaunch it.

image

kshultzCB:
I'm doing the close/reopen dance to trigger another build. The  `windows-newer-core` branch timed out.

kshultzCB:
Here's the reopen part of the dance.

abayer:
@kshultzCB FYI, I just pushed a change to master to bump the job timeout to an hour from 90 minutes, since queue time can hose us.

abayer:
...I also might get rid of `ExecuteConvertedTest` soon - it eats a lot of time but doesn't provide much real value, with some small tweaks to `ConvertRoundTripTest` to make sure it's covering every angle.

abayer:
Re-running yet again to pick up the job timeout bump to 2 hours. sigh.

abayer:
Test failure is bogus, I'm just merging.

PR #304, opened 2018-11-29
Combine and remove redundant tests to speed up build time

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * This is just the first part - gonna keep tweaking. `ExecuteConvertedTest` is basically completely redundant - it only runs Jenkinsfiles we already run elsewhere, and we already do all the conversion in `ConvertRoundTripTest` too. I've never actually seen a failure in `ExecuteConvertedTest` without a corresponding failure elsewhere as well.
    * This initial change cut like 8% of my total test time in a local run.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...


abayer:
Oh, and that's like 8% of what it takes on Linux, which means actually running all the Docker tests - that ends up accounting for like ~1/6 of the total test run time on my desktop, so the percentage improvement on Windows (where Docker tests get skipped) should be more significant.

abayer:
Coverage is basically identical vs master - like three additional branches of 84k total missed. But I'll keep the original baseline coverage around to compare to and make sure we recover whatever we end up missing down the road.

PR #309, opened 2018-12-17
JENKINS-52744 Showing meaningful error when an environment variable i…

joseblas:
…s not declared properly

* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-52744
* Description:
    * Using not existing environment variables in environment GString expression causes strange error message
* Documentation changes:
    * Changing message error, no need for documentation update
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -454,6 +454,7 @@ class ModelInterpreter implements Serializable {
                 try{
                     "${k}=${v.call()}"
                 }catch (NullPointerException e){
+
                     throw new IllegalArgumentException(format("Invalid var declared in environment: %s", k))
                 }
             }.findAll { it != null}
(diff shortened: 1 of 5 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -450,8 +451,13 @@ class ModelInterpreter implements Serializable {
     def withEnvBlock(Map<String,Closure> envVars, Closure body) {
         if (envVars != null && !envVars.isEmpty()) {
             List<String> evaledEnv = envVars.collect { k, v ->
-                "${k}=${v.call()}"
-            }
+                try{
+                    "${k}=${v.call()}"
+                }catch (NullPointerException e){
+
+                    throw new IllegalArgumentException(format("Invalid var declared in environment: %s", k))
```
abayer: Maybe reword this to be more clear - is it that the environment variable `FOO` is invalid, or that something went wrong with its value?

joseblas: Changed to: 

> Invalid variable (FOO) declared in environment; There are some problems with its value

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -455,7 +455,7 @@ class ModelInterpreter implements Serializable {
                     "${k}=${v.call()}"
                 }catch (NullPointerException e){
 
-                    throw new IllegalArgumentException(format("Invalid var declared in environment: %s", k))
+                    throw new IllegalArgumentException(format("Invalid variable (%s) declared in environment; There are some problems with its value", k))
```
abayer: So in practice, this is only going to show up if there's a reference to a non-existent variable in the value, right? Hence the `NullPointerException` - so in that case, we could probably be more specific and say something like `One or more variables in its value do not exist`?

Also, this should go in `Messages.properties` so that it can be localized.

PR #311, opened 2019-01-29
[JENKINS-56294] pass commandLine to dockerFingerprintFrom for dockerfile agent

dwightguth:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-56294
* Description:
    * the dockerFingerprintFrom step uses an optional parameter commandLine which it parses for the build args to be passed to docker and in turn requires these arguments in order to correctly determine the from image for dockerfiles where the FROM directive contains a build argument in its expansion. This manifests as an IOException thrown by the dockerFingerprintFrom step when used within a declarative pipeline with a dockerfile whose FROM directive contains a build argument. I believe passing this commandLine argument should resolve that issue and restore parity with the docker workflow plugin.
* Documentation changes:
    * This is a pure bug fix and should not require any new user documentation
* Users/aliases to notify:
    * ...


dwightguth:
I just realized this may need to wait for the upcoming docker-workflow-plugin version 1.18 for it to actually affect anything, but it doesn't seem to break anything as it currently stands.

dwightguth:
Am I missing something or did CI time out because it had been waiting for an executor for over 2 hours? That seems really strange to me. What are my next steps?

dwightguth:
@abayer are you just waiting for the docker-workflow-plugin to be updated or is there something else I need to do to get this reviewed?

dwightguth:
@abayer docker-workflow-plugin 1.18 has been released, so all that remains to resolve this issue is to merge this PR. Can I get a review please?

abayer:
Sorry for the delay!

PR #312, opened 2019-02-11
Fix errors in documentation

basil:
#45 moved `ast-schema.json` from `pipeline-model-definition` to `pipeline-model-api`; however, the documentation was never updated to refer to the new location. This change updates the documentation to refer to the new location. I also fixed a few other typos I noticed in the documentation while I was there.


abayer:
Thanks!

PR #300, opened 2018-11-28
JENKINS-49997 - Added Unsuccessful build condition for _post_

joseblas:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-49997
* Description:
    * Pipeline Declarative post unsuccessful block 
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/1909
* Users/aliases to notify:
    * N/A

PR #313, opened 2019-02-20
[JENKINS-55459] Make parallel stages that fail fast result in build failure

dwnusbaum:
* JENKINS issue(s):
    * [JENKINS-55459](https://issues.jenkins-ci.org/browse/JENKINS-55459)
    * [JENKINS-55476](https://issues.jenkins-ci.org/browse/JENKINS-55476)
* Description:
    * This changes the behavior of Declarative Pipeline to mimic Scripted Pipeline when a parallel stage configured to fail fast, fails fast. Previously, the overall build result would be set to ABORTED, whereas in Scripted Pipeline (and Declarative after this change), the overall build result would be set to FAILURE.
    * This PR also fixes a bug causing the `unsuccessful` condition to always be executed in post blocks (JENKINS-55476).
    * Open questions:
      * How does this affect Blue Ocean/Stage View visualizations?
      * Does removing the setting of build results in `ModelInterpreter.groovy` break post conditions in some cases?
      * Does the trick in `Aborted.groovy` (which fixes regressions in tests due to the above changes) have side effects that are undesirable? Should similar behavior be added to any of the other build conditions?
      * The build result is still set explicitly in `ModelInterpreter#runPostConditions` and `ModelInterpreter#evaluateStage`, and removing that code caused many test failures, so I left them, but will their existence break this fix in some cases?
* Documentation changes:
    * Not sure if the current behavior is documented anywhere, would need to be investigated.
* Users/aliases to notify:
    * @nofarb

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.pipeline.modeldefinition.model.conditions
 import hudson.Extension
 import hudson.model.Result
 import org.jenkinsci.Symbol
+import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
 import org.jenkinsci.plugins.workflow.job.WorkflowRun
 
@@ -47,7 +48,11 @@ class Aborted extends BuildCondition {
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
         Result execResult = getExecutionResult(r)
-        return execResult == Result.ABORTED || r.getResult() == Result.ABORTED
+        Result errorResult = Result.SUCCESS;
+        if (error != null) {
+            errorResult = Utils.getResultFromException(error)
+        }
+        return execResult == Result.ABORTED || r.getResult() == Result.ABORTED || errorResult == Result.ABORTED
     }
 
     @Override
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
@@ -48,7 +48,7 @@ class Aborted extends BuildCondition {
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
         Result execResult = getExecutionResult(r)
-        Result errorResult = Result.SUCCESS;
+        Result errorResult = null
         if (error != null) {
             errorResult = Utils.getResultFromException(error)
         }
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Unsuccessful.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Unsuccessful.groovy
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.pipeline.modeldefinition.model.conditions
 import hudson.Extension
 import hudson.model.Result
 import org.jenkinsci.Symbol
+import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
 import org.jenkinsci.plugins.workflow.job.WorkflowRun
 
@@ -46,16 +47,14 @@ class Unsuccessful extends BuildCondition {
 
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
-        return isNotNull(r) || ! is(r, Result.SUCCESS)
-    }
-
-    private boolean isNotNull(WorkflowRun r){
-        return r.getResult() != null && getExecutionResult(r) != null
-    }
-
-    private boolean is(WorkflowRun r, Result res){
         Result execResult = getExecutionResult(r)
-        return execResult == res || r.getResult() == res
+        Result errorResult = null
+        if (error != null) {
+            errorResult = Utils.getResultFromException(error)
+        }
+        return (execResult != null && execResult != Result.SUCCESS) ||
+                (r.getResult() != null && r.getResult() != Result.SUCCESS) ||
+                (errorResult != null && errorResult != Result.SUCCESS)
     }
 
     @Override
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -168,7 +168,6 @@ class ModelInterpreter implements Serializable {
                     try {
                         evaluateStage(root, thisStage.agent ?: root.agent, thisStage, firstError, parent, skippedReason).call()
                     } catch (Throwable e) {
-                        script.getProperty("currentBuild").result = Utils.getResultFromException(e)
                         Utils.markStageFailedAndContinued(thisStage.name)
                         if (firstError == null) {
                             firstError = e
@@ -684,7 +683,6 @@ class ModelInterpreter implements Serializable {
                 delegateAndExecute(thisStage.steps.closure)
             }
         } catch (Throwable e) {
-            script.getProperty("currentBuild").result = Utils.getResultFromException(e)
             Utils.markStageFailedAndContinued(thisStage.name)
             if (stageError == null) {
                 stageError = e
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -83,7 +83,7 @@ class ModelInterpreter implements Serializable {
                                     // Execute post-build actions now that we've finished all parallel.
                                     try {
                                         postBuildRun = true
-                                        executePostBuild(root)
+                                        executePostBuild(root, firstError)
                                     } catch (Throwable e) {
                                         if (firstError == null) {
                                             firstError = e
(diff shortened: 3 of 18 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -1384,13 +1384,13 @@ public void classInJenkinsfile() throws Exception {
     @Issue("JENKINS-47109")
     @Test
     public void parallelStagesFailFast() throws Exception {
-        expect(Result.ABORTED, "parallelStagesFailFast")
+        expect(Result.FAILURE, "parallelStagesFailFast")
```
dwnusbaum: I haven't added any new tests since the modifications needed here and in `parallelStagesFailFastWithOption` are the behavior we are testing. Might be good to add some some more complicated tests involving post conditions and nested stages. Should also add another `@Issue` for the current ticket.

abayer:
Looks good to me so far!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Aborted.groovy
```diff
@@ -47,7 +48,11 @@ class Aborted extends BuildCondition {
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
         Result execResult = getExecutionResult(r)
-        return execResult == Result.ABORTED || r.getResult() == Result.ABORTED
+        Result errorResult = Result.SUCCESS;
+        if (error != null) {
+            errorResult = Utils.getResultFromException(error)
+        }
+        return execResult == Result.ABORTED || r.getResult() == Result.ABORTED || errorResult == Result.ABORTED
```
jglick: My head hurts.

dwnusbaum:
`BuildConditionResponderTest#notBuiltFlowInterruptedException` failed on the Windows branch, but passed on the other branches, which is a little concerning. [See the test results for BuildConditionResponderTest](https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/PR-313/1/testReport/org.jenkinsci.plugins.pipeline.modeldefinition/BuildConditionResponderTest/) and search for `notBuiltFlowInterruptedException`.

abayer:
@dwnusbaum There's a chance that's a flaky test - can't remember for sure. I'm replaying it to check.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
kshultzCB: There's probably existing tests that could be similarly modified. Heck, there's probably a bunch of them that'll need tweaking (or even consolidation?) anyway.

abayer:
@dwnusbaum So the rerun did fail that test on both Windows variants it ran, so it is possible this made that test more flakey. I'm not sure how Windows specifically is making it fail, though - it could still just be bad luck.

abayer:
Ok, the test is failing consistently enough now that I am beginning to assume it's legit. Which is weird as hell, but...

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Unsuccessful.groovy
```diff
@@ -46,16 +47,14 @@ class Unsuccessful extends BuildCondition {
 
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
-        return isNotNull(r) || ! is(r, Result.SUCCESS)
```
dwnusbaum: #300 had a bug; this should have been `isNotNull(r) && ! is(r, Result.SUCCESS)`. The previous code made the condition run even if the build is successful. This was already reported as [JENKINS-55476](https://issues.jenkins-ci.org/browse/JENKINS-55476), and I've tweaked the tests so they'll catch any regressions here in the future.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -235,8 +234,6 @@ class ModelInterpreter implements Serializable {
     def evaluateStage(Root root, Agent parentAgent, Stage thisStage, Throwable firstError, Stage parent,
                       SkippedStageReason skippedReason) {
         return {
-            def thisError = null
```
dwnusbaum: This variable was only written to, never read.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/PostStageTest.java
```diff
@@ -111,6 +113,7 @@ public void withAllLocalUnsuccessfulWithNotBuilt() throws Exception {
         env(s).put("MAKE_RESULT", Result.NOT_BUILT.toString()).set();
         expect(Result.NOT_BUILT, "unsuccessful")
                 .logContains("I LOVE YOU VIRGINIA")
+                .logContains("I FAILED YOU, SORRY")
```
dwnusbaum: I added this to make it clear what the actual behavior is, but do we want the unsuccessful condition to fire if the result is `NOT_BUILT`?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/PostStageTest.java
```diff
@@ -98,11 +98,13 @@ public void withAllLocalUnsuccessfulWithFailure() throws Exception {
 
     }
 
+    @Issue("JENKINS-55476")
     @Test
     public void withAllLocalUnsuccessfulWithSuccess() throws Exception {
         env(s).put("MAKE_RESULT", Result.SUCCESS.toString()).set();
         expect(Result.SUCCESS, "unsuccessful")
                 .logContains("I LOVE YOU VIRGINIA")
+                .logNotContains("I FAILED YOU, SORRY")
```
dwnusbaum: This is the line that would've caught JENKINS-55476.

dwnusbaum:
CI build was good for 0f88740, but failed for 39c7ed1 because `BuildConditionResponderTest#postFailureAfterUnstable` failed on the newer cores. I have tried running that test locally with the newer core but have not yet seen a flake. Notably in my local runs, the tests ends with `Finished: FAILURE`, as expected, but the CI tests did not contain a `Finished: ` line at all (I would've expected to see `Finished: Unstable`). Also notable is that the even the CI logs for the failing run contained [this line](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/ce298b76d6e71ef694794fc5d6b7719a9c191894/pipeline-model-definition/src/test/resources/postFailureAfterUnstable.groovy#L42), so the post condition did fire, the build result just didn't get updated from Unstable to Failure.

dwnusbaum:
I think that the `Finished: foo` line is printed by [BuildListener.java](https://github.com/jenkinsci/jenkins/blob/22aa2e6e766074d11249893e3f35e0b99e20d3d0/core/src/main/java/hudson/model/BuildListener.java), so perhaps that is just a race condition with the listener and `JenkinsRule#waitForCompletion` (used [here](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/ce298b76d6e71ef694794fc5d6b7719a9c191894/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java#L488)). 

Perhaps the issue is that since `JenkinsRule#waitForCompletion` checks `Run#isBuilding()` in a loop, and `WorkflowRun#isBuilding` is implemented as [`return result == null || isInProgress()`](https://github.com/jenkinsci/workflow-job-plugin/blob/3e5c4995eecc51479044a9fbfb50a20eb550687d/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L741), the fact that the failing tests explicitly set `currentBuild.result` in their body [here](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/ce298b76d6e71ef694794fc5d6b7719a9c191894/pipeline-model-definition/src/test/resources/postFailureAfterUnstable.groovy#L32) means that `JenkinsRule#waitForCompletion` is completing too early and the build is not actually complete. 

I'm thinking about a way to modify `AbstractModelDefTest` to make it wait for actual completion of the build for now, but maybe it would be better to change the implementation of `WorkflowRun#isBuilding` since having the build report as completed according to that method when the build result is set explicitly seems like it could be a source of other errors.

dwnusbaum:
Triggering a new build just to see how many failures we get.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
```diff
@@ -485,7 +485,35 @@ public WorkflowRun go() throws Exception {
...
+                // Here is an extract of logs where this happened while running BuildConditionResponderTest.postFailureAfterUnstable(),
+                // note that WorkflowRun#finish was called _after_ the build result was checked for the first time!
+                /*
+                  14.530 [test0 #1] [Pipeline] echo
+                  14.531 [test0 #1] I FAILED
+                  14.531 [test0 #1] [Pipeline] }
+                Build result did not initially match FAILURE
+                  14.583 [test0 #1] [Pipeline] // stage
+                  14.583 [test0 #1] [Pipeline] End of Pipeline
+                  14.589 [id=52]    INFO    o.j.p.workflow.job.WorkflowRun#finish: test0 #1 completed: FAILURE
```
dwnusbaum: I think I was on the right track with https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/313#issuecomment-467040926, but was thinking about the wrong half of the method. When we have a manually-set build result, `WorkflowRun.isBuilding()` calls `WorkflowRun.isInProgress()`, and I think that method could return `false` before the final build result has been set in some cases (a somewhat subtle race condition).

I'm not sure if this would be an issue in real-world use cases, since I think would only affect someone polling `WorkflowRun.isBuilding()` in a tight loop to wait for a build to complete, but it would be nice to fix this somehow in workflow-job.

dwnusbaum:
Ok, it looks like the workaround did trigger for the tests on a few branches (see [the linux branch](https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/PR-313/11/testReport/org.jenkinsci.plugins.pipeline.modeldefinition/BuildConditionResponderTest/build___linux___postFailureAfterUnstable/) for example), so I think that was the cause of the flakes:

```
  37.981 [test11 #1] [Pipeline] echo
  38.032 [test11 #1] I FAILED
  38.082 [test11 #1] [Pipeline] }
  38.132 [test11 #1] [Pipeline] // stage
  38.248 [test11 #1] [Pipeline] End of Pipeline
Build result did not initially match FAILURE
  38.285 [id=161]	INFO	o.j.p.workflow.job.WorkflowRun#finish: test11 #1 completed: FAILURE
  38.288 [test11 #1] ERROR: I AM FAILING NOW
  38.288 [test11 #1] Finished: FAILURE
But after 1s build result _did_ match FAILURE
```

abayer:
yay!

PR #314, opened 2019-02-25
Bumb version of jackson-api-plugin to 2.9.8.

uhafner:
2.8 versions contain a broken JSON library that fails warnings-ng tests:
https://travis-ci.org/jenkinsci/warnings-ng-plugin/jobs/498444524#L7229


uhafner:
After removing dependency to pipeline-model everything is ok again:
https://travis-ci.org/jenkinsci/warnings-ng-plugin/builds/498694580

PR #319, opened 2019-03-08
[JENKINS-56402] Set build result to the logical result in post conditions

dwnusbaum:
* JENKINS issue(s):
    * See [JENKINS-56402](https://issues.jenkins-ci.org/browse/JENKINS-56402)
* Description:
    * Set the build result in post conditions to the logical build result that will be used to determine which post blocks run. The result is unset after the post block so that JENKINS-55459 doesn't regress. If this seems too risky or doesn't seem like it would fix all of the reported issues, then we can go with #320 which just reverts #313 and then re-applies the patch for JENKINS-55475.
* Documentation changes:
    * N/A
* Users/aliases to notify:
    * N/A

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -779,10 +779,15 @@ class ModelInterpreter implements Serializable {
                           String stageName = null,
                           Object context = null) {
         BuildCondition.orderedConditionNames.each { conditionName ->
+            RunWrapper runWrapper = script.getProperty("currentBuild")
+            String originalResultString = runWrapper.result
+            Result originalResult = originalResultString == null ? null : Result.fromString(originalResultString)
+            Result logicalResult = BuildCondition.getCombinedResult(runWrapper.rawBuild, stageError)
             try {
-                Closure c = responder.closureForSatisfiedCondition(conditionName, script.getProperty("currentBuild"),
+                Closure c = responder.closureForSatisfiedCondition(conditionName, runWrapper,
                     context, stageError)
                 if (c != null) {
+                    runWrapper.rawBuild.@result = logicalResult
                     catchRequiredContextForNode(agentContext) {
                         delegateAndExecute(c)
                     }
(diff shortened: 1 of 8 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -796,6 +801,18 @@ class ModelInterpreter implements Serializable {
...
+                Result currentResult = Result.fromString(runWrapper.currentResult);
+                // We can't leave the result set because it will interfere with the
+                // final build result, so we reset it to the original value unless
+                // the actual value changed inside of the post condition.However,
+                // if the build result is set to the logical build result manually,
+                // then after the post condition it will be set back to the original
+                // value. This seems ok since the logical result is the same and so
+                // will be reset for any later post conditions.
+                if (currentResult == logicalResult && logicalResult != originalResult) {
+                    runWrapper.rawBuild.@result = originalResult // Intentionally using .@ to bypass the setter which does not allow nulls.
```
dwnusbaum: `Run.result` is a protected field, so I am a little surprised that this works. Maybe just a Groovy thing? Also surprising to that accessing `runWrapper.rawBuild` does not seem to be forbidden by the Groovy sandbox.

abayer: Oh, but it is! `ModelInterpreter` isn't subject to the sandbox any more - initially it was, hence various `@Whitelisted` annotations floating around still, but when global shared libraries were removed from the sandbox, this kind of usage fell into the same bucket, so...no more whitelist.

And yeah, you can cheat like mad with regards to protected/private with Groovy. It's not a good thing, really, but might as well exploit that loophole, since we know damned well that the API isn't changing out from under us.

dwnusbaum: > when global shared libraries were removed from the sandbox, this kind of usage fell into the same bucket, so...no more whitelist

Ah ok that makes sense, thanks for the clarification!

abayer:
Merging and cutting 1.3.7-beta-1 right now.

PR #322, opened 2019-03-19
[JENKINS-56544] Don't set build result when a stage with an agent fails

dwnusbaum:
* JENKINS issue(s):
    * [JENKINS-56544]()
* Description:
    * Failure inside of a stage with an agent previously modified the build result. Now, the build result is left unmodified. 
    * The trigger for the failure post condition has been modified so that it is not executed along with the aborted post condition in some cases.
    * All post condition triggers for a specific build result now have identical implementations apart from the result they check for.
    * TODO: Add a JSON file for the newly added test?
    * TODO: Add some tests with Docker agents and failFast?
* Documentation changes:
    * TODO: Does the limitation mentioned under [the failure post condition in the documentation](https://jenkins.io/doc/book/pipeline/syntax/#post-conditions) still apply?
        > Note that if you manually set currentBuild.result = 'FAILURE' in a stage and have a failure post condition on that stage, the failure will not fire for that stage.
* Users/aliases to notify:
    * N/A

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
@@ -48,16 +48,7 @@ class Failure extends BuildCondition {
 
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
-        Result execResult = getExecutionResult(r)
-        if (context instanceof Stage && execResult != Result.ABORTED && r.getResult() != Result.ABORTED) {
-            return error != null
-        }
-        Result errorResult = null
-        if (error != null) {
-            errorResult = Utils.getResultFromException(error)
-        }
-        return execResult != Result.ABORTED &&
-            (execResult == Result.FAILURE || r.getResult() == Result.FAILURE || errorResult == Result.FAILURE)
+        return combineResults(r, error) == Result.FAILURE
     }
 
     @Override
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Success.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Success.groovy
@@ -48,17 +48,7 @@ class Success extends BuildCondition {
 
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
-        Result execResult = getExecutionResult(r)
-        if (context instanceof Stage && (execResult == Result.FAILURE || r.getResult() == Result.FAILURE)) {
-            return error == null
-        }
-        Result errorResult = null
-        if (error != null) {
-            errorResult = Utils.getResultFromException(error)
-        }
-        return (execResult == null || execResult.isBetterOrEqualTo(Result.SUCCESS)) &&
-                (r.getResult() == null || r.getResult().isBetterOrEqualTo(Result.SUCCESS)) &&
-                (errorResult == null || errorResult.isBetterOrEqualTo(Result.SUCCESS))
+        return combineResults(r, error) == Result.SUCCESS
     }
 
     @Override
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
@@ -39,13 +39,8 @@ class LabelScript extends DeclarativeAgentScript<Label> {
     @Override
     Closure run(Closure body) {
         return {
-            try {
-                script.node(describable?.label) {
-                    CheckoutScript.doCheckout(script, describable, describable.customWorkspace, body).call()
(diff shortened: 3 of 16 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
```diff
@@ -48,16 +48,7 @@ class Failure extends BuildCondition {
...
-        if (context instanceof Stage && execResult != Result.ABORTED && r.getResult() != Result.ABORTED) {
-            return error != null
-        }
-        Result errorResult = null
-        if (error != null) {
-            errorResult = Utils.getResultFromException(error)
-        }
-        return execResult != Result.ABORTED &&
-            (execResult == Result.FAILURE || r.getResult() == Result.FAILURE || errorResult == Result.FAILURE)
+        return combineResults(r, error) == Result.FAILURE
```
dwnusbaum: A little scary since I don't completely understand what all of the edge cases were for previously, but I looked at the regression tests that were added for some of the changes and they still seem to be passing, so maybe now that the build result is not set until later a lot of the edge cases don't matter? Happy to add some more tests for this if there is anything in particular you think would be a problem.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Success.groovy
```diff
@@ -48,17 +48,7 @@ class Success extends BuildCondition {
...
-            return error == null
-        }
-        Result errorResult = null
-        if (error != null) {
-            errorResult = Utils.getResultFromException(error)
-        }
-        return (execResult == null || execResult.isBetterOrEqualTo(Result.SUCCESS)) &&
-                (r.getResult() == null || r.getResult().isBetterOrEqualTo(Result.SUCCESS)) &&
-                (errorResult == null || errorResult.isBetterOrEqualTo(Result.SUCCESS))
+        return combineResults(r, error) == Result.SUCCESS
```
dwnusbaum: ditto, I definitely don't fully understand all of the edge cases that were being handled previously.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy
```diff
@@ -39,13 +39,8 @@ class LabelScript extends DeclarativeAgentScript<Label> {
...
     Closure run(Closure body) {
         return {
-            try {
-                script.node(describable?.label) {
-                    CheckoutScript.doCheckout(script, describable, describable.customWorkspace, body).call()
-                }
-            } catch (Exception e) {
-                script.getProperty("currentBuild").result = Utils.getResultFromException(e)
-                throw e
+            script.node(describable?.label) {
```
dwnusbaum: The main fix is just to stop setting the build result on errors, same for `DockerPipelineScript` and `DockerPipelineFromDockerfileScript`.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -1390,8 +1390,11 @@ public void parallelStagesFailFast() throws Exception {
                         "[Pipeline] { (first)",
                         "{ (Branch: second)",
                         "[Pipeline] { (second)",
-                        "SECOND STAGE FAILED")
-                .logNotContains("Second branch")
+                        "FIRST STAGE FAILURE",
+                        "SECOND STAGE ABORTED")
+                .logNotContains("Second branch",
+                        "FIRST STAGE ABORTED",
+                        "SECOND STAGE FAILURE")
```
dwnusbaum: I added post conditions for these tests to show exactly what conditions fire. Previously, for branch "second", which is aborted because of the failFast setting when branch "first" fails, both the aborted and failure post condition ran. Now, only the aborted condition runs. I did not go back to check what the behavior was _before_ my other PR, but that would probably be a good idea.

pipeline-model-definition/src/test/resources/parallelStagesFailFastWithAgent.groovy
```diff
@@ -0,0 +1,65 @@
...
+                        }
+                        failure {
+                            echo "FIRST STAGE FAILURE"
+                        }
+                    }
+                }
+                stage("second") {
+                    agent any
+                    options {
+                        skipDefaultCheckout()
```
dwnusbaum: Checkout is skipped to avoid timing issues if the branch is aborted while the checkout is happening because that causes the post conditions to not be executed.

dwnusbaum:
Ok, I verified locally that the following documentation is outdated:

> Note that if you manually set currentBuild.result = 'FAILURE' in a stage and have a failure post condition on that stage, the failure will not fire for that stage.

The post condition is executed in the following example:

```
pipeline {
    agent none
    stages {
        stage("foo") {
            steps {
                script {
                    currentBuild.result = 'FAILURE'
                }
                echo "MANUALLY CHANGED RESULT TO ${currentBuild.result}"
            }
            post {
                failure {
                    echo "FAILURE CONDITION RAN"
                }
            }
        }
    }
}
```

Want me to add the test here and file a PR to remove that line from the docs once this gets into a release?

abayer:
@dwnusbaum Sounds good!

PR #325, opened 2019-04-23
Update dependencies

jglick:
Same deal as https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/67: this repository is currently testing against rather old versions of critical dependencies.


pom.xml
```diff
@@ -85,12 +85,12 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-cps</artifactId>
-        <version>2.61.1</version>
+        <version>2.67</version>
```
jglick: Preparing for test of https://github.com/jenkinsci/workflow-cps-plugin/pull/211.

pom.xml
```diff
@@ -167,12 +167,12 @@
...
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-support</artifactId>
-        <version>2.22</version>
+        <version>${workflow-support-plugin.version}</version>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-support</artifactId>
-        <version>2.22</version>
+        <version>${workflow-support-plugin.version}</version>
```
jglick: Important to use properties to define versions of dependencies used with two classifiers, or you risk having them drift apart.

pom.xml
```diff
@@ -317,7 +317,6 @@
       <plugin>
         <groupId>org.codehaus.gmaven</groupId>
         <artifactId>gmaven-plugin</artifactId>
-        <version>1.5-jenkins-3</version>
```
jglick: Specified in parent.

pom.xml
```diff
@@ -340,10 +339,9 @@
           <tagNameFormat>pipeline-model-definition-@{project.version}</tagNameFormat>
         </configuration>
       </plugin>
-      <plugin>
+      <plugin> <!-- TODO can probably be deleted -->
         <groupId>org.jacoco</groupId>
         <artifactId>jacoco-maven-plugin</artifactId>
-        <version>0.7.9</version>
```
jglick: Newer specified in parent.

pom.xml
```diff
@@ -386,9 +384,8 @@
     <jenkins.version>2.121.1</jenkins.version>
     <java.level>8</java.level>
     <groovy.version>2.4.11</groovy.version>
-    <jenkins-test-harness.version>2.39</jenkins-test-harness.version>
```
jglick: Newer specified in parent.

pom.xml
```diff
@@ -386,9 +384,8 @@
...
-    <concurrency>1</concurrency>
```
jglick: Deprecated name, and that is anyway the default in the parent.

jglick:
This `Jenkinsfile` seems to apply a build-wide `timeout`, which does not work well on ci.jenkins.io which often has ridiculously long queue times. `buildPlugin.groovy` applies `timeout` only _within_ `node` for this reason. Not sure if there is a way to do the same in Declarative, besides using the wrapper and repeating it in each section.

jglick:
In the meantime I am running these tests locally and have collected a couple of failures so far which I know how to fix. This needs to be run in PCT!

pom.xml
```diff
@@ -167,12 +167,12 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-support</artifactId>
-        <version>2.22</version>
+        <version>${workflow-support-plugin.version}</version>
```
bitwiseman: Aren't these two identical entries?

Jenkinsfile
```diff
@@ -1,3 +1,4 @@
+// TODO replace all this with: buildPlugin(configurations: buildPlugin.recommendedConfigurations()) (yes, Scripted, so what)
```
bitwiseman: Submit a PR that does the replacement or file a JIRA for this task.
This does not belong in a TODO in this file and certainly not in this PR.

jglick:
Please do not push edits to my pull requests. I do not agree with e6e3d7fe130522e341a3c1780143446ef06663d6—the intent was to inherit configuration from the parent POM rather than having snowflake configuration.

pom.xml
```diff
@@ -339,7 +339,7 @@
           <tagNameFormat>pipeline-model-definition-@{project.version}</tagNameFormat>
         </configuration>
       </plugin>
-      <plugin> <!-- TODO can probably be deleted -->
+      <plugin>
```
bitwiseman: Similar to the other TODO: you want to delete it, do a PR (or do the deletion on this one) or file a JIRA. 

I'm happy to merge this PR without the TODO or we can wait and you can determine that this section can be deleted and change to deleting the section.  Either way the TODO doesn't belong.

bitwiseman:
Waiting for further input from @jglick.

pom.xml
jglick: > file a JIRA

Way too heavyweight for a minor bit of code cleanup, and not discoverable from the code itself. Just never mind.

PR #326, opened 2019-05-14
Move a bunch of test resources

bitwiseman:
No functional change, just moving tests to make them easier to navigate.


bitwiseman:
I'm still seeing one failure locally with `readTrusted` in `evaluateFrom`.  
I can't make sense of it let alone fix it.

PR #327, opened 2019-05-27
[JENKINS-57162] Dockerfile agent hash based on build args

keitwb:
JENKINS issue(s):
 - [JENKINS-57162](https://issues.jenkins-ci.org/browse/JENKINS-57162)

Description:

This extends the hash used to determine the image tag to include the
build args of the image since those generally result in a distinct image.

I had to do a bit of a hack in the tests to be able to use a fixed project name since that is part of the hash as well.


keitwb:
cc @abayer since you seem to have written most of the related code.

bertjwregeer:
:+1: this would be fantastic!

PR #330, opened 2019-06-05
[FIXED JENKINS-57826] Post should handle catchError stageResult

abayer:
* JENKINS issue(s):
    * [JENKINS-57826](https://issues.jenkins-ci.org/browse/JENKINS-57826)
* Description:
    * This probably needs `StatusAndTiming#findWarningBetween` to be made public rather than copy-pasting here.
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -66,7 +66,6 @@ import org.jenkinsci.plugins.workflow.actions.ThreadNameAction
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution
 import org.jenkinsci.plugins.workflow.cps.CpsScript
 import org.jenkinsci.plugins.workflow.cps.CpsThread
-import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode
 import org.jenkinsci.plugins.workflow.flow.FlowExecution
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode
 import org.jenkinsci.plugins.workflow.graph.BlockStartNode
@@ -78,7 +77,6 @@ import org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
 import org.jenkinsci.plugins.workflow.steps.Step
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor
-import org.jenkinsci.plugins.workflow.support.steps.StageStep
 import org.kohsuke.accmod.Restricted
 import org.kohsuke.accmod.restrictions.NoExternalUse
 
@@ -168,29 +166,6 @@ class Utils {
         }
     }
 
-    static Predicate<FlowNode> isStageWithOptionalName(final String stageName = null) {
-        return new Predicate<FlowNode>() {
-            @Override
-            boolean apply(@Nullable FlowNode input) {
-                if (input != null) {
-                    if (input instanceof StepStartNode &&
-                        ((StepStartNode) input).descriptor instanceof StageStep.DescriptorImpl &&
-                        (stageName == null || input.displayName == stageName)) {
-                        // This is a true stage.
-                        return true
-                    } else if (input.getAction(LabelAction.class) != null &&
-                        input.getAction(ThreadNameAction.class) != null &&
-                        (stageName == null || input.getAction(ThreadNameAction)?.threadName == stageName)) {
-                        // This is actually a parallel block
-                        return true
-                    }
-                }
-
-                return false
-            }
-        }
-    }
-
     static String stringToSHA1(String s) {
         return DigestUtils.sha1Hex(s)
     }
@@ -219,7 +194,7 @@ class Utils {
         LinearBlockHoppingScanner scanner = new LinearBlockHoppingScanner()
 
         FlowNode stageNode = execution.currentHeads.find { h ->
-            scanner.findFirstMatch(h, isStageWithOptionalName())
+            scanner.findFirstMatch(h, CommonUtils.isStageWithOptionalName())
         }
 
         return stageNode != null
@@ -252,7 +227,7 @@ class Utils {
 
         ForkScanner scanner = new ForkScanner()
 
-        FlowNode stage = scanner.findFirstMatch(execution.currentHeads, null, isStageWithOptionalName(stageName))
+        FlowNode stage = scanner.findFirstMatch(execution.currentHeads, null, CommonUtils.isStageWithOptionalName(stageName))
 
         if (stage != null) {
             nodes.add(stage)
@@ -351,7 +326,7 @@ class Utils {
 
         ForkScanner scanner = new ForkScanner()
 
-        FlowNode stage = scanner.findFirstMatch(execution.currentHeads, null, isStageWithOptionalName(stageName))
+        FlowNode stage = scanner.findFirstMatch(execution.currentHeads, null, CommonUtils.isStageWithOptionalName(stageName))
 
         if (stage != null && stage instanceof BlockStartNode) {
             nodes.add(stage)
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Changed.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Changed.groovy
@@ -51,7 +51,7 @@ class Changed extends BuildCondition {
 
         // Get the *worst* result of either the execution or the run. If the run's result is null, that's effectively
         // SUCCESS.
-        Result runResult = combineResults(r, error)
+        Result runResult = combineResults(r, error, null)
 
         // If there's no previous build, we're inherently changed.
         if (prev == null) {
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
@@ -46,7 +46,7 @@ class Failure extends BuildCondition {
 
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
-        return combineResults(r, error) == Result.FAILURE
+        return combineResults(r, error, context) == Result.FAILURE
     }
 
     @Override
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
@@ -51,7 +51,7 @@ class Fixed extends BuildCondition {
 
         // Get the *worst* result of either the execution or the run. If the run's result is null, that's effectively
         // SUCCESS.
-        Result runResult = combineResults(r, error)
+        Result runResult = combineResults(r, error, null)
 
         // If there's no previous build, we can't exactly be fixed, can we?
         if (prev == null) {
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/NotBuilt.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/NotBuilt.groovy
@@ -46,7 +46,7 @@ class NotBuilt extends BuildCondition {
 
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
-        return combineResults(r, error) == Result.NOT_BUILT
+        return combineResults(r, error, context) == Result.NOT_BUILT
     }
 
     @Override
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -36,6 +36,7 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.options.DeclarativeOption
 import org.jenkinsci.plugins.pipeline.modeldefinition.steps.CredentialWrapper
 import org.jenkinsci.plugins.pipeline.modeldefinition.when.DeclarativeStageConditional
 import org.jenkinsci.plugins.workflow.cps.CpsScript
+import org.jenkinsci.plugins.workflow.job.WorkflowRun
 import org.jenkinsci.plugins.workflow.steps.MissingContextVariableException
 import org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper
 
@@ -185,7 +186,7 @@ class ModelInterpreter implements Serializable {
                 if (skippedReason == null && parent != null &&
                     root.hasSatisfiedConditions(parent.post, script.getProperty("currentBuild"), parent, firstError)) {
                     Utils.logToTaskListener("Post stage")
-                    firstError = runPostConditions(parent.post, parent.agent ?: root.agent, firstError, parent.name, parent)
+                    firstError = runPostConditions(parent.post, parent.agent ?: root.agent, firstError, parent.name)
                 }
             }
 
@@ -337,7 +338,7 @@ class ModelInterpreter implements Serializable {
                         root.hasSatisfiedConditions(thisStage.post, script.getProperty("currentBuild"), thisStage, firstError) &&
                         thisStage?.parallel != null) {
                         Utils.logToTaskListener("Post stage")
-                        firstError = runPostConditions(thisStage.post, thisStage.agent ?: parentAgent, firstError, thisStage.name, thisStage)
+                        firstError = runPostConditions(thisStage.post, thisStage.agent ?: parentAgent, firstError, thisStage.name)
                     }
                 }
 
@@ -679,6 +680,7 @@ class ModelInterpreter implements Serializable {
         Throwable stageError = null
         try {
             catchRequiredContextForNode(thisStage.agent ?: parentAgent) {
+                System.err.println("GOT CONTEXT FOR ${thisStage.name}")
                 delegateAndExecute(thisStage.steps.closure)
             }
         } catch (Throwable e) {
@@ -690,7 +692,7 @@ class ModelInterpreter implements Serializable {
             // And finally, run the post stage steps.
             if (root.hasSatisfiedConditions(thisStage.post, script.getProperty("currentBuild"), thisStage, stageError)) {
                 Utils.logToTaskListener("Post stage")
-                stageError = runPostConditions(thisStage.post, thisStage.agent ?: parentAgent, stageError, thisStage.name, thisStage)
+                stageError = runPostConditions(thisStage.post, thisStage.agent ?: parentAgent, stageError, thisStage.name)
             }
         }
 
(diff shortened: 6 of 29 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/NotBuilt.groovy
```diff
@@ -46,7 +46,7 @@ class NotBuilt extends BuildCondition {
 
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
-        return combineResults(r, error) == Result.NOT_BUILT
+        return combineResults(r, error, null, null) == Result.NOT_BUILT
```
dwnusbaum: Someone could use NOT_BUILT inside of catchError (although it would probably not make much sense). My vote would be to make this match all of the other specific-result conditions.

pipeline-model-definition/src/test/resources/postStage/catchErrorStageAborted.groovy
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+    }
+
+    post {
+        success {
+            echo "The build should be a success"
+        }
+        aborted {
+            echo "The build shouldn't be aborted"
```
dwnusbaum: I'm actually a little surprised that this isn't happening (even though I think it's good that it isn't). Why don't we find the `WarningAction` for the nested stage?

In Blue Ocean this doesn't really matter because the parents of nested stages don't have their actual status reported, but I would've thought that in `isInGivenStage` here you would need to find nodes where the only enclosing block is the current stage, not where any of the enclosing blocks is the current stage.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
```diff
@@ -130,8 +140,28 @@ protected Result getExecutionResult(@Nonnull WorkflowRun r) {
...
     }
 
+    @Deprecated
     @Nonnull
     public static Result getCombinedResult(@Nonnull WorkflowRun run, @CheckForNull Throwable error) {
+        return getCombinedResult(run, error, null, null);
+    }
+
+    @Nonnull
+    public static Result getCombinedResult(@Nonnull WorkflowRun run, @CheckForNull Throwable error, @CheckForNull FlowNode startNode, @CheckForNull FlowNode endNode) {
```
dwnusbaum: FWIW I think you could remove some duplicated code by changing this to `getCombinedResult(@Nonnull WorkflowRun run, @CheckForNull Throwable error, @CheckForNull String stageName)` and doing the following here rather than in all the specific-result conditions:

```
FlowNode startNode = null
FlowNode endNode = null
if (context instanceof Stage) {
    (startNode, endNode) = Utils.findPossiblyUnfinishedEndNodeForCurrentStage(((Stage)context).name, r.getExecution())
}
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/NotBuilt.groovy
abayer: Good point.

pipeline-model-definition/src/test/resources/postStage/catchErrorStageAborted.groovy
abayer: So `ABORTED` shouldn't be the overall build status, but there definitely should be testing for nested stages, which I don't have here.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
abayer: ...if I could, I would, but `Stage` is in `pipeline-model-definition`, while this is in `pipeline-model-extensions`...

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
```diff
@@ -130,8 +140,28 @@ protected Result getExecutionResult(@Nonnull WorkflowRun r) {
...
+            // TODO: Remove this in favor of moving StatusAndTiming#findWarningBetween public.
+            DepthFirstScanner scanner = new DepthFirstScanner();
+            if (scanner.setup(endNode, Collections.singletonList(startNode))) {
+                WarningAction warningAction = StreamSupport.stream(scanner.spliterator(), false)
+                        .map(node -> node.getPersistentAction(WarningAction.class))
+                        .filter(Objects::nonNull)
+                        .max(Comparator.comparing(warning -> warning.getResult().ordinal))
+                        .orElse(null);
+                if (warningAction != null) {
+                    return warningAction.getResult();
```
dwnusbaum: I think you don't want an explicit return here, because you want to use the worst of the stage, exec, run, and error results, right?

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
dwnusbaum: But you just need the name, right? You can do the `(Stage)context` cast up in the conditions and just pass down the String, although I guess you'd need to move `findPossiblyUnfinishedEndNodeForCurrentStage` to pipeline-model-extensions to make it work.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
abayer: Do we? I'm legitimately not sure.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
abayer: Yeah. I have regrets about the organization. =)

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
abayer: That is to say - do we want to have a successful individual stage trigger `post` `success` even if the overall build status is failure? I feel like we will...

dwnusbaum: Hmm. My concern is breaking existing use cases where people are setting the build result explicitly, but now that I'm thinking about it it might be more nuanced in this case.

Here is the kind of thing where I'm not sure what makes the most sense:

```
pipeline {
    agent any

     stages {
        stage("foo") {
            steps {
                // Default Settings
                catchError(buildResult: "FAILURE", stageResult: "SUCCESS") {
                    error("uhoh")
                }
                catchError(buildResult: "SUCCESS", stageResult: "UNSTABLE") {
                    error("uhoh")
                }
            }
            post {
                unstable {
                    echo "Does this run?"
                }
                failure {
                    echo "Or this?"
                }
            }
        }
    }
}
```
Anyone using `catchError(buildResult: 'SUCCESS', stageResult: 'blah')` is probably going to be careful to not set the build result anywhere intentionally, so maybe the current behavior is the best option?

dwnusbaum: > do we want to have a successful individual stage trigger post success even if the overall build status is failure?

`catchError` does not create a `WarningAction` for `SUCCESS`, so it wouldn't hit the return and we'd use the overall build result. (Although we could change that behavior if it would make more sense for the use case here since it wouldn't break anything else).

abayer: Ok, yeah, let's just fallback then.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -168,6 +168,18 @@ class Utils {
         }
     }
 
+    static Predicate<FlowNode> isInGivenStage(final FlowNode stageStartNode) {
```
bitwiseman:
```suggestion
    static Predicate<FlowNode> containedInStage(final FlowNode stageStartNode) {
```
Or...
```suggestion
    static Predicate<FlowNode> childOf(final FlowNode stageStartNode) {
```
Or
```suggestion
    static Predicate<FlowNode> enclosedBy(final FlowNode stageStartNode) {
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Changed.groovy
```diff
@@ -26,7 +26,10 @@ package org.jenkinsci.plugins.pipeline.modeldefinition.model.conditions
 import hudson.Extension
 import hudson.model.Result
 import org.jenkinsci.Symbol
+import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.Stage
+import org.jenkinsci.plugins.workflow.graph.FlowNode
```
bitwiseman: Are these import changes needed?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
```diff
@@ -26,7 +26,10 @@ package org.jenkinsci.plugins.pipeline.modeldefinition.model.conditions
 import hudson.Extension
 import hudson.model.Result
 import org.jenkinsci.Symbol
+import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.Stage
+import org.jenkinsci.plugins.workflow.graph.FlowNode
```
bitwiseman: Are these import changes needed?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
```diff
@@ -46,7 +49,12 @@ class Failure extends BuildCondition {
 
     @Override
     boolean meetsCondition(@Nonnull WorkflowRun r, Object context, Throwable error) {
-        return combineResults(r, error) == Result.FAILURE
+        FlowNode startNode = null
+        FlowNode endNode = null
+        if (context instanceof Stage) {
+            (startNode, endNode) = Utils.findPossiblyUnfinishedEndNodeForCurrentStage(((Stage)context).name, r.getExecution())
+        }
+        return combineResults(r, error, startNode, endNode) == Result.FAILURE
```
bitwiseman: Refactor to a shared method the returns `boolean`?
```java
        return combinedResultEquals(r, error, context, Result.FAILURE);
```
Or at least: 
```java
        return combineResults(r, error, context) == Result.FAILURE;
```

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
```diff
@@ -77,9 +81,15 @@ protected final Result combineResults(@Nonnull WorkflowRun run) {
...
 
+    @Deprecated
     @Nonnull
     protected final Result combineResults(@Nonnull WorkflowRun run, @CheckForNull Throwable error) {
-        return BuildCondition.getCombinedResult(run, error);
+        return combineResults(run, error, null, null);
+    }
+
+    @Nonnull
+    protected final Result combineResults(@Nonnull WorkflowRun run, @CheckForNull Throwable error, @CheckForNull FlowNode startNode, @CheckForNull FlowNode endNode) {
```
bitwiseman:
```suggestion
    protected final Result combineResults(@Nonnull WorkflowRun run, @CheckForNull Throwable error, @CheckForNull Object context) {
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -168,6 +168,25 @@ class Utils {
...
     }
 
+    // TODO: Is there some magic trick for filtering out *nested* stages from a list? If so, that'd be handy.
+    /**
+     * This will return true for flow nodes in *child* stages, not just the immediate enclosing stage.
+     *
+     * @param stageStartNode
+     * @return A predicate that returns true if the applied input is somewhere within the given stage
+     */
+    static Predicate<FlowNode> isSomewhereWithinGivenStage(final FlowNode stageStartNode) {
```
bitwiseman:
```suggestion
    static Predicate<FlowNode> isSomewhereWithinStage(final FlowNode stageStartNode) {
```
Sorry, it's the "Given" that's bothering me.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Changed.groovy
abayer: oops!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Fixed.groovy
abayer: Oops!

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
abayer: Wish I could, but `Stage` is in `pipeline-model-definition`, so we're kinda stuck with this.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
abayer: heheh, ok.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/conditions/Failure.groovy
abayer: Well, it's gotta be `(r, error, startNode, endNode)` (as explained elsewhere), so I tend towards just leaving it as is.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
bitwiseman: Well, then make a static method over in `pipeline-model-definition`?  Or a new base class local to `pipeline-model-definition`?   There are options.

abayer: We need `BuildCondition` to be here so that it can theoretically be used to implement new conditions outside of Declarative proper, so...yeah. This is what it is. I regret the way things are organized here, but I don't see this as near bad enough to merit the refactoring involved and potential impact.

bitwiseman: okay, I'm not going to block on this, but if the advantage is purely theoretical and causes code smells we should keep track of it as an issue for the theoretical future where we take the time to fix design issues.

abayer: The idea is that the extension points, such as `BuildCondition`, should only need to depend on `pipeline-model-extensions`, not `pipeline-model-definition`, and shouldn't need to know anything about the runtime or parse-time models Declarative uses (with the exception of `DeclarativeValidatorContributor` needing to know about the various `ModelAST*` classes for fairly obvious reasons) - so the more I think about it, the more I think that passing `FlowNode`s to `#combineResults` is actually preferable to passing a `Stage`. That lets the `BuildCondition` implementations not be adversely affected by changes to the runtime model - `FlowNode` is basically a primitive construct here, while `Stage` carries a lot more specific meaning that we might want to alter in the future.

pipeline-model-definition/src/test/resources/postStage/catchErrorStageAborted.groovy
dwnusbaum: Ok, would be good to check to make sure, because if it does use the `WarningAction` from nested stages you'll need to tweak `findWorstWarningBetween` for your use case anyway so no point in making it public over in pipeline-graph-analysis.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
dwnusbaum: Yeah, my main thought is that if https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/330#discussion_r290757003 is a problem, you'll need the name of the stage wherever your `findWarningBetween` logic lives to make it able to skip nodes in nested stages, so since `Util.findPossiblyUnfinishedEndNodeForCurrentStage` and `Util.isSomewhereWithinStage` are new methods anyway, might as well move them all into pipeline-model-extensions and just pass the stage name to `BuildCondition.whatever` and then do all of the work in one `findWorstWarningActionForStage(String stageName)` method over there. It isn't quite that clean though because you'd have to dupe or move `Util.isStageWithOptionalName`. If that turns out not to be a problem, then it doesn't matter.

pipeline-model-definition/src/test/resources/postStage/catchErrorStageAborted.groovy
abayer: Yeah, I added a nested test. For now, I decided to just stick with the logic used elsewhere and treat parent stages with a child stage with a `WarningAction` as if the `WarningAction` was directly on the parent stage. I'm honestly not sure what the ideal behavior there _should_ be, but this feels fine for now.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
abayer: Hmm. Lemme think on that.

pipeline-model-definition/src/test/resources/postStage/catchErrorStageFailureNested.groovy
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+    }
+
+    post {
+        success {
+            echo "The build should be a success"
+        }
+        failure {
+            echo "The build shouldn't be a failure"
```
dwnusbaum: I'm still confused as to why this isn't being run. I would think the relevant flow node start would be the start of the very first stage, and then `findWorstWarningBetween` would search the whole Pipeline and find the `WarningAction` in `foo`... I might pull down the PR and print out what FlowNodes are being used in the search to understand what is happening.

abayer: This is the `post` for the *build* - so it's only looking at the overall build result, which is explicitly set to `SUCCESS`.

dwnusbaum: Ah, gotcha, thanks for clarifying!

abayer:
Ok, rewrote/moved a bunch of stuff around. I realized all we actually needed was the `stageName`, so I'm using that as the `context` now, but the signatures are all still `Object context` 'cos I didn't want to break existing APIs or throw in a pile of deprecation etc.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -185,7 +186,7 @@ class ModelInterpreter implements Serializable {
                 if (skippedReason == null && parent != null &&
                     root.hasSatisfiedConditions(parent.post, script.getProperty("currentBuild"), parent, firstError)) {
```
KalleOlaviNiemitalo, line 187: This call passes `Stage parent` as `Object context` to Root.hasSatisfiedConditions, from where the context propagates to BuildCondition.getCombinedResult, which however expects the context to be the name of the stage as a String, rather than a Stage instance. Filed [JENKINS-68288](https://issues.jenkins.io/browse/JENKINS-68288).

PR #332, opened 2019-06-07
Refactor Model AST methods to reduce repeated code

bitwiseman:
* JENKINS issue(s):
    * n/a
* Description:
    * Refactor of AST methods to reduce repeated code and improve understandability.
* Documentation changes:
    * Not a notable changed externally unless we want to document the added static helpers
* Users/aliases to notify:
    * @dwnusbaum

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
@@ -30,9 +30,7 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.validator.ModelValidator;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 
 public abstract class ModelASTElement implements ModelASTMarkerInterface {
@@ -151,7 +149,7 @@ public abstract class ModelASTElement implements ModelASTMarkerInterface {
      *
      * @return A simple {@link String} of Groovy code for this element and its children.
      */
-    protected static <T extends ModelASTMarkerInterface> String toGroovyArgList(List<T> list) {
+    protected static <T extends ModelASTMarkerInterface> String toGroovyArgList(Collection<T> list) {
         StringBuilder result = new StringBuilder();
         boolean first = true;
         for (T item : list) {
@@ -244,19 +242,25 @@ public abstract class ModelASTElement implements ModelASTMarkerInterface {
         // No-op
     }
 
-    protected static void validate(@Nonnull ModelValidator validator, @CheckForNull ModelASTMarkerInterface item) {
-        if (item != null) {
-            item.validate(validator);
+    protected static void validate(@Nonnull ModelValidator validator, @CheckForNull ModelASTMarkerInterface... items) {
+        if (items != null && items.length > 0) {
+            validate(validator, Arrays.asList(items));
         }
     }
 
-    protected static <T extends ModelASTMarkerInterface> void validate(@Nonnull ModelValidator validator, @CheckForNull List<T> list) {
+    protected static <T extends ModelASTMarkerInterface> void validate(@Nonnull ModelValidator validator, @CheckForNull List<T> list, @CheckForNull ModelASTMarkerInterface... items) {
+        validate(validator, items);
         if (list != null) {
-            list.forEach( item -> item.validate(validator));
+            for (T item : list) {
+                if (item != null) {
+                    item.validate(validator);
+                }
+            }
         }
     }
 
-    protected static <K extends  ModelASTMarkerInterface, V extends  ModelASTMarkerInterface> void validate(@Nonnull ModelValidator validator, @CheckForNull Map<K, V> map) {
+    protected static <K extends  ModelASTMarkerInterface, V extends  ModelASTMarkerInterface> void validate(@Nonnull ModelValidator validator, @CheckForNull Map<K, V> map, @CheckForNull ModelASTMarkerInterface... items) {
+        validate(validator, items);
         if (map != null) {
             for (Map.Entry<K, V> entry : map.entrySet()) {
                 entry.getKey().validate(validator);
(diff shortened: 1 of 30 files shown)


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidator.java
```diff
@@ -29,6 +29,12 @@
 
 
 public interface ModelValidator {
+//    <T extends  ModelASTMarkerInterface> void validate(@CheckForNull T item);
+//
+//    <T extends  ModelASTMarkerInterface> void validate(@CheckForNull List<T> items);
+//
+//    <K extends  ModelASTMarkerInterface, V extends  ModelASTMarkerInterface> void validate(@CheckForNull Map<K, V> map);
```
dwnusbaum: What is this about?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -257,7 +257,7 @@ class ModelParser implements Parser {
 
         r.validate(validator)
 
-        // Lazily evaluate r.toJSON() - i.e., only if AST_DEBUG_LOGGING is true.
+        // Lazily evaluate r.toJSONObject() - i.e., only if AST_DEBUG_LOGGING is true.
```
dwnusbaum: The code wasn't changed, maybe this should stay as `toJSON`?

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/DeclarativeValidatorContributor.java
```diff
@@ -47,7 +47,7 @@
 public abstract class DeclarativeValidatorContributor implements ExtensionPoint {
 
     /**
-     * Implementations default to not optional. Can be overridden.
+     * Implementations default to not element. Can be overridden.
```
dwnusbaum: Maybe an accidental replace-all? Looks like it should be reverted.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
```diff
@@ -73,13 +244,66 @@ public void validate(@Nonnull ModelValidator validator) {
...
      * Removes the source location value from this element.
      */
     public void removeSourceLocation() {
         sourceLocation = null;
     }
 
+    /**
+     * Removes the source location value from this element.
+     */
+    protected static void removeSourceLocation(@CheckForNull ModelASTMarkerInterface item) {
```
dwnusbaum: Could you instead change the original `removeSourceLocation` to something like:

```
protected static void removeSourceLocation(ModelASTMarkerInterface... items) {
    sourceLocation = null;
    if (items != null) { // If someone calls `removeSourceLocation(nullArg)`
        for (ModelASTMarkerInterface item : items) {
            if (item != null) {
                item.removeSourceLocation();
            }
        }
    }
}
```
And then delete the new method? That way code like this:
```
super.removeSourceLocation();
removeSourceLocation(agentType);
removeSourceLocation(variables);
```
Would just become this:
```
super.removeSourceLocation(agentType, variables);
```
I think existing callers of `removeSourceLocation()` would be fine, although I'm not sure.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStage.java
```diff
@@ -112,87 +78,43 @@ public void validate(@Nonnull final ModelValidator validator) {
...
-            result.append("}\n");
-        }
+        result.append(toGroovy(agent));
+        result.append(toGroovy(when));
+        result.append(toGroovy(tools));
+        result.append(toGroovy(environment));
+        result.append(toGroovy(options));
+        result.append(toGroovy(input));
+        result.append(toGroovy(post));
+        result.append(toGroovy(stages));
```
dwnusbaum: Does this still have the surrounding `stages { ... }`? Same for `parallel` and `matrix` below. Maybe I am looking at the wrong `toJSON` overload? 

Ah, ok, It looks like you moved that part down into the types themselves, i.e. `ModelASTStages.toGroovy`.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStages.java
```diff
@@ -38,26 +34,20 @@ public void validate(@Nonnull final ModelValidator validator) {
 
     public void validate(final ModelValidator validator, boolean isWithinParallel) {
         validator.validateElement(this);
-        for (ModelASTStage stage : stages) {
-            stage.validate(validator, isWithinParallel);
+        if (stages != null) {
```
dwnusbaum: Looking at the usage in the class (`equals`) I think `stages` is meant to be `@Nonnull` and the setter should be more restrictive. Either way this doesn't look like a refactoring, maybe handle it separately?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidator.java
bitwiseman: Garbage.

dwnusbaum:
Although it looks like there is a compilation failure somewhere.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTElement.java
bitwiseman: @dwnusbaum  
I like this in principle.
For anything the derives directly from element this would be fine.
But for deeper derived classes, we still need to call `super.removeSourceLocation()` to traverse up the class tree clearing source locations at each level right?  

I I like the `ModelASTMarkerInterface...  items` though.

dwnusbaum: > But for deeper derived classes, we still need to call super.removeSourceLocation() to traverse up the class tree clearing source locations at each level right?

I think as long as there was only one `removeSourceLocation(...)` method, and every implementor of that method called `super.removeSourceLocation(...)` concatenating passed arguments with any additional fields from the current class, it would work, but the need to concatenate args would probably make it more complicated that what you have right now, so it seems best to just ignore my suggestion.

bitwiseman:
Now I'm seeing new errors which I'm pretty sure are unrelated to my change:

```
Running in Durability level: MAX_SURVIVABILITY
[Pipeline] Start of Pipeline
[Pipeline] End of Pipeline
java.lang.UnsupportedOperationException: Refusing to marshal org.codehaus.groovy.ast.expr.ConstantExpression for security reasons; see https://jenkins.io/redirect/class-filter/
at hudson.util.XStream2$BlacklistedTypesConverter.marshal(XStream2.java:546)
at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:69)
at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:58)
at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:84)
at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:265)
at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:252)
Caused: java.lang.RuntimeException: Failed to serialize org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTElement#sourceLocation for class org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTSingleArgument
at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:256)
at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:224)
```

abayer:
Pretty sure it is related - a source location isn't getting removed, causing serialization to fail.

bitwiseman:
@abayer All fixed.

PR #334, opened 2019-06-11
Remove call to System.err.println in ModelInterprer.groovy

dwnusbaum:
* JENKINS issue(s):
    * N/A
* Description:
    * Looks like this was unintentionally committed as part of https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/330, noticed it while running some stuff locally.
* Documentation changes:
* Users/aliases to notify:
    * N/A


abayer:
Whoops. Thanks!

daniel-beck:
How about we get a release of 1.3.x with this nonsense removed?

This has cost me hours to find over the last few days and was a big distraction when trying to find what's going wrong with ci.j.io.

PR #335, opened 2019-06-11
Using standard buildPlugin

jglick:
As per https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/325#discussion_r292496197.


Jenkinsfile
```diff
@@ -1,107 +1 @@
-// This Jenkinsfile's main purpose is to show a real-world-ish example
-// of what Pipeline config syntax actually looks like.
```
jglick: Obsolete as there are plenty of other examples of Declarative including in tests, on jenkins.io, etc.

Jenkinsfile
```diff
@@ -1,107 +1 @@
...
-        jdk "jdk8"
-    }
-
-    agent none
-
-    // Set log rotation, timeout and timestamps in the console
-    options {
-        buildDiscarder(logRotator(numToKeepStr:'10'))
-        timestamps()
-        timeout(time: 120, unit: 'MINUTES')
```
jglick: Undesirable to set a build timeout, as oftentimes particular stages are waiting for a long time in queue through no fault of this plugin.

Jenkinsfile
```diff
@@ -1,107 +1 @@
...
-    stages {
-        // While there is only one stage here, you can specify as many stages as you like!
-        stage("build") {
-            parallel {
-                stage("linux") {
-                    agent {
-                        label "highmem"
-                    }
-                    steps {
-                        sh "mvn -B clean install -Dmaven.test.failure.ignore=true -Djenkins.test.timeout=${TEST_TIMEOUT}"
```
jglick: TBD whether this still needs to be overridden. Are there individual tests which normally run >10m?!

Jenkinsfile
```diff
@@ -1,107 +1 @@
...
-                    }
-                    post {
-                        // No matter what the build status is, run this step. There are other conditions
-                        // available as well, such as "success", "failed", "unstable", and "changed".
-                        always {
-                            junit testResults: '*/target/surefire-reports/*.xml', keepLongStdio: true
-                        }
-                        success {
-                            archive "**/target/*.hpi"
-                            archive "**/target/site/jacoco/jacoco.xml"
```
jglick: IIRC there is an actual JaCoCo publisher available if anyone is looking at the results.

Jenkinsfile
```diff
@@ -1,107 +1 @@
...
-                        }
-                    }
-                }
-                stage("windows") {
-                    agent {
-                        label "windows"
-                    }
-                    steps {
-                        bat "mvn -B clean install -Dconcurrency=1 -Dmaven.test.failure.ignore=true -Dcodenarc.skip=true -Djenkins.test.timeout=${TEST_TIMEOUT}"
```
jglick: Should be harmless to run codenarc in all branches I suppose.

Jenkinsfile
```diff
@@ -1,107 +1 @@
...
-        buildDiscarder(logRotator(numToKeepStr:'10'))
-        timestamps()
-        timeout(time: 120, unit: 'MINUTES')
-    }
-
-    // Make sure we have GIT_COMMITTER_NAME and GIT_COMMITTER_EMAIL set due to machine weirdness.
-    environment {
-        GIT_COMMITTER_NAME = "jenkins"
-        GIT_COMMITTER_EMAIL = "jenkins@jenkins.io"
-        NEWER_CORE_VERSION = "2.164.3"
```
jglick: Handled better by the standard library.

jglick:
Hmm, on Java 11 / 2.164.1 I am getting a gmaven error that I thought would have been solved by https://github.com/jenkinsci/jenkins/pull/3504:

```
[INFO] --- gmaven-plugin:1.5-jenkins-3:compile (default) @ pipeline-model-definition ---
Exception in thread "main" BUG! exception in phase 'class generation' in source unit '…/pipeline-model-definition-plugin/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy' unsupported Target MODULE
	at org.codehaus.groovy.vmplugin.v5.Java5.getElementCode(Java5.java:307)
	at org.codehaus.groovy.vmplugin.v5.Java5.configureAnnotationFromDefinition(Java5.java:204)
	at org.codehaus.groovy.vmplugin.v5.Java5.configureAnnotation(Java5.java:214)
	at org.codehaus.groovy.classgen.AnnotationVisitor.visit(AnnotationVisitor.java:81)
	at org.codehaus.groovy.classgen.ExtendedVerifier.visitAnnotation(ExtendedVerifier.java:153)
	at org.codehaus.groovy.classgen.ExtendedVerifier.visitAnnotations(ExtendedVerifier.java:114)
	at org.codehaus.groovy.classgen.ExtendedVerifier.visitField(ExtendedVerifier.java:63)
	at org.codehaus.groovy.ast.ClassNode.visitContents(ClassNode.java:1048)
	at org.codehaus.groovy.classgen.ExtendedVerifier.visitClass(ExtendedVerifier.java:59)
	at org.codehaus.groovy.control.CompilationUnit$14.call(CompilationUnit.java:746)
	at org.codehaus.groovy.control.CompilationUnit.applyToPrimaryClassNodes(CompilationUnit.java:967)
	at org.codehaus.groovy.control.CompilationUnit.doPhaseOperation(CompilationUnit.java:546)
	at org.codehaus.groovy.control.CompilationUnit.processPhaseOperations(CompilationUnit.java:524)
	at org.codehaus.groovy.control.CompilationUnit.compile(CompilationUnit.java:501)
	at org.codehaus.groovy.control.CompilationUnit.compile(CompilationUnit.java:480)
	at org.codehaus.gmaven.runtime.v1_8.ClassCompilerFeature$ClassCompilerImpl.compile(ClassCompilerFeature.java:148)
	at org.codehaus.gmaven.plugin.compile.AbstractCompileMojo.compile(AbstractCompileMojo.java:211)
	at org.codehaus.gmaven.plugin.compile.AbstractCompileMojo.process(AbstractCompileMojo.java:175)
	at org.codehaus.gmaven.plugin.ComponentMojoSupport.doExecute(ComponentMojoSupport.java:60)
	at org.codehaus.gmaven.plugin.MojoSupport.execute(MojoSupport.java:71)
	at …
```

Ring a bell @oleg-nenashev / @jenkinsci/java11-support?

jglick:
No mention of `MODULE` in the source tree, and none in `pipeline-model-definition/target/generated-sources/groovy-stubs/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.java` either.

oleg-nenashev:
@jglick In https://github.com/jenkinsci/jenkins/pull/3504 we fixed Jenkins itself, not developer tools. I suspect old Groovy is used in gmaven or in infra

oleg-nenashev:
CC @jenkinsci/java11-support just in case

jglick:
Trying

```diff
diff --git a/pom.xml b/pom.xml
index 47587594..2d5c407b 100644
--- a/pom.xml
+++ b/pom.xml
@@ -313,7 +313,7 @@
         <groupId>org.codehaus.gmaven</groupId>
         <artifactId>gmaven-plugin</artifactId>
         <configuration>
-          <providerSelection>1.8</providerSelection>
+          <providerSelection>2.0</providerSelection>
         </configuration>
         <executions>
           <execution>
```

but it gives me cryptic errors

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project pipeline-model-definition: Compilation failure: Compilation failure: 
[ERROR] …/pipeline-model-definition-plugin/pipeline-model-definition/target/generated-sources/groovy-stubs/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.java:[69,230] incompatible types: org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTBuildConditionsContainer cannot be converted to R
[ERROR] …/pipeline-model-definition-plugin/pipeline-model-definition/target/generated-sources/groovy-stubs/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java:[45,124] incompatible types: org.jenkinsci.plugins.pipeline.StageTagsMetadata cannot be converted to T
[ERROR] …/pipeline-model-definition-plugin/pipeline-model-definition/target/generated-sources/groovy-stubs/org/jenkinsci/plugins/pipeline/modeldefinition/parser/JSONParser.java:[53,228] incompatible types: org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTBuildConditionsContainer cannot be converted to R
```

IMO we should just ban `src/main/groovy/` from plugins. (As far as I know it is not actually needed here, despite use of Groovy at runtime. It is just “for fun”.)

Jenkinsfile
bitwiseman: @jglick But is it available on ci.jenkins.io?

bitwiseman:
@jglick
Thanks for doing this. I think using the standard plugin build pipeline for this plugin is a great idea. 

We didn't run Java11 on this plugin before this change.  Enabling java 11 can be a separate PR. 

I would like to keep the jacoco running file publishing for consistency.    

Re: groovy: I'm not understanding your point about banning groovy and "just for fun".

jglick:
> I'm not understanding your point about banning groovy and "just for fun".

So there are two kinds of Groovy sources in this plugin (outside tests):

* `pipeline-model-definition/src/main/resources/**/*.groovy` are loaded as resources at runtime and passed through `CpsTransformer`, so they effectively form part of the program when using Declarative Pipeline.
* `pipeline-model-definition/src/main/groovy/**/*.groovy` are compiled alongside `pipeline-model-definition/src/main/java/**/*.java` during the Maven build and simply form part of the plugin’s code.

The second category is what is causing problems for Java 11 compatibility here, since it relies on a Jenkins fork of `gmaven-plugin`, which was abandoned by its authors years ago. (The suggested replacement is to use the Eclipse compiler—for everything, including Java.) There is no particular reason (that I am aware of) for any parts of the plugin to be written in Groovy; @abayer just felt like it.

See https://github.com/jenkinsci/jenkins/pull/3184 for some historical context.

Jenkinsfile
jglick: Indeed there are, though I doubt the previous configuration actually affected the timeout: passing a system property to the `mvn` process does not automatically cause it to be propagated to the Surefire booter process.

Jenkinsfile
jglick: Looks like it is waiting for https://github.com/jenkins-infra/pipeline-library/pull/21. Not that the previous version actually did anything useful with the results; as you can see [here](https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/master/389/) for example, the raw output is available for download but there is no human-readable summary, `UNSTABLE` status when coverage regresses, etc.

Anyway this is a cross-cutting concern—nothing specific to this plugin.

Jenkinsfile
jglick: I see an error (https://github.com/gleclaire/codenarc-maven-plugin/issues/12) but it does not seem to be fatal.

Jenkinsfile
jglick: > passing a system property to the `mvn` process does not automatically cause it to be propagated to the Surefire booter process

Correction: it does, but only if it is passed directly from the command line using `-D`; an entry in `<properties>` is _not_ propagated.

pom.xml
```diff
@@ -398,8 +405,6 @@
     <jenkins.version>2.150.1</jenkins.version>
     <java.level>8</java.level>
     <groovy.version>2.4.12</groovy.version>
-    <jenkins-test-harness.version>2.39</jenkins-test-harness.version>
-    <concurrency>1</concurrency>
```
jglick: Mistake in #339.

jglick:
`ToolsTest.buildPluginParentPOM` is failing and I am trying to understand why.

jglick:
Got a timeout in a Windows branch, so this would not help:

```diff
diff --git a/Jenkinsfile b/Jenkinsfile
index 87a086da..3c18388c 100644
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1 +1 @@
-buildPlugin(configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(configurations: buildPlugin.recommendedConfigurations().collect({cfg -> cfg.platform == 'linux' ? cfg + [platform: 'highmem'] : cfg}))
```

jglick:
[INFRA-2075](https://issues.jenkins-ci.org/browse/INFRA-2075) this time.

pipeline-model-definition/src/test/resources/buildPluginParentPOM.groovy
```diff
@@ -17,11 +17,11 @@ pipeline {
                         if (isUnix()) {
                             sh 'echo "M2_HOME: ${M2_HOME}"'
                             sh 'echo "JAVA_HOME: ${JAVA_HOME}"'
-                            sh 'mvn clean verify -Dmaven.test.failure.ignore=true'
+                            sh 'mvn clean verify -Dmaven.test.failure.ignore=true -e'
                         } else {
                             bat 'echo "M2_HOME: %M2_HOME%"'
                             bat 'echo "JAVA_HOME: %JAVA_HOME%"'
-                            bat 'mvn clean verify -Dmaven.test.failure.ignore=true'
+                            bat 'mvn clean verify -Dmaven.test.failure.ignore=true -e'
```
bitwiseman: Seems reasonable to have this in case of error.

bitwiseman:
@jglick 
It looks like the build is taking almost twice along after this change.  Maybe the `reuseForks=false` is the source of that.  I'm okay merging this and figuring out how to improve it later.

pipeline-model-definition/src/test/resources/buildPluginParentPOM.groovy
jglick: I was trying to track down a failure which did not in fact reappear.

jglick:
> Maybe the `reuseForks=false` is the source of that.

Maybe, though there are not that many test suites (~ `*Test.java`) and the JVM startup time should be negligible compared to their run time. Or maybe it is `s/highmem/linux/`, though when I looked it was the Windows builds that were taking ~1h vs. the Linux builds taking ~30m.

PR #337, opened 2019-06-12
Switch to gmavenplus to build groovy

bitwiseman:
* JENKINS issue(s):
    * N/A
* Description:
    * Based on a comment from @jglick I figured I'd see if gmaven-plus could work for us (and unblock java 11 build).  Probably not, but can't hurt to check.
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
@@ -672,11 +672,11 @@ class RuntimeASTTransformer {
         return constX(null)
     }
 
-    private Expression valueOrNull(@CheckForNull ModelASTValue value, Object defaultValue = null) {
+    private Expression valueOrNull(@CheckForNull ModelASTValue value) {
         if (value?.sourceLocation instanceof Expression) {
             return (Expression)value.sourceLocation
         } else {
-            return constX(defaultValue)
+            return constX(null)
         }
     }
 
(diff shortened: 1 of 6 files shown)


jglick:
> * Based on a comment from @jglick

Presumably https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/335#issuecomment-500918508 or similar.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -672,11 +672,11 @@ class RuntimeASTTransformer {
         return constX(null)
     }
 
-    private Expression valueOrNull(@CheckForNull ModelASTValue value, Object defaultValue = null) {
```
jglick: This would be dropping the argument passed at line 666 above:

```groovy
valueOrNull(original.id, stageName),
```

jglick: Hence test failures like

```
groovy.lang.MissingMethodException: No signature of method: org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.valueOrNull() is applicable for argument types: (org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTValue$ConstantValue, java.lang.String) values: [ModelASTValue{value=simple-input, isLiteral=true}, One]
```

bitwiseman:
@jenkinsci/java11-support @oleg-nenashev 
In #335 @jglick mentioned issues with the custom version of gmaven used by the Jenkins project build. 

This change switches to a publicly available tool called gmavenplus.  It was a very straightforward move. 
See #340 for an additional change that was needed to disable gmaven execution inherited from the parent-pom. 

I've opened an issue to track adding this plugin to the parent-pom and removing gmaven: https://issues.jenkins-ci.org/browse/JENKINS-58028

PR #343, opened 2019-07-21
Consistently use sandbox for CpsFlowDefinitions

basil:
While looking at the test suite for this plugin, I noticed the `CpsFlowDefinition`s in the tests don't consistently use the script security sandbox. Using the script security sandbox results in a more realistic environment given that the script security sandbox should always be enabled in production.

In this change, I replaced any usages of the deprecated single-argument constructor for `CpsFlowDefinition` with usages of the non-deprecated two-argument constructor, passing in `true` as the second argument in order to always enable the script security sandbox.


bitwiseman:
@basil 
Could you amend and repush your last commit?  This looks like an infrastructure failure.

basil:
@bitwiseman The run is green.

PR #347, opened 2019-08-05
Adding validation and tests

bitwiseman:
Adding validation and tests for matrix


abayer:
Whole lotta test failures with errors like:
```
java.lang.NullPointerException: Cannot invoke method addAll() on null object
at org.codehaus.groovy.runtime.NullObject.invokeMethod(NullObject.java:91)
at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:48)
at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
at org.codehaus.groovy.runtime.callsite.NullCallSite.call(NullCallSite.java:35)
at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
at org.jenkinsci.plugins.pipeline.modeldefinition.parser.ModelParser$_parseExcludeAxis_closure17.doCall(ModelParser.groovy:491)
```

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAxis.java
```diff
@@ -19,7 +19,7 @@
 public class ModelASTAxis extends ModelASTElement {
 
     private ModelASTKey name;
-    private List<ModelASTValue> values = new ArrayList<>();
+    private List<ModelASTValue> values;
```
abayer: So this change is causing test failures - I far prefer always initing collections to empty rather than null, to avoid this kind of problem (and to avoid needing to bake in initing to empty in multiple places instead). What's the reasoning for this change?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAxisContainer.java
```diff
@@ -30,6 +29,7 @@ public JSONArray toJSON() {
     @Override
     public void validate(@Nonnull ModelValidator validator) {
         // nothing to validate on this element
```
abayer: Not exactly true any more. =) Same thing on `ModelASTExclude` and `ModelASTExcludes`.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
```diff
@@ -259,16 +259,16 @@ class ModelParser implements Parser {
         r.validate(validator)
 
         // Lazily evaluate r.toJSON() - i.e., only if AST_DEBUG_LOGGING is true.
-        astDebugLog {xf
-            "Model as JSON: ${r.toJSON().toString(2)}"
+        astDebugLog {
+            "Model as JSON:\\n${r.toJSON().toString(2)}"
```
abayer: This doesn't end up just printing out `\n`?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -690,15 +690,12 @@ class ModelValidatorImpl implements ModelValidator {
             stepsStagesParallelCount += 1
         }
 
-        if (isWithinParallel && (stage.branches.size() > 1 || stage.parallel != null)) {
+        if (isWithinParallel && (stage.branches.size() > 1 || stage.parallel != null || stage.matrix != null)) {
```
abayer: At this point, it may be worth adding a method to `ModelASTStage` that does the `(stage.branches.size() > 1 || ... )` and just calling that here.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -733,41 +730,122 @@ class ModelValidatorImpl implements ModelValidator {
...
+                valid = false
+            }
         }
 
         return validateFromContributors(stages, valid)
     }
 
     boolean validateElement(@Nonnull ModelASTParallel parallel) {
-        return true;
+        return validateElement((ModelASTStages)parallel)
```
abayer: I always forget how casting and method signatures work - I assume this does what it's meant to since otherwise you'd end up with `StackOverflowError`s. =)

bitwiseman: Yes. :)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
bitwiseman: Yeah, some clearer behavior....

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTAxis.java
bitwiseman: I was attempting to be able to know when `values` was missing vs when it has no args passed, but It is definitely more trouble than it is worth.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
bitwiseman: I tried it and it doesn't make it clearer... I'm going to leave it for now.

PR #352, opened 2019-08-29
Switch Matrix to be based on stage

bitwiseman:
This is not great work, but it is sufficient to start.

* JENKINS issue(s):
    * Issue link or n/a
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
```diff
@@ -72,26 +74,27 @@ class Stage implements Serializable {
...
           Environment environment, boolean failFast, StageOptions options, StageInput input,
-          Stages stages, Parallel parallel, Matrix matrix) {
+          Stages stages, Parallel parallel, Matrix matrix, Environment preAgentEnvironment) {
         this.name = name
         this.agent = agent
         this.post = post
         this.when = when
         this.tools = tools
         this.environment = environment
+        this.preAgentEnvironment = preAgentEnvironment
```
bitwiseman: @abayer @dwnusbaum  
JENKINS-43911 coming back to try to thwart my dreams - this is my response.  
Without this the matrix cell values are not available at the time that agent is resolved.  :sob:

This functionality is only in the runtime for now.  If you have a better naming suggestion, we might consider exposing this a syntax change in a future change.  For now though, this get all values able to resolve based on the matrix cell values.

abayer: So definitely no syntax support - there are too many things that could go weird and it would be verrrrrrry confusing for users to know what to use when, etc.

abayer:
So some initial thoughts:
* Enabling this in syntax still wouldn't work without further refactoring because we expect a string for the values in `agent`, not something we process later, so some functionality might explode.
* Might it just be simpler to handle the environment twice, in this special case? You'll get some bogus values (like anything referencing `$WORKSPACE`) on the first pass, but the second pass should fix them. We'd also need to look at the `credentials` and remove anything that needs to write to disk (i.e., `FileCredentials` and the like).

bitwiseman:
> * Enabling this in syntax still wouldn't work without further refactoring because we expect a string for the values in `agent`, not something we process later, so some functionality might explode.

We don't require string values in agent.  If an env value is provided from the containing stage or elsewhere it works. 

> * Might it just be simpler to handle the environment twice, in this special case? You'll get some bogus values (like anything referencing `$WORKSPACE`) on the first pass, but the second pass should fix them. We'd also need to look at the `credentials` and remove anything that needs to write to disk (i.e., `FileCredentials` and the like).

Can we process the env twice? I thought that might throw an exception on missing values in some cases.

abayer:
The values should just show up as empty string, I think? And I honestly thought that `agent` looked for a groovy string literal at parse time, which would cause problems if you used anything else (like `env.FOO` or `"${env.FOO}" in parsing.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
```diff
@@ -98,7 +110,7 @@ class Stage implements Serializable {
         this.input = input
         this.stages = stages
         this.parallel = parallel
-        this.matrix = matrix
```
bitwiseman: The matrix field on runtime was serving no purpose but making code more complex.  `Matrix` extends `Parallel` and at runtime they are equivalent.

PR #353, opened 2019-09-04
Added new when conditions stashed and archived

rsandell:
* JENKINS issue(s):
    * N/A
* Description:
    * Two new when conditions to be able to check if a previous stage stashed or archived something.
* Documentation changes:
    I was originally planning to add these to the `declarative-pipeline-when-conditions-plugin` since these conditions might be a bit niche. But after an hour of fixing "require upper bounds" issues in the pom I found out that the API for allowing other plugins to contribute when conditions wasn't "fixed" yet.
* Users/aliases to notify:
    * @abayer

### DIFF
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -265,5 +265,6 @@
 
   <properties>
     <no-test-jar>false</no-test-jar>
+    <useBeta>true</useBeta>
   </properties>
 </project>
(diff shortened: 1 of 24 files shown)


rsandell:
Tickling the PR builder.

rsandell:
@bitwiseman ping?

bitwiseman:
@rsandell Sorry, distracted by other work.  Looks like this still needs to pass CI. Trying again.

pipeline-model-definition/pom.xml
```diff
@@ -265,5 +265,6 @@
 
   <properties>
     <no-test-jar>false</no-test-jar>
+    <useBeta>true</useBeta>
```
bitwiseman: @rsandell  What is this setting for?

rsandell:
Docs, right :)
It was a long time since I did this so need to find some time to brush up on it.

bitwiseman:
@rsandell Do you have time to do docs for this, or would you like me to pick this up?

PR #355, opened 2019-10-11
Matrix/method size

bitwiseman:
* JENKINS issue(s):
    * Issue link or n/a
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Matrix.groovy
```diff
@@ -40,7 +40,7 @@ import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted
 class Matrix extends Parallel {
 
     @Whitelisted
-    Matrix(List<Stage> stages) {
```
bitwiseman: Since there is a limit on ListExpression size, we pass a list of lists and then concatenate.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -50,16 +57,94 @@ import javax.annotation.Nonnull
...
+    def constLookup = [:]
+//    def variableClassNode = null
+//    def variableValues = [:]
+
+
+    RuntimeASTTransformer(ModuleNode moduleNode) {
+        this.moduleNode = moduleNode
+    }
+
+    Expression ctorXFunction(ClassNode type, Expression args) {
```
bitwiseman: Helper function that turns a ConstructorExpression into a MethodCall to a generated static method that returns that ConstructorExpression.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -50,16 +57,94 @@ import javax.annotation.Nonnull
...
+                block(
+                        returnS(ctorX(type, args))
+                )
+
+        MethodNode method = new MethodNode(name, ACC_PUBLIC | ACC_STATIC, ClassHelper.make(Stage.class), [] as Parameter[], [] as ClassNode[], methodBody)
+        classNode.addMethod(method)
+        return callX(classNode, name)
     }
 
+    Expression constXString(String value) {
```
bitwiseman: WIP trying to do further optimization.  In progress.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/model/Stage.groovy
```diff
@@ -157,4 +157,6 @@ class Stage implements Serializable {
             return [:]
         }
     }
+
```
abayer: extra whitespace. =)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -519,7 +605,7 @@ class RuntimeASTTransformer {
             return ctorX(ClassHelper.make(Libraries.class), args(listArg))
         }
 
-        return constX(null)
+        return ConstantExpression.NULL
```
bitwiseman: This was one idea , to reduce the number of constants, but it hasn't helped.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -776,15 +863,32 @@ class RuntimeASTTransformer {
             filterExcludes(expansion, original.excludes)
 
             // for each combination
+
+            def listLimit = 250
+            if (expansion.size() > listLimit * listLimit) {
+                throw new IllegalArgumentException("Matrix supports up to ${listLimit} cells. Found ${expansion.size()}.")
+            }
+
+            ListExpression expandedMatrixStages = new ListExpression()
```
bitwiseman: @kshultzCB 
This fixes the case of matrixes with more than 250 cells.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -50,16 +57,94 @@ import javax.annotation.Nonnull
...
+                methodCount[typeName] = 0
+                methodClassNode[typeName] = new ClassNode("DeclarativeRuntime_${typeName}_${Math.abs(type.hashCode())}", ACC_PUBLIC, ClassHelper.make(Object))
+                this.moduleNode.addClass(methodClassNode[typeName])
+            }
+        }
+
+        ClassNode classNode = methodClassNode[typeName]
+
+        String name = "${typeName}_${methodCount[typeName]++}"
+        while (classNode.methods.find {it.name == name }) {
```
abayer: Took me a bit to figure out what this is doing - if I'm understanding correctly, it's trying various `Foo_1`, `Foo_2`, etc methods to see if it's already added one to settle on a good name? A comment explaining that would be worthwhile.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
abayer: This is some wonderfully demented magic. I approve.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
bitwiseman: Simplified in latest changes.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -837,7 +899,7 @@ class RuntimeASTTransformer {
             // TODO: Do I need to create a new ModelASTStage each time?  I don't think so.
             String name = "Matrix - " + cellLabels.join(", ")
 
-            return ctorX(ClassHelper.make(Stage.class),
+            return ctorXFunction(ClassHelper.make(Stage.class),
```
dwnusbaum: I wonder how this affects variable resolution for stuff inside of the matrix, for example if someone defines a Groovy method at the top of the Jenkinsfile (not recommended, but it happens), now that the matrix stages are generated inside of a static method in a separate class, will those methods still be resolvable?

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -579,7 +579,7 @@ class ModelInterpreter implements Serializable {
         toolsList.each { l ->
             String k = l.get(0)
             Closure v = (Closure)l.get(1)
-            String toolVer = v.call()
+            String toolVer = delegateAndExecute(v)
```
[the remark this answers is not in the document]
bitwiseman: Added because of test failures.  Tried removing after latest changes, tests fail

This wasn't really required before this change because the `closure.owner` would be tried and `closure.owner` was the next outer closure which masked is bug. 
Now that the `owner` isn't the next closure out, delegate needs to be set properly.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
bitwiseman: Good question. Tried it manually and it worked. 

@abayer  
Is there a test for this scenario? (The only down side to extensive testing it not know where all the tests are.)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/RuntimeContainerBase.groovy
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ *
+ *
+ * @author Liam Newman
+ */
+@SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
+class RuntimeContainerBase {
+
+    protected static CpsScript that
```
dwnusbaum: Spotbugs wants you to restrict the visibility of this member (not sure if it is looking at Groovy directly or at the Java code generated from the Groovy code).

> RuntimeContainerBase.that should be package protected [org.jenkinsci.plugins.pipeline.modeldefinition.RuntimeContainerBase] In RuntimeContainerBase.groovy MS_PKGPROTECT

It looks like it is only used in this file and could be private.

```suggestion
    private static CpsScript that
```

bitwiseman: Got it.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -135,7 +135,7 @@ class Utils {
     }
 
 
-    // Note that we're not using StringUtils.strip(s, "'\"") here because we want to make sure we only get rid of
+    // Note workflowScript we're not using StringUtils.strip(s, "'\"") here because we want to make sure we only get rid of
```
dwnusbaum: I think you replaced `that` with `workflowScript` a bit too liberally 😄

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -181,7 +302,7 @@ class RuntimeASTTransformer {
      * @param valueType Either {@link ModelASTInternalFunctionCall} for credentials or {@link ModelASTValue} for env.
      * @return The AST for instantiating the resolver.
      */
-    private Expression generateEnvironmentResolver(@Nonnull Map<ModelASTKey, ModelASTEnvironmentValue> variables, @Nonnull Class valueType) {
+    private Expression generateEnvironmentResolver(@Nonnull Map<ModelASTKey, ModelASTEnvironmentValue> variables, @Nonnull Class valueType, boolean disableWrapping = false) {
```
dwnusbaum: SpotBugs doesn't like that when `RuntimeASTTransformer` is converted to Java, a version of this method without `disableWrapping` is generated (that's how default parameters are implemented by Groovy), which is never called:

> Private method org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.generateEnvironmentResolver(Map, Class) is never called [org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer] In RuntimeASTTransformer.groovy UPM_UNCALLED_PRIVATE_METHOD

If `disableWrapping` is always specified, and there's not concern about backwards compat, then you can just drop the default expression `= false`, otherwise just suppress the warning (or convert the class to Java since it's in `src/main` anyway, but that's more work).

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
bitwiseman: There are tests for this and they caught the problem but I'm not sure that they cover all the cases.

dwnusbaum: FWIW, I think it would be a good idea to add a system property/public static non-final variable to disable the new behavior and force the old code path in case it causes issues for existing Declarative users in scenarios that aren't covered by tests. Something like this in `RuntimeASTTransformer`:

```
@SuppressFBWarnings(warningId="the real code for the non-final static vars are bad warning", justification="For access from script console")
public static boolean FORCE_SINGLE_EXPRESSION_TRANSFORMATION = Boolean.getBoolean(RuntimeASTTransformer.class.getName() + ".FORCE_SINGLE_EXPRESSION_TRANSFORMATION");
```

And then inside of `ctorXFunction` and friends you look at `FORCE_SINGLE_EXPRESSION_TRANSFORMATION` (or whatever you want to call it) and fall back to `ctorX` if it is true.

bitwiseman: @dwnusbaum 
Solid idea. Done.

jbennett2091:
Thanks for all the hard work on this.  We really need this fix.  Hope it gets released soon.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/RuntimeContainerBase.java
```diff
@@ -0,0 +1,89 @@
...
+ * Examples:
+ * <ul>
+ * <li> this class must inherit the sleep method from CpsScript (or implement the same behavior itself).
+ * <li> script step resolution will be unable to find globals
+ * <li> Groovy's getProperty is not whitelisted by default except for on CpsScript
+ * </ul>
+ *
+ * @author Liam Newman
+ */
+public class RuntimeContainerBase extends CpsScript {
```
jglick: This…scares me.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/RuntimeContainerBase.java
```diff
@@ -0,0 +1,89 @@
...
+    @Whitelisted
+    public RuntimeContainerBase(@Nonnull CpsScript workflowScript) throws IOException {
+        super();
+        this.workflowScript = workflowScript;
+        this.setBinding(workflowScript.getBinding());
+    }
+
+    @Override
+    public Object run() {
+        return null;
```
jglick:
```suggestion
        throw new AssertionError();
```

?

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -932,8 +997,375 @@ class RuntimeASTTransformer {
...
+            // Local variable in groovy ast is a variable that ... accesses itself.
+            VariableExpression variable = varX(createStableUniqueName('extendedList'))
+            variable.accessedVariable = variable
+
+            Expression declarationExpression = new DeclarationExpression(variable, ASSIGN, new ListExpression())
+            BlockStatement block = block()
+            block.addStatement(stmt(declarationExpression))
+
+            int count = 0
+            expressions.each { item ->
```
jglick: FWIW I have long advocated that `src/main/groovy/` here be rewritten to Java for maintainability. (`src/main/resources/**/*.groovy` is unrelated—that is CPS-transformed.)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/RuntimeContainerBase.java
bitwiseman: Indeed.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
bitwiseman: I don't understand what that has to do with this line of code.

jglick: Not this line specifically, just the whole long blocks of new Groovy.

jglick:
If memory serves right, at some point @abayer switched from _interpreting_ the Declarative structure, to generating a class which would run it, in order to solve some thorny issues with variable resolution? And that decision then led to the problem being solved here?

The subclass of `CpsScript` makes me very uncomfortable. Is there really no other way to move code into other classes? You tried just using `evaluate`?

bitwiseman:
@jglick 
I'm not sure. Despite what this looks like I did not try to reinvent the wheel. I felt sticking close to the existing patterns would be better. 

How would `evaluate` help here?

bitwiseman:
@jglick I see your point about `CpsScript` subclass. It's not ideal and it makes me somewhat nervous as well.  On the other hand, what exactly does it open up that isn't already open?

jglick:
> at some point @abayer switched…

This was #174, for reference.

> How would `evaluate` help here?

Bear in mind that I only dimly understand what this PR is even doing, or how the original code works for that matter. My understanding is that (hand-waiving here) you are taking a Declarative script and transforming it into an actual executable program, and that in this PR some of the code blocks are being moved to separate JVM classes without otherwise changing behavior. So the straightforward way to do something like that in Pipeline script (without making any reference to `workflow-cps` internals, or even exiting the sandbox) would be to replace

```groovy
def somethingLong() {
  // dozens of lines here
}
somethingLong()
```

with, say,

```groovy
def somethingLong = evaluate '''
{ ->
  // dozens of lines here
}
'''
somethingLong()
```

so that the closure gets compiled into a separate class, something like `Script19` which will in fact be a subtype of `CpsScript`; see `CpsGroovyShellFactory.makeConfig`.

(Admittedly the above still makes for a long constant entry, but in the actual problem here, the “dozens of lines” would be computed at runtime somehow.)

> what exactly does it open up that isn't already open?

It is hard to say, but by directly using this type from `workflow-cps` which was never intended for outside consumption, you are entering dangerous territory.

I note that none of the newly introduced tests seem to use `RestartableJenkinsRule`, so it is not even clear that the complicated script persistence mechanisms in `workflow-cps` will be engaged.

I understand that there are plenty of votes for this issue, but a lot of them presumably apply to users of Scripted, who will not be helped by this. Are there really so many people with massive Declarative scripts that trigger the limit to justify such a high-risk change?

jglick:
On the other hand, this plugin is already horribly intertwined with `workflow-cps` internals it seems. My real recommendation at a high level is that the current execution mode of Declarative be pretty much frozen and deprecated; and that a separate mode be introduced which runs without low-level access to `workflow-cps`, using an interpreter rather than code generation (like a very simplified variant of the plugin prior to #174), but which does not support `script` blocks or any code outside `pipeline`, libraries, or any `GString` interpolation beyond simple variable substitution.

bitwiseman:
> Are there really so many people with massive Declarative scripts that trigger the limit to justify such a high-risk change?

The way this was found was during `matrix` development.  In that scenario, yes, it is easy to create declarative pipelines that encounter this.

bitwiseman:
@jglick 
A question for you:
I was able to get this working without extending `CpsScript`.  Jenkins infra is borked, but I've tested it locally.

However, on reviewing with @dwnusbaum I ran into another issue.  When pausing the generated classes are serialized.  When they are deserialized, I don't believe they retain the reference to the current `WorkflowScript`.  I'm still attempting to verify that - the classes are currently only called very early in the run, basically as the Declarative pipeline run starts.   So, the current pause restart test works and I've also verified manually with some more involved scenarios that restart works, but it could be 

My question: is there an event or other api that I can hook into get a reference to the current instance of `WorkflowScript`?  I could have every generated method on the generated classes pass a reference to it, but ... ew.

jglick:
> In that scenario, yes, it is easy to create declarative pipelines that encounter this.

Because of the exponential combination of axes I suppose? This plugin is generating code (like a `stage`) for each combination?

> is there an event or other api that I can hook into get a reference to the current instance of `WorkflowScript`?

Not that I know of. Normally if a class needs a reference to the root script (and it might not), it is either held as a strong reference and thus serialized as part of the program state (recall that Java serialization, and JBoss too, handles cycles), or passed in as an argument to particular methods. IIRC both are documented idioms when using `src/` in libraries.

Bear in mind that I can just barely follow how program resumption works in Scripted when using nothing out of the ordinary (for example, only stock sandbox calls and steps). If you are talking about programmatically generating classes, I cannot make any promises.

dwnusbaum:
> it is either held as a strong reference and thus serialized as part of the program state (recall that Java serialization, and JBoss too, handles cycles)

Oops, forgot that cycles are handled, sorry for the misleading info yesterday @bitwiseman. I guess in that case, going by the in-progress code you showed me yesterday, I would make the field that holds the script non-transient again and see if that works (not sure if you are saying you already tried that and it didn't work, or something else).

bitwiseman:
> Because of the exponential combination of axes I suppose? This plugin is generating code (like a stage) for each combination?

Well, it does when this change is not enabled. With this it can reuse items from variables. 

> I would make the field that holds the script non-transient again and see if that works (not sure if you are saying you already tried that and it didn't work, or something else).

Ah, of, course, why didn't I think of that.  Yay!

bitwiseman:
Built and tested locally:
```
[INFO] Results:
[INFO]
[WARNING] Tests run: 1418, Failures: 0, Errors: 0, Skipped: 19
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -766,30 +815,39 @@ class RuntimeASTTransformer {
...
             filterExcludes(expansion, original.excludes)
 
-            // for each combination
+            Expression stagesExpression = transformStages(original.stages)
+
+            // With script splitting stagesExpression is a method that creates a new stages expression at runtime
+            // If disabled, we still want to make this a script closure call so we can reuse it instead of generating code for every cell
+            if (!SCRIPT_SPLITTING_TRANSFORMATION) {
+                stagesExpression = wrapper.asWrappedScriptContextVariable(stagesExpression, true)
+            }
```
bitwiseman: @jglick  Per your comment, this forces the use of a single expression (method or a closure), which is then called for each cell instead of generating code for each cell.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/RuntimeContainerBase.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Base class for classes generated by {@link RuntimeASTTransformer.Wrapper} when code splitting transform is enabled.
+ * This class is as close a possible to a transparent pass-through to CpsScript.
+ * However, this class does not get the special whitelisting granted to CpsScript instances, so
+ * care should be be taken regarding what code is added to these classes' methods.
+ *
+ * @author Liam Newman
+ */
+public class RuntimeContainerBase extends Script implements Serializable {
```
bitwiseman: @jglick @dwnusbaum 
Now extends `Script` instead of `CpsScript`.  Yay!

PR #359, opened 2019-10-31
Disable matrix execution

bitwiseman:
* Description:

To get a 1.3.x release out without turning on matrix, I've set it to fail at runtime.
This is simpler than trying to tease apart what's been added in the last few months.

@dwnusbaum @olamy @kshultzCB @joseblas


bitwiseman:
@dwnusbaum No that is a fair way to go and clearly indicates the level of risk.

PR #362, opened 2019-11-14
[FIXES JENKINS-49500] Prevent triggers from being restarted if same

scoheb:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-49500
* Description:
    * Prevent triggers from being restarted if they have not changed between builds.

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -35,6 +35,7 @@ import hudson.model.*
 import hudson.triggers.Trigger
 import jenkins.model.Jenkins
 import org.apache.commons.codec.digest.DigestUtils
+import org.apache.commons.collections.CollectionUtils
 import org.codehaus.groovy.ast.ASTNode
 import org.codehaus.groovy.ast.stmt.BlockStatement
 import org.codehaus.groovy.ast.stmt.Statement
@@ -630,17 +631,24 @@ class Utils {
                 }
             }
 
-            // Remove the triggers/parameters properties regardless.
-            j.removeProperty(PipelineTriggersJobProperty.class)
+            def currentTriggerProp = j.getProperty(PipelineTriggersJobProperty.class)
+            def newTriggerProp = new PipelineTriggersJobProperty(triggersToApply)
+            // Only Remove/Add if the triggers are different.
+            if (!areTriggersInPipelineTriggerJobPropertiesEqual(currentTriggerProp, newTriggerProp)) {
+                j.removeProperty(PipelineTriggersJobProperty.class)
+                if (!triggersToApply.isEmpty()) {
+                    j.addProperty(newTriggerProp)
+                }
+            }
+
+
+            // Remove the parameters properties regardless.
             j.removeProperty(ParametersDefinitionProperty.class)
 
             // Remove the job properties we defined in previous Jenkinsfiles but don't any more.
             propsToRemove.each { j.removeProperty(it) }
 
-            // If there are any triggers and if there are any parameters, add those properties.
-            if (!triggersToApply.isEmpty()) {
-                j.addProperty(new PipelineTriggersJobProperty(triggersToApply))
-            }
+            // if there are any parameters, add those properties.
             if (!parametersToApply.isEmpty()) {
                 j.addProperty(new ParametersDefinitionProperty(parametersToApply))
             }
@@ -661,6 +669,14 @@ class Utils {
         } finally {
             bc.abort()
         }
+    }
+
+    // Utility method to determine is triggers in PipelineTriggersJobProperties are equal
+    static boolean areTriggersInPipelineTriggerJobPropertiesEqual(PipelineTriggersJobProperty left, PipelineTriggersJobProperty right) {
+        if (left == right) return true;
+        if (left == null || right == null) return false;
+        boolean triggersEqual = CollectionUtils.isEqualCollection(left.triggers, right.triggers);
+        return triggersEqual;
     }
 
     /**
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/TriggersTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/TriggersTest.java
@@ -24,9 +24,14 @@
 
 package org.jenkinsci.plugins.pipeline.modeldefinition;
 
+import hudson.Extension;
+import hudson.model.Item;
 import hudson.triggers.SCMTrigger;
 import hudson.triggers.TimerTrigger;
 import hudson.triggers.Trigger;
+import hudson.triggers.TriggerDescriptor;
+import jenkins.model.Jenkins;
+import org.jenkinsci.Symbol;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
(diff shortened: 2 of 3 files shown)


scoheb:
@abayer for your review.

bitwiseman:
@scoheb  I'll probably be the one to review this.  Could you give a little more details about this?

scoheb:
@bitwiseman I will describe my use case for this....

We use the jms-messaging plugin for our triggers for our pipelines. The plugin creates a thread that subscribes to a topic on a message broker. If the declarative pipeline contains a `triggers` section, then every single time a build starts up, the trigger is blindly stopped/started (even though no changes were made to the trigger configuration). Since we use a thread to monitor for new messages, then we are constantly and needlessly destroying/creating threads and there is a risk that we even miss a message since we might not be subscribed since we are stopping/starting.  

This PR attempts to detect that the trigger has not been changed and thus it prevents the stop/start that can be disruptive.

hope that helps clarify the issue!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -630,17 +631,24 @@ class Utils {
                 }
             }
 
-            // Remove the triggers/parameters properties regardless.
-            j.removeProperty(PipelineTriggersJobProperty.class)
+            def currentTriggerProp = j.getProperty(PipelineTriggersJobProperty.class)
```
bitwiseman: All the triggers are in one property, but there can be multiple, so plural. 
```suggestion
            def currentTriggersProp = j.getProperty(PipelineTriggersJobProperty.class)
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -630,17 +631,24 @@ class Utils {
...
+            def newTriggerProp = new PipelineTriggersJobProperty(triggersToApply)
```
bitwiseman:
```suggestion
            def newTriggersProp = new PipelineTriggersJobProperty(triggersToApply)
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -630,17 +631,24 @@ class Utils {
...
+            // Only Remove/Add if the triggers are different.
+            if (!areTriggersInPipelineTriggerJobPropertiesEqual(currentTriggerProp, newTriggerProp)) {
```
bitwiseman:
```suggestion
            if (!areTriggersInPipelineTriggerJobPropertiesEqual(currentTriggersProp, newTriggersProp)) {
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -630,17 +631,24 @@ class Utils {
...
+                j.removeProperty(PipelineTriggersJobProperty.class)
+                if (!triggersToApply.isEmpty()) {
+                    j.addProperty(newTriggerProp)
```
bitwiseman:
```suggestion
                    j.addProperty(newTriggersProp)
```

Also, this replaces all the triggers even if only one has changed, right?
It would be better if made it so only the _changed_ triggers are replaced.  But that might be work for another PR.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -663,6 +671,14 @@ class Utils {
         }
     }
 
+    // Utility method to determine is triggers in PipelineTriggersJobProperties are equal
+    static boolean areTriggersInPipelineTriggerJobPropertiesEqual(PipelineTriggersJobProperty left, PipelineTriggersJobProperty right) {
+        if (left == right) return true;
+        if (left == null || right == null) return false;
+        boolean triggersEqual = CollectionUtils.isEqualCollection(left.triggers, right.triggers);
```
bitwiseman: Is this comparison unordered?  Will it return true even if only the order of the triggers has changed?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/TriggersTest.java
```diff
@@ -162,4 +176,113 @@ public void actualTriggerCorrectScope() throws Exception {
         expect("actualTriggerCorrectScope")
                 .go();
     }
+
+    @LocalData
+    @Test
+    public void doNotRestartEqualTriggers() throws Exception {
```
bitwiseman: This is good. 
We also need a test to show this with two triggers and that if the order of the triggers changes (but they are otherwise the same) that this still works and does not recreate the triggers.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
gulyaev13: @bitwiseman , @scoheb , sorry for interrupting your review, but I think this will not work.
`CollectionUtils.isEqualCollection()` use hashCode for comparing objects, but Trigger and child classes don't override this method, so it will not work.

But the problem is more complex, than the problem with Triggers. 
The whole method `org.jenkinsci.plugins.pipeline.modeldefinitionUtils.updateJobProperties` should be refactored, because it isn't thread-safe (I can be one of the reasons of [JENKINS-44809](https://issues.jenkins-ci.org/browse/JENKINS-44809)) and affect performance in case of parallel pipeline execution (save configuration each time, even if it has not been changed).

I'm working on a solution, which will resolve performance problems.
I want to add some hints, based on specifics of tipis of JobProperties. 
But exactly for triggers, I can't find a solution better than comparing Strings after XSTREAM serialization. Maybe you can advise me something before PR?

bitwiseman: @gulyaev13 
Interesting. Given what you're saying with hash code I wonder how the test succeeded? 

As to multi-threading - that would mean the the current running implementation is also not thread-safe.  Is what you have seen and are trying to fix?

gulyaev13: @bitwiseman , there is no magic. TestTrigger defined in the test has overwritten methods hashCode() and equals().

Yes, the current implementation isn't seen thread-safe, because of method updateJobProperties change global object WorkflowJob without any sync. 
I've seen the effects of concurrent changing a few times ago. The job's parameters were deleted from one thread and were saved to disk by another thread.
The main problem which I'm trying to solve is reducing a lot of IO calls caused by saving same configuration each job run (Save action is thread-save and become a bottleneck of concurrent pipeline execution).

scoheb: @gulyaev13 @bitwiseman I think the PR still improves the situation for triggers that do implement `equals()` and `hashcode()`

gulyaev13: @scoheb, I think I'm ready to send my PR which will solve your problem, too!
Am I add your test to my PR?

scoheb: @gulyaev13 of course you can use my test! thanks! looking forward to your PR!

bitwiseman:
@scoheb @gulyaev13 
Please coordinate your efforts here if you can.  @scoheb I assume you'll want to close this change and use #365 ?  Would you be willing to write some more test for #365, perhaps?

scoheb:
@bitwiseman yes, I will close this in favour of #365

PR #363, opened 2019-11-14
JENKINS-60115: Make toGroovy() and toJSON() @Nonnull

bitwiseman:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-60115
* Description:
    * It not really a surprise that there is a code path that produces `null` from some AST element `toGroovy()`, but it really shouldn't be possible.
* Documentation changes:
    *  NA
* Users/aliases to notify:
    * None yet.

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTSingleArgument.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTSingleArgument.java
@@ -21,6 +21,7 @@ public final class ModelASTSingleArgument extends ModelASTArgumentList {
     }
 
     @Override
+    @Nonnull
     public Object toJSON() {
         return toJSON(value);
     }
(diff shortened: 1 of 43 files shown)


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTSingleArgument.java
```diff
@@ -14,15 +14,21 @@
  * @author Andrew Bayer
  */
 public final class ModelASTSingleArgument extends ModelASTArgumentList {
+
+    /**
+     * While not @Nonnull, if this field is null parsing or validation errors will occur before
```
dwnusbaum: I'm not sure how the Javadoc parser interprets `@Nonnull` literally, might cause errors depending on how strict it is (I think it's ok as long as it doesn't start a line, so it isn't seen as an unrecognized 
 Javadoc tag, but I'm not sure). If Javadoc is generated during the CI build here, then it's fine, but if it is only generated during releases, it could cause a problem.

PR #364, opened 2019-11-15
Update plugins.jenkins.io GitHub links

MarkEWaite:
# Fix plugins.jenkins.io GitHub links

* JENKINS issue(s):
    * No Jira issue
* Description:
    * Links to GitHub from the plugins.jenkins.io site are broken
* Documentation changes:
    * Documentation updated

### DIFF
--- a/pipeline-stage-tags-metadata/pom.xml
+++ b/pipeline-stage-tags-metadata/pom.xml
@@ -36,6 +36,12 @@
   <description>Library plugin for Pipeline stage tag metadata</description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Model+Definition+Plugin</url>
 
+  <scm>
+    <connection>scm:git:https://github.com/jenkinsci/pipeline-model-definition-plugin</connection>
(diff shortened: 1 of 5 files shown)


pipeline-stage-tags-metadata/pom.xml
```diff
@@ -36,6 +36,12 @@
   <description>Library plugin for Pipeline stage tag metadata</description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Model+Definition+Plugin</url>
 
+  <scm>
```
bitwiseman: @MarkEWaite Aren't these inherited?

MarkEWaite: They are inherited but the inherited values are wrong.  The effective pom (from `mvn help_effective-pom`) in the nested modules include an incorrect value for `scm.url`.  If the parent repository is https://github.com/pipeline with a module named my-module then the module scm.url will be https://github.com/pipeline/my-module, even though it should be https://github.com/pipeline

bitwiseman: 😭

PR #365, opened 2019-11-18
Update only changed Job Properties

gulyaev13:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-49500
    * https://issues.jenkins-ci.org/browse/JENKINS-44809
* Description:
    I've found a problem, that concurrent execution of the same pipeline produces a bunch of problems:
     - Jenkins saves job configuration each run, even if the configuration isn't changed. Save to the same file is blocking operation, so all parallel runs stuck for wait IO. In case of installed Job Config History Plugin IO calls multiplied 2 times due to reordering of some JobProperties during marshaling. Also, rewriting all job properties stops not changed triggers (JENKINS-49500)
     - Updating of WorkflowJob object in method `Utils.updateJobProperties()` isn't thread-safe, so it produces undefined behavior. It can't be a root cause of JENKINS-44809. 
     - One more seen issue. One thread deletes parameters, but another save this state to file (BulkChange prevent save to disk before commit only for one thread). The next run hadn't contained parameters. If a job doesn't define parameters in Jenkinsfile, the parameters will not be restored at all.

* Users/aliases to notify:
    * @abayer , @bitwiseman , @scoheb, for your  #review

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -826,14 +826,19 @@ class Utils {
     private static List<ParameterDefinition> getParametersToApply(@CheckForNull List<ParameterDefinition> newParameters,
                                                                   @Nonnull List<ParameterDefinition> existingParameters,
                                                                   @Nonnull Set<String> prevDefined) {
-        //Store parameters with unique name
-        Set<ParameterDefinition> toApply = new TreeSet<>(Comparator.comparing({ ParameterDefinition p -> p.name }))
-        toApply.addAll(existingParameters)
-        //Remove all parameters defined in previous Jenkinsfile and may be deprecated
-        toApply.removeAll { it.name in prevDefined }
-        //Replace all parameters defined in current Jenkinsfile
-        toApply.addAll(newParameters)
-        return toApply.asList()
+        Set<String> seenNames = new HashSet<>()
+        List<ParameterDefinition> toApply = []
+        if (newParameters != null) {
+            toApply.addAll(newParameters)
+            seenNames.addAll(newParameters.collect { it.name })
+        }
+        // Find all existing parameters that aren't of names we've explicitly defined, *and* aren't
+        // in the set of names of parameters defined by the Jenkinsfile in the previous build. Add those too.
+        toApply.addAll(existingParameters.findAll {
+            !(it.name in seenNames) && !(it.name in prevDefined)
+        })
+
+        return toApply
     }
 
     /**
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -562,9 +562,9 @@ class Utils {
         List<ParameterDefinition> existingParameters = existingParametersForJob(j)
 
         Set<String> previousProperties = new HashSet<>()
-        Set<String> previousTriggers = new HashSet<>()
+        Set<String> previousTriggers   = new HashSet<>()
         Set<String> previousParameters = new HashSet<>()
-        Set<String> previousOptions = new HashSet<>()
+        Set<String> previousOptions    = new HashSet<>()
 
         // First, use the action from the job if it's present.
         DeclarativeJobPropertyTrackerAction previousAction = j.getAction(DeclarativeJobPropertyTrackerAction.class)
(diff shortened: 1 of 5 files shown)


scoheb:
Tested this code in my local env with jobs with triggers and the triggers were not restarted when executing builds
looks good!

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -557,110 +557,200 @@ class Utils {
...
-
-        // First, use the action from the job if it's present.
-        DeclarativeJobPropertyTrackerAction previousAction = j.getAction(DeclarativeJobPropertyTrackerAction.class)
-
-        // Fall back to previous build for compatibility reasons.
-        if (previousAction == null) {
-            WorkflowRun previousBuild = r.getPreviousCompletedBuild()
-            if (previousBuild != null) {
-                previousAction = previousBuild.getAction(DeclarativeJobPropertyTrackerAction.class)
+        synchronized (j) {
```
bitwiseman: Synchronized based on the WorkflowJob.  Okay.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -557,110 +557,200 @@ class Utils {
...
+     * Used for objects:
+     * - all fields are unknown before comparison
+     * - without overwritten methods equals() and hashCode()
+     *
+     * @param first  First object for comparing
+     * @param second Second object for comparing
+     *
+     * @return {@code true} if string representation of objects in XML format are equals; {@code false} otherwise
+     */
+    private static boolean isObjectsEqualsXStream(@Nonnull Object first, @Nonnull Object second) {
```
bitwiseman: I understand that this works even if the Trigger classes do not implement `getHashCode` or `equals`, but this seems like a really heavy way to check for equality.  Those classes should provide those implementations, right?  

This works but wouldn't you agree it would be better to fix those classes?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/TriggersTest.java
```diff
@@ -162,4 +175,102 @@ public void actualTriggerCorrectScope() throws Exception {
         expect("actualTriggerCorrectScope")
                 .go();
     }
-}
+
+    @LocalData
+    @Test
+    public void doNotRestartEqualTriggers() throws Exception {
```
bitwiseman: Same feedback as for this test in the other PR.  We need tests showing the behavior works when there are more than one trigger and shows that if the triggers are reordered they still are detected as unchanged.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
gulyaev13: > Those classes should provide those implementations, right?

Yes, they should!

> This works but wouldn't you agree it would be better to fix those classes?

Yes, they would. But those methods should be implemented in child classes of Trigger, ParameterDefinition, and JobProperty across all plugins. How can we do this?

The suggested method seems heavy at first look, but now all job properties are serialized and written to disk at each run. Approximately this solution will use the same memory resources, but will not block each thread for slow IO.

gulyaev13:
@bitwiseman , I will write more tests!
But I have no idea how to test that synchronization works properly.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
bitwiseman: @gulyaev13 
> Yes, they would. But those methods should be implemented in child classes of Trigger, ParameterDefinition, and JobProperty across all plugins. How can we do this?

It will take some time and effort, basically contributing fixes to a number of plugins. It is fine if you want to file a separate JIRA for them, but the goal should be to be to not depend on serialized XML strings for equality comparison.

How about this: use `equals()` comparison between items.  If that returns `true` accept the items as unchanged.  If they return false, fall back to checking `isObjectsEqualsXStream()`.  This should result in better performance for class that define `equals()/hashCode()`, but still working for those that don't.   

We'll need a test for each path - with and without.  But then we could also test the performance difference.

bitwiseman:
> @bitwiseman , I will write more tests!
> But I have no idea how to test that synchronization works properly.

Fair point, it isn't easy.  Ah, how about extending the test that we started with, but instead of running one job at each step, run 4 at once.  So, 4 jobs that have the same triggers, then another 4 jobs with new triggers that are the same as each other?  There should be only one update.  At least it has a chance of detecting lack of synchronization.  The lack of tests is part of why this is the way it is.

scoheb:
**retested** latest commits in my local env with jobs with triggers and the triggers were not restarted when executing builds.

bitwiseman:
Updated with an ignore for spotbugs issue with groovy.

bitwiseman:
@gulyaev13 
I also found that the trigger property doesn't support fine-grained updates, so I simplified the trigger updates.

scoheb:
@bitwiseman  When do you predict this being available in a release? Thanks alot for seeing this thru @gulyaev13 and @bitwiseman !!

bitwiseman:
@scoheb 
I predict it was available about 5 hours ago in v1.5.0.  😄

PR #366, opened 2019-11-19
Support comparator pattern for the changeset directive

v1v:
* JENKINS issue(s):
    * [JENKINS-60217](https://issues.jenkins-ci.org/browse/JENKINS-60217)
* Description:
    * Uses the `comparator` argument as already done with `tag` and `branch`.
* Documentation changes:
    * `help-*.html`.
    * https://github.com/jenkins-infra/jenkins.io/pull/2664
* Users/aliases to notify:

## Questions
- `caseSensitive` might not be relevant anymore as the `(?i)` with the `REGEXP` comparator should be enough. Do you think it might be worth to deprecate it? Or just remove it?

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeSetConditional.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeSetConditional.java
@@ -25,6 +25,7 @@
 
 package org.jenkinsci.plugins.pipeline.modeldefinition.when.impl;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.scm.ChangeLogSet;
 import hudson.util.ListBoxModel;
@@ -47,6 +48,7 @@ import java.io.IOException;
  *
  * The build must first have collected the changelog via for example <code>checkout scm</code>.
  */
+@SuppressFBWarnings(value = "SE_NO_SERIALVERSIONID")
 public class ChangeSetConditional extends DeclarativeStageConditional<ChangeSetConditional> {
 
     @Deprecated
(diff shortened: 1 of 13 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeSetConditional.java
```diff
@@ -25,48 +25,96 @@
...
+        this.pattern = pattern;
         this.caseSensitive = false;
     }
 
+    @Deprecated
     public String getGlob() {
         return glob;
     }
 
+    @Deprecated
```
bitwiseman: The problem here is that we still have `caseSensitive` in the UI and there may be code out there using it as well.  
Can you leave it in?  people probably still want the option for case insensitive globs.

v1v: Sure thing. I thought it might be easier to deprecate it and encourage people to use regexp.

v1v: See https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/366/commits/d138068ca09c45b8e235350a039bf2c913cb3b49 in conjunction with https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/366/commits/2e17299977deb9b102caa3983eec019196ed0e89 where I added the support to the case sensitive for the GLOB comparator.

v1v:
> Still needs tests showing the regex succeeding not just skipping.

Not sure if I understood, there are already a few asserts to compare whether the log contains or not contain the regex:
- https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/366/files#diff-906b17f322d66db859241e6332221c85R94
- https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/366/files#diff-906b17f322d66db859241e6332221c85R110
- https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/366/files#diff-23084e0f09c3db5f0b6a7f0bb28f36c9R170

Is that what you meant?

bitwiseman:
@v1v 
Tests: Ah, I see.  I didn't read them careful enough.

bitwiseman:
@v1v 
I'm releasing 1.5.0 today. This is good addition.  Your work is good, there's just cruft in the system.

PR #367, opened 2019-12-10
:wip: Enable changeset when pattern comparator with fullMatch

v1v:
* JENKINS issue(s):
    * Issue link or n/a
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/test/resources/when/conditions/changelog/changeset.groovy
+++ b/pipeline-model-definition/src/test/resources/when/conditions/changelog/changeset.groovy
@@ -51,5 +51,13 @@ pipeline {
                 echo "With regexp"
             }
         }
+        stage("Four") {
+            when {
+                changeset pattern: '(.*\\.js|Jenkinsfile)', comparator: 'regexp', shouldMatchAll: true
+            }
+            steps {
+                echo "With shouldMatchAll"
+            }
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


pipeline-model-definition/src/test/resources/when/conditions/changelog/changeset.groovy
```diff
@@ -51,5 +51,13 @@ pipeline {
                 echo "With regexp"
             }
         }
+        stage("Four") {
+            when {
+                changeset pattern: '(.*\\.js|Jenkinsfile)', comparator: 'regexp', shouldMatchAll: true
```
bitwiseman: If we need this, it looks good.   But could the same behavior be achieved like this?  
 
```suggestion
                changeset pattern: '^(.*\\.js|Jenkinsfile)$', comparator: 'regexp'
```

v1v: If this is related to the test example, probably it could be improved, if it's related to this PR, then I've got some doubts if the regex could help since the current pattern is searched with `anyMatch ` https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/367/files#diff-cbce163a23e232096d8a5d0ef5bddf31L113

For instance, one particular use case could be when changes are not related to the source code but some markdown files or files in the docs folder, therefore the pattern `(.*\\.md|docs/.*)`  could help, but what if the changeset contains docs and source code, then this particular feature could help to enable stages for compiling/testing the project.

Please let me know if I misunderstood your question, thanks

josh-tepper-clarifai: +1 -- I'm looking for exactly this feature.  @bitwiseman -- this is for matching all files in the changeset and _not_ about matching the entire filename

PR #368, opened 2019-12-10
WIP: enable comparator in the environment when

v1v:
* JENKINS issue(s):
    * Issue link or n/a
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/EnvironmentConditional/help-comparator.html
```diff
@@ -0,0 +1,36 @@
...
+<p>
+    What way to compare the pattern specified to the actual value.
+    <!-- TODO Update when new Comparators are added-->
+    Possible values are:
+<ul>
+    <li><strong>EQUALS</strong> - A plain string comparison</li>
+    <li><strong>GLOB</strong> - <a href="https://ant.apache.org/manual/dirtasks.html#patterns">ANT style pattern</a></li>
+    <li><strong>REGEXP</strong> - <a href="https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html">Regular Expression</a></li>
+</ul>
+If not specified <strong>GLOB</strong> will be used as default.
```
v1v:
```suggestion
If not specified <strong>EQUALS</strong> will be used as default.
```

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/EnvironmentConditional.java
```diff
@@ -25,28 +25,31 @@
...
+import java.io.IOException;
 
 /**
  * Stage condition based on environment variable equality.
  */
+@SuppressFBWarnings(value = "SE_NO_SERIALVERSIONID")
 public class EnvironmentConditional extends DeclarativeStageConditional<EnvironmentConditional> {
     private final String name;
-    private final String value;
+    private String value;
```
v1v:
```suggestion
    private final String value;
```

PR #369, opened 2019-12-11
Store full pipeline definition in ExecutionModelAction and add abstract ModelValidator class

dwnusbaum:
* JENKINS issue(s):
    * [JENKINS-60531](https://issues.jenkins-ci.org/browse/JENKINS-60531)
* Description:
    * Developer-only change. Right now, there are two issues that make it difficult to use the Declarative AST from another plugin to statically analyze a Declarative Pipeline:
      * `ExecutionModelAction` does not contain the full Pipeline, only the stages, so there is no way to analyze the top-level agent, options, environment, etc. My proposed fix for this issue is to make `ExecutionModelAction` store the full Pipeline going forward, along with implementing a bit of compatibility code for old builds where only the stages were stored.
      * I think the best way to walk/analyze the AST of a Declarative Pipeline is to use `ModelValidator` from `pipeline-model-api`, however, that interface recently got some incompatible changes (new methods were added) in relation to the matrix feature. If any other plugins were using that interface, it would have broken binary compatibility. 

        I'm not sure about the best way to prevent this kind of issue going forward, so this PR just adds a new `AbstractModelValidator` class that implements the `ModelValidator` interface and does nothing in all methods. Whenever `ModelValidator` gets incompatible changes (methods added), the idea would be to change the implementation of `AbstractModelValidator` to preserve binary compatibility. A different approach could be to add some notes to `ModelValidator` explaining that any new methods that are added need to be a `default` method to avoid issues with binary compatibility, and that methods cannot be removed/renamed.
* Documentation changes:
    * None? (I don't think anything extensive is needed for developer-only changes, but not sure what is normally done for this kind of thing).

### DIFF
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTPipelineDef.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTPipelineDef.java
@@ -118,7 +118,7 @@ public final class ModelASTPipelineDef extends ModelASTElement {
     @Override
     public void removeSourceLocation() {
         super.removeSourceLocation();
-        removeSourceLocationsFrom(stages, libraries, postBuild, environment, tools, options, parameters, triggers);
+        removeSourceLocationsFrom(agent, stages, libraries, postBuild, environment, tools, options, parameters, triggers);
     }
 
     private static String indent(int count) {
--- a/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/AbstractModelValidator.java
+++ b/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/AbstractModelValidator.java
@@ -1,0 +1,233 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2019 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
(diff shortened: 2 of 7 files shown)


pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTPipelineDef.java
```diff
@@ -118,7 +118,7 @@ public static String toIndentedGroovy(@Nonnull String orig) {
     @Override
     public void removeSourceLocation() {
         super.removeSourceLocation();
-        removeSourceLocationsFrom(stages, libraries, postBuild, environment, tools, options, parameters, triggers);
+        removeSourceLocationsFrom(agent, stages, libraries, postBuild, environment, tools, options, parameters, triggers);
```
dwnusbaum: We never removed source locations from `agent` previously. I'm not sure if it mattered, but I noticed  it because some test started failing when I started storing the whole Pipeline in `ExecutionModelAction` instead of just the stages.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -166,14 +166,44 @@ public void executionModelAction() throws Exception {
         ExecutionModelAction action = b.getAction(ExecutionModelAction.class);
         assertNotNull(action);
         ModelASTStages stages = action.getStages();
-        assertNull(stages.getSourceLocation());
+        assertExecutionModelActionStageContents(b, stages);
+    }
+
+    @Test
+    public void executionModelActionFullPipeline() throws Exception {
```
dwnusbaum: Essentially the same as the `executionModelAction` test, but uses `ExecutionModelAction.getPipelineDef` and checks for some info that isn't available if you use `ExecutionModelAction.getStages`. I thought it was useful to still keep the old test around to help maintain compatibility for any existing callers.

dwnusbaum:
Ok, I filed [JENKINS-60531](https://issues.jenkins-ci.org/browse/JENKINS-60531) to track this, so I think it's ready for review.

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/validator/AbstractModelValidator.java
```diff
@@ -0,0 +1,233 @@
...
+import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTTrigger;
+import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTTriggers;
+import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTValue;
+import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTWhen;
+import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTWhenCondition;
+
+/**
+ * Abstract implementation of {@link ModelValidator}.
+ *
+ * Use this class as a generic AST visitor instead of {@link ModelValidator} to prevent binary compatibility issues
```
dwnusbaum: Maybe also good to update {{ModelValidator}}'s Javadoc to explicitly note that it is not a stable API and should not be used directly by plugins other than Declarative itself?

pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTPipelineDef.java
bitwiseman: Oops, yes, it does matter.  This is another reason to do this.  Cool.

bitwiseman:
Aside from the random windows failure. I'm good with this.

dwnusbaum:
@bitwiseman Ok, do you want me to update the Javadoc for ModelValidator, or do you want to just leave it for now?

bitwiseman:
@dwnusbaum 
It is up to you, but it looks like you've already written the text. 😄 It is worth preserving your observations in code rather than just a PR.

dwnusbaum:
Ok, I just added Javadoc for ModelValidator, hopefully the random Windows failures go away.

dwnusbaum:
`MatrixTest` timed out. Not sure if it's flaky or there were infra issues. Trying a rebuild to see what happens.

PR #370, opened 2019-12-16
Remove dependency on obsolete tests jar for workflow-scm-step

dwnusbaum:
* JENKINS issue(s):
    * N/A
* Description:
    * Previously, workflow-scm-step releases included a `-tests` jar with some utilities. Long ago, in https://github.com/jenkinsci/workflow-scm-step-plugin/pull/15 those utilities were moved to other plugins, and in https://github.com/jenkinsci/workflow-scm-step-plugin/pull/17, workflow-scm-step was changed to no longer publish a tests jar. pipeline-model-definition still depends on an older version of workflow-scm-step where the `-tests` jar was present, which causes a compilation failure in the PCT when run against newer versions of workflow-scm-step. Since the utilities were moved to other plugins that are already in use here, I think it should be safe to just remove the dependency, but if the CI build fails I might need to tweak some test imports.
      
* Documentation changes:
    * Only affects testing using the PCT, no doc changes necessary.
* Users/aliases to notify:
    * N/A

### DIFF
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -271,12 +271,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-scm-step</artifactId>
-      <classifier>tests</classifier>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-support</artifactId>
         <classifier>tests</classifier>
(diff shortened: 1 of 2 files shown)


bitwiseman:
Unclear why this only failed on Windows 8.

dwnusbaum:
> Unclear why this only failed on Windows 8.

Yeah, IDK. Looking at the test failures, it seems like `Jenkins.instance` was null when it shouldn't have been, and then that lead to a bunch of errors. I'm not sure what the catalyst was, perhaps we have a buggy test somewhere that isn't cleaning up after itself. I think updating the parent pom to 3.54 (or directly updating to jenkins-test-harness 2.56) to pull in https://github.com/jenkinsci/jenkins-test-harness/pull/166 would make tests fail if there are errors cleaning up the `JenkinsRule` at the end of the test where previously those errors would have gone unnoticed.

dwnusbaum:
ci.jenkins.io never triggered a second build, trying again.

dwnusbaum:
Ok, looks like all of the checks passed that time.

PR #372, opened 2020-01-28
[JENKINS-58732] Preparatory refactoring

jglick:
Subsumes #374. Preparation for [JENKINS-58732](https://issues.jenkins-ci.org/browse/JENKINS-58732); will need to be released as 1.5.1 before I can continue, since there were failures of abstraction here.

### DIFF
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
@@ -64,7 +64,7 @@ public abstract class DeclarativeAgent<A extends DeclarativeAgent<A>> extends Wi
 
     public void initialize(Map<String, DeclarativeOption> options, boolean explicitAgentInStage) {}
 
-    public boolean useRootAgent(Map<String, DeclarativeOption> options) {
+    public boolean reuseRootAgent(Map<String, DeclarativeOption> options) {
         return false;
     }
 
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
@@ -25,12 +25,13 @@
 package org.jenkinsci.plugins.pipeline.modeldefinition.agent;
 
 import hudson.ExtensionPoint;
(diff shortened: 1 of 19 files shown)


pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
```diff
@@ -61,6 +62,12 @@ public boolean isInStage() {
         return inStage;
     }
 
+    public void initialize(Map<String, DeclarativeOption> options, boolean explicitAgentInStage) {}
+
+    public boolean useRootAgent(Map<String, DeclarativeOption> options) {
```
jglick: Not the prettiest signatures; just trying to extract what seems to have been rather ad-hoc existing logic. Suggestions for clearer names and/or parameters welcome, especially something that seems likely to be more generally applicable.

dwnusbaum: The signature of `initialize` seems fine to me. The only thing that comes to mind would be to have `Object context` as a parameter instead of computing `explicitAgentInStage` in the caller in case other implementations want to check some other condition by casting to `Root` or `Stage`, but that's pretty awkward to work with. Besides, given that no other implementations of `DeclarativeAgent` have needed a method like `initialize` yet, it's probably better to keep things simple.

Maybe rename `useRootAgent` to `reuseRootAgent`? I am not very familiar with how everything hooks together to make `newContainerPerStage` work and what exactly the effect is, or if it could make sense for another implementation of `DeclarativeAgent` to ever return true from this method, so I'm not sure about what naming makes the most sense.

jglick: > casting to `Root` or `Stage`, but that's pretty awkward

Quite awkward, especially as these types are defined in Groovy (for no good reason that I can see).

> I'm not sure about what naming makes the most sense

I really do not know.

jglick:
FTR the complicated stuff comes from #246. Read @abayer starting from the word _magically_. :-)

As far as the `Options`, JIRA says

> I went with an `option` instead of a flag on `docker` and `dockerfile` - the option is `containerPerStage()`. I feel like this may be relevant in other agent types down the road, so wanted to decouple the syntax from the Docker/Dockerfile implementations.

Obviously without an API, this could not have been used in other implementations, but for what it is worth I cannot recall any similar need ever coming up with `kubernetes`. Then again, I am usually seeing Scripted syntax.

PR #336, opened 2019-06-11
Simplifying ExpectationsBuilder.go

jglick:
As discussed in https://github.com/jenkinsci/workflow-job-plugin/pull/131#discussion_r292131627, of which this is downstream.

PR #374, opened 2020-01-28
<tag>${scmTag}</tag>

jglick:
I think #364 broke the incrementalification I did in #336.


jglick:
Some test flakes here which I believe are traceable to https://github.com/jenkinsci/script-security-plugin/pull/279 released in 1.69. Not going to do such an update here though.

jglick:
Waiting for a stable build to verify that incremental deployment actually happened.

PR #376, opened 2020-02-04
adding timeout before node with top level agents

joseblas:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-60994
* Description:
    * Describe why timeout are different between top level and stage level
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/2853


pipeline-model-definition/src/test/resources/options/provisioningTimeout.groovy
```diff
@@ -0,0 +1,46 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+pipeline {
+    agent any
+    options {
+        provisioningTimeout()
```
bitwiseman: I'm not liking adding a new option for this, but maybe my dislike is due to lack of understanding.  Could you comment a bit more on what you're trying to achieve?  Why not have the top level timeout include provisioning by default?

joseblas: The main idea is not to change the current behaviour, just to create a new option for users.

kshultzCB: Could an option be added to the existing `timeout` option? So you might have something like this:

```groovy
    options {
        timeout(time:10, unit: "MINUTES", includeProvisioning: true
    }
```

One thing that @joseblas 's proposal would provide is more granularity. If I've understood it correctly, there would two distinct timeouts - one for spinning up the agent, the other for the actual build work.

joseblas: Indeed, the idea is for applying the timeout to the whole process: agent provisioning and the actual work. @kshultzCB, adding that option to timeout was also on my mind, could be an option.

bitwiseman: The point of timeouts is to limit run duration and having the default not include checkout or agent provisioning is unexpected.  

My reflex is to make change the default behavior to include checkout and agent provisioning.  Especially at top level. If this breaks people who had their timeout set to values that didn't include checkout and provisioning, I'm okay with that.  

We could provide a way for checkout or provisioning to have an additional timeout within the larger timeout, but do we really need to?  

Giving people more options isn't always better.

joseblas: What if you don't care about provisioning? You just care about timeout in your tests?
Indeed, @kshultzCB idea about specific timeout for provisioning sounds good to me.

bitwiseman: But that isn't what this change is about. 
If someone wanted to do a timeout one part of a stage they can already do that using the `timeout` wrapper inside of `steps`.  If they want a timeout on a stage as a whole (including provisioning and checkout), they use the stage's `option`.  

Right now, declarative users cannot get a timeout on their whole pipeline including the provisioning of their top-level agent.  That is what we need done.

bitwiseman: The existing behavior is a bug not a feature. 😄

car-roll: I do seem benefit in the extra granularity in that it might be able to speed up when something is going wrong with the pipeline.  Many times people will put in a large timeout due to a heavy test, but often times there can be an issue with provisioning hanging relatively quickly so that the tests never even get executed. Unless you're babysitting the pipeline you've pretty much wasted a good hour or so.  In that kind of scenario, a way to only consider provisioning timeout is usefule.

dwnusbaum: I don't see the issue here as being specific to `timeout` or provisioning time, and I don't think the fix should be specific to `timeout`. My interpretation of the core issue is that at the top level, `agent` is evaluated before `options`, but at the stage level, `agent` is evaluated after `options`, which is inconsistent, and users have no way to control the ordering.

The fix that makes the most sense to me would be to add a `beforeAgent` option to `options`, which would default to `false` at the top level and `true` at the stage level. Users who want `timeout` to include agent provisioning time at the top level could use `beforeAgent(true)` in their top-level `options`. This could also be used for any other block-scope step that a user wants to execute at the top level before the agent (or vice versa for stage-level). My main concern with this approach is that maybe there are cases where someone wants `timeout` to execute before `agent` but also to have a block-scope step at the top level that requires an `agent` context, which wouldn't be possible with this fix.

Special-casing `timeout` in top-level options seems weird to me. Changing top-level `options` to execute before top-level `agent` by default is probably off the table because it would break too many Pipelines. Maybe another fix could be a new `beforeAgentOptions` `options` block at the top level, but that seems awkward too.

bitwiseman: > I don't see the issue here as being specific to `timeout` or provisioning time, and I don't think the fix should be specific to `timeout`.
...
> Special-casing timeout in top-level options seems weird to me. 

I agree with this.  

> The fix that makes the most sense to me would be to add a `beforeAgent` option to...

Why would we want `options` processing to default to a different behavior at the top-level ? 

> My main concern with this approach is that maybe there are cases where someone wants timeout to execute before agent but also to have a block-scope step at the top level that requires an agent context, which wouldn't be possible with this fix.

What would that be currently?  If we want to add functionality in the future, that is a separate change. Unless we have a strong, concrete example of where this breaks an existing option we should change this to be consistent.  

Let me run some tests.

bitwiseman:
I created #377 to test the results of the most direct change. 

Testing locally I saw only a few errors.  The most obvious one was: 

```
[ERROR] withCredentialsWrapper(org.jenkinsci.plugins.pipeline.modeldefinition.OptionsTest)  Time elapsed: 0.561 s  <<< FAILURE!
[ ... ]
org.jenkinsci.plugins.workflow.steps.MissingContextVariableException: Required context class hudson.FilePath is missing
Perhaps you forgot to surround the code with a step that provides this, such as: node
```

There's the wrapper requiring context that @dwnusbaum  referred to. It looks like we don't have a test for the `withCredentials` option at stage level that also has an `agent` set for that stage.  If we did, it would fail this same way.  

https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/408ec8ff4c0fcfcf58040e60e9b074b5e4201e05/pipeline-model-definition/src/test/resources/options/withCredentialsWrapper.groovy#L25-L31


I understand the adding reordering to sounds great but each time we add another option for order we expand our test matrix and make it harder for users to understand what is going on. The general solution that declarative has used for cases where ordering is needed is adding a child stage (similar to the above test).  If something can be achieved by adding a child stage, I would push for it not being added as an option. 

At this time we cannot do a true timeout for the whole pipeline. If we permanently switched the order as in #377, that would be fixed.  Further if someone wanted the current behavior back the could do that - have your top level `agent any`, add a child stage with the `timeout()` and you are back to current behavior.  it's not pretty but it is stable and consistent. 


@dwnusbaum @joseblas  
If the team overall decided  we must add an ordering control, I'd rather add a `beforeOptions` option to `agent` and have it default to `false` (top-level and stage).  This could cause breakage for some pipelines but adding `beforeOptions(true)` would work around it. 

So, we have at least three ideas of what to do here.   🎉

joseblas:
This issue was created because there is a problem with timeout only. Then, has been noticed that options are applied inconsistently, before agent at stage level and the other way around in top level. If the stage level version is applied to top level, there are some errors due to some wrapper.

 My ideal solution would be one providing flexibility to users in this case I think a `beforeAgentOptions` `block` for top level only. Which options will be available for that? If is timeout only, when special-casing timeout is not that weird.

PR #377, opened 2020-02-06
Move Options before agent in top-level

bitwiseman:
Trying this to get concrete data we can all look at.  

This is **not** a replacement for #376.


bitwiseman:
Running this locally has gotten me a very small number of errors.  
I want to see if CI agrees.

PR #379, opened 2020-02-19
JENKINS-60994 Fixing timeout tests

joseblas:
This PR is for amending [#376](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/376)
 regarding https://issues.jenkins-ci.org/browse/JENKINS-60994

Documentation changes:
jenkins-infra/jenkins.io#2853


jglick:
IIUC this is just amending tests from #376 so it would be better to clarify that in PR title & description.

PR #380, opened 2020-03-02
adding directive generator for matrix

joseblas:
updating matrix deritective generator only. No real changes in *working* code


bitwiseman:
Looking good.  I ran this up manually and notice the following: 

I agree what you said in our other conversation - `axes`, `axis`, `excludes`, and `exclude` should not be in the top level drop down if possible:

Screen Shot 2020-03-04 at 12 05 45 AM

The "Name"  field for `axis` and `exclude` should automatically wrap in quotes (that is what stage "Name" does): 

Screen Shot 2020-03-04 at 12 04 17 AM

Screen Shot 2020-03-04 at 12 04 04 AM

I think the items in "Values" should also automatically wrap in quotes - there is no valid case where they are not single-quote string literals. 

Finally, `exclude { axis { ... } }` should allow a `notValues` probably via a "Not" checkbox. 

I know you still have this marked as "WIP" but I wanted to get you a review near the start of your day.  

This is awesome! Thanks!

joseblas:
btw, why the stage generator is not like the others?

bitwiseman:
> btw, why the stage generator is not like the others?

Not the same in what way?

joseblas:
I guess should be added like agent or input.

bitwiseman:
> I guess should be added like agent or input.

Could you say a bit more?  I still don't know what you're referring to.

joseblas:
An Stage can have:
   1. One of Steps, parallel or sequential stages.
   2. Any of these , but only once: Agents, Inputs... 

 When adding the latter, is done using directive generators; but when using steps, parallel or sequential stages is done *manually*, not using a directive generator. Could be added another hetero-list for that.
 For Steps will be like it is currently, but for the rest a stage directive could be used. 

Also, that is a different conversation than this change

bitwiseman:
> An Stage can have:
> 
> 1. One of Steps, parallel or sequential stages.
> 2. Any of these , but only once: Agents, Inputs...
> 
> When adding the latter, is done using directive generators; but when using steps, parallel or sequential stages is done _manually_, not using a directive generator. Could be added another hetero-list for that.
> For Steps will be like it is currently, but for the rest a stage directive could be used.
> 
> Also, that is a different conversation than this change

I think if you changed the code to allow you to generate a `stage -> stages -> stage`, the second `stage` would have browser element id collision or stapler collisions with the first `stage`.  Andrew chose to limit the depth rather than try to fight with getting the data binding to not collide. 

But yes, that is a separate issue.  I'll load up the new version of this and test it today.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/AxesDirective/help-axis.html
```diff
@@ -0,0 +1,26 @@
...
+  ~ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  ~ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  ~ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+<div>
+    Multiple options can be specified.
+</div>
```
kshultzCB: I don't know if there is a fix for this, because of the way the inline help is handled in the Jenkins UI. This comment is probably way beyond the scope of this PR. But I noticed it, so I thought I'd at least bring it up for discussion.

Here's a screen shot of me using `axes` --> `axis` in the generator.

image

I bring this up because it's impossible to tell which (?) icon lines up with which field in the page. For example, which of the three (?) icons shown below do I use to surface this help text?

image

Things get more cluttered when I start opening more than one help at the same time:

image

What would be really cool is if we could draw dotted lines around stuff according to what (?) button goes with a particular field. The Delete buttons already do this - if you hover over them, they draw a box around what I'm about to delete:

image

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/AxesDirective.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+
+        @Override
+        @Nonnull
+        public String toGroovy(@Nonnull AxesDirective axes) {
+            StringBuffer result = new StringBuffer("axes {\n");
+            if(axes.axis != null){
+                axes.axis.stream().forEach( axis -> {
+//                    System.out.println("# " + axis.toGroovy(true));
```
kshultzCB: This could be removed.

PR #381, opened 2020-03-03
Removing Provisioning test as they are flakey

joseblas:
Removing tests added in [here](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/376) because they are flakey


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ProvisioningTimeoutModelDefTest.java
```diff
@@ -1,69 +0,0 @@
...
-
-    @Parameterized.Parameters
-    public static Object[][] data() {
-        return new Object[10][0];
-    }
-
-    @Test
-    @Issue("JENKINS-60994")
-    public void topLevelAgent() throws Exception {
-        expect("options/topLevelAgentSuccessWithTimeout")
```
dwnusbaum: Should the script files be deleted as well, or are they still used by other tests?

jglick: This is what irritates me about the idiom of keeping test pipelines as separate files: syntax highlighting etc. are certainly nicer than when inline as Java strings, but it is harder to review diffs in some cases and you can wind up with stale files.

joseblas:
I do agree Jesse, but also irritates me that those tests are not stable. It should be possible to have a test with predictable timeout. Even within a provisioned node.

PR #356, opened 2019-10-15
[JENKINS-51865] Introduce beforeOptions in 'when'

famod:
* JENKINS issue(s):
    * https://issues.jenkins-ci.org/browse/JENKINS-51865
* Description:
    * Introduces `beforeOptions` in `when` to be able to skip any (costly and pointless) `options` like `lock` (https://github.com/jenkinsci/lockable-resources-plugin).
* Documentation changes:
    * https://github.com/jenkins-infra/jenkins.io/pull/2569
* Users/aliases to notify:
    * @abayer
    * @bitwiseman

Some remarks:
- A single `before` attribute with possible values `'options'` XOR `'input'` XOR `'agent'` would be more concise but `beforeAgent` and `beforeInput` are already out there "in the wild".
- Strangely, `beforeInput` was missing from the generator (I fixed that).
- Also strangely, `beforeInput` and `beforeAgent` are _not_ evaluated when using `parallel` (but the new `beforeOptions` _is_ evaluated because I did not see why it shouldn't).
- I tried to give all user facing aspects a logical order (options, input, agent) but for the rest I oriented myself to `beforeInput` which was introduced after `beforeAgent`.
- I did not touch the matrix based tests because I wouldn't really know what I was doing...

PR #382, opened 2020-03-09
[JENKINS-61007] Do not use @SuppressFBWarnings in src/main/resources/**/*.groovy

jglick:
[JENKINS-61007](https://issues.jenkins-ci.org/browse/JENKINS-61007)

Apparently the annotation introduced in #356 cannot be resolved in this context. Groovy may have different rules for lax resolution of annotations than Java. I do not know how to reproduce inside a functional test due to the different class loading environment. Anyway, I can confirm that this patch fixes the problem for me locally.


jglick:
CI crashed. [INFRA-2075](https://issues.jenkins-ci.org/browse/INFRA-2075), perhaps.

dwnusbaum:
It looks like the Windows branch has been having a variety of potentially infra-related issues (something to do with `ToolInstallations`) on the master branch as well. The fact that the Linux branches are passing with this change seems good enough to me, so I will go ahead and merge.

PR #373, opened 2020-01-28
[JENKINS-58732] Removing Docker agent functionality

jglick:
[JENKINS-58732](https://issues.jenkins-ci.org/browse/JENKINS-58732)

Downstream of https://github.com/jenkinsci/docker-workflow-plugin/pull/202 for logistical reasons; see that PR’s description for a release guide.

PR #383, opened 2020-03-10
[JENKINS-58732] Make docker-workflow dep optional

jglick:
As mentioned in https://github.com/jenkinsci/kubernetes-plugin/pull/735, follows up #373 to make this dependency optional so `docker-workflow` could actually be disabled. (Tested that scenario via `hpi:run`.)


jglick:
Any plans to release this?

PR #384, opened 2020-03-28
fixed behaviour of cli declarative validator

danieltaub96:
* JENKINS issue(s):
    * [JENKINS-59378](https://issues.jenkins-ci.org/browse/JENKINS-59378?page=com.atlassian.streams.streams-jira-plugin%3Aactivity-stream-issue-tab)
* Description:
    * As described in the bug, when using declarative-linter from jenkins-cli it's not perform the same validation as in the ModelConverterAction.java file that handles the api for validate Jenkinsfile,

I've to mention that it would be much nicer if the validator would work also for scripted pipelines and not just declarative pipelines.


bitwiseman:
@danieltaub96 


For scripted pipelines, the validation would either be extremely limited or extremely hard.  But please make sure there is an issue open for that and describe where you'd like it to start.  

Also, I invite you to join the [Pipeline Authoring SIG](https://www.jenkins.io/sigs/pipeline-authoring/).  At minimum, you'll find others interested in this area.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/cli/DeclarativeLinterCommand.java
```diff
@@ -53,8 +54,15 @@ protected int run() throws Exception {
         if (script != null) {
             try {
                 Converter.scriptToPipelineDef(script);
-                output.add("Jenkinsfile successfully validated.");
-                retVal = 0;
+
+                ModelASTPipelineDef pipelineDef = Converter.scriptToPipelineDef(script);
```
bitwiseman, line 58: Do you want to remove line 56?

README.md
```diff
@@ -18,7 +18,7 @@ The Pipeline Model Definition Plugin provides a config-like syntax for defining
 - To start developing plugin: 
 - Clone the plugin repository
 
-`cd pipeline-model-definition && mvn install && mvn hpi:run`
+`cd pipeline-model-definition-plugin && mvn install && mvn hpi:run`
```
dwnusbaum, line 21: This doesn't work as-is because you can't run `hpi:run` from the root of a multi-module repo:
```suggestion
`cd pipeline-model-definition-plugin && mvn install && mvn -f pipeline-model-definition/pom.xml hpi:run`
```

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/cli/DeclarativeLinterCommand.java
```diff
@@ -53,8 +54,15 @@ protected int run() throws Exception {
         if (script != null) {
             try {
                 Converter.scriptToPipelineDef(script);
```
dwnusbaum, line 56: Right now the Pipeline gets converted twice.

```suggestion
```

README.md
kshultzCB, line 21: @dwnusbaum already spotted this, but running this way won't provide Declarative's functionality. You'll just end up with a pretty vanilla Jenkins that can only create a Freestyle job:

image

`mvn -f pipeline-model-definition/pom.xml hpi:run` can work, or you can `cd pipeline-model-definition` then `mvn hpi:run` that way.

My suggestion: do away with the change to the README. I think this could be made more clear, but maybe it's for a different PR.

bitwiseman:
```suggestion
`cd pipeline-model-definition && mvn install && mvn hpi:run`
```

bitwiseman:
@dwnusbaum @kshultzCB 
I've committed the changes you suggested.  Merging.

kshultzCB:
Here's the 👍 from me even though I'm late to provide it. Was distracted.

korablin:
@danieltaub96 have you opened an issue to validate scripted pipelines as well?

PR #386, opened 2020-04-12
Remove dependency on pipeline-model-declarative-agent

zbynek:
... and provide simple readme that can be displayed on plugins.jenkins.io for the deprecated plugin (AFAIK it doesn't have its own repo).

Note that <0.5% of declarative-agent users are not using the tombstone release and out of those  most are on old cores: https://stats.jenkins.io/pluginversions/pipeline-model-declarative-agent.html

### DIFF
--- a/pipeline-model-declarative-agent/README.md
+++ b/pipeline-model-declarative-agent/README.md
@@ -1,0 +1,1 @@
+This plugin is deprecated, please use [Pipeline: Declarative](https://plugins.jenkins.io/pipeline-model-definition/) plugin instead.
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -180,13 +180,6 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>pipeline-input-step</artifactId>
     </dependency>
-    <!-- Dependency on tombstoned declarative agent plugin to make sure we don't end up with a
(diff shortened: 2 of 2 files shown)


pipeline-model-declarative-agent/README.md
```diff
@@ -0,0 +1 @@
+This plugin is deprecated, please use [Pipeline: Declarative](https://plugins.jenkins.io/pipeline-model-definition/) plugin instead.
```
dwnusbaum, line 1: I was confused about how this would show up on plugins.jenkins.io, but it looks like there is a way to override the URL used per-plugin that would be used to point to this file, see https://github.com/jenkins-infra/update-center2/pull/367.

timja:
Would be good to get a release of this, we can't really ship https://github.com/jenkinsci/jenkins/pull/4073 without it, as the default recommended plugins ship with a deprecated dependency and no way to uninstall it.

jetersen:
Would be great with that release 👍

timja:
Ping @bitwiseman @dwnusbaum could we get a release please

jetersen:
Perhaps better to ping for release permissions? 😅

dwnusbaum:
I can try to release in the next few days or so.

jetersen:
@dwnusbaum much appreciated!

bitwiseman:
@jetersen v1.7.0 released.

PR #387, opened 2020-04-19
test case for user AST with declarative pipeline

tavin:
* JENKINS issue(s):
    * [JENKINS-49135](https://issues.jenkins-ci.org/browse/JENKINS-49135)
* Description:
    * This is currently working functionality which I'd like to use as a foundation for including reusable segments in declarative pipelines. Therefore the test case is just to make sure it will keep working.
* Documentation changes:
    * None needed as far as I know.
* Users/aliases to notify:
    * @abayer

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
@@ -377,6 +377,61 @@ public abstract class AbstractModelDefTest extends AbstractDeclarativeTest {
                 "}")
                 , "\n"));
 
+        File mypkg = new File(globalLibRepo.workspace, "src/mypkg");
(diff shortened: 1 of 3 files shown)


bitwiseman:
Really interesting use of annotations. 

This is fine for groovy but what about YAML? [That project](https://plugins.jenkins.io/pipeline-as-yaml/) is in incubation, but this highlights the risks of using Java/Groovy-specific syntax features. 

This is obviously something you care about, perhaps you'd be willing to join the [Pipeline Authoring SIG](https://www.jenkins.io/sigs/pipeline-authoring/) and work with others in the community on the project?

tavin:
I took this a bit further in my own project with an `@Include` annotation. It worked for trivial pipelines, but later I ran into issues with ModelParser trying to extract (nonexistent) code snippets from the original source file.

If that could be solved, is there any conflict with YAML pipelines? Clearly a YAML include mechanism would be different.

Thanks, I'm looking into the SIG.

tavin:
I guess there is no interest in this -- closing.

PR #394, opened 2020-06-12
[JENKINS-26097] Add auto-completion and validation of label expressions.

Zastai:
* JENKINS issue(s):
    * [JENKINS-26097](https://issues.jenkins-ci.org/browse/JENKINS-26097)
* Description:
    * Add auto-completion and validation of label expressions.
* Documentation changes:
    * Only affects declarative generator; probably no doc needed


Zastai:
Given I only touched UX parts (validation and auto-completion), I don't see how they trigger these test failures; I guess they must be compabilitiy issues with the more recent Jenkins main.

timja:
Probably this:

```
  20.515 [id=7429]	WARNING	h.i.i.InstallUncaughtExceptionHandler#handleException: Caught unhandled exception with ID 242d9287-d857-4257-ab7b-66246f9b0344
org.apache.commons.jelly.JellyTagException: jar:file:/home/jenkins/workspace/e-model-definition-plugin_PR-394%40tmp/m2repo/org/jenkins-ci/main/jenkins-core/2.241-rc30038.0428cd872667/jenkins-core-2.241-rc30038.0428cd872667.jar!/hudson/model/Run/console.jelly:66:27: <j:whitespace> com/trilead/ssh2/crypto/Base64
```

Likely a plugin needs to be updated to one that has a direct dependency on trilead-api

Zastai:
I’ll try to dig into it tomorrow.

Zastai:
Interestingly, but annoyingly, I can't reproduce these myself when running the tests in an IDE.
But I'll push a commit that adds a test dependency on `trilead-ssh2` in the hopes that it resolves the test errors during the CI build.

timja:
It’s likely the ssh-credentials version that needs bumping.

But if you introduce the plugin bom https://github.com/jenkinsci/bom it should fix these issues

Zastai:
Ah I assumed the BOM would be pulled in by the plugin parent.
Will look into it.

pom.xml
```diff
@@ -337,12 +231,12 @@
   <properties>
     <revision>1.7.1</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.176.4</jenkins.version>
+    <jenkins.version>2.243</jenkins.version>
```
bitwiseman, line 234: We definitely can't take this plugin to this as the minimum version.

Zastai: That’s fair enough - I fully expect this PR to stay open until there are other reasons to go to a version after 2.243.

Zastai:
Given that the plugin is now on a much higher Jenkins version, this should hopefully work fine now.

joseblas (approved):

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/Label.java
```diff
@@ -71,12 +73,15 @@ public String getDisplayName() {
...
         }
 
+        public AutoCompletionCandidates doAutoCompleteLabel(@QueryParameter String value) {
+            return LabelExpression.autoComplete(value);
+        }
+
         public FormValidation doCheckLabel(@QueryParameter String label) {
-            if (StringUtils.isEmpty(Util.fixEmptyAndTrim(label))) {
+            if (Util.fixEmptyAndTrim(label) == null) {
                 return FormValidation.error("Label is required.");
```
jglick, line 82: Seems to contradict https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/9f9982419a20aa302e60d658908e0283632b1500/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/Label.java#L50. There does seem to be a potentially valid reason to use a null label instead of just `Any`: specifying `retries`. (But specifying neither seems wrong.)

PR #395, opened 2020-06-16
Use Jenkins BOM to manage dependencies

Vlatombe:
Introduces Jenkins BOM in order to simplify management of dependencies.

Also
* Replaced multiple occurrences of version by maven properties
* Cleaned up Hamcrest references (only `org.hamcrest:hamcrest` should be referenced directly, others are deprecated)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,61 +72,22 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <!-- Require upper bound dependencies -->
-        <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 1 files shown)


MarkEWaite:
@Vlatombe if you're interested in extending this pull request with the potential to also fix a [recently reported issue](https://issues.jenkins-ci.org/browse/JENKINS-62988), I pushed spotbugs annotation changes to my [use-bom branch](https://github.com/MarkEWaite/pipeline-model-definition-plugin/tree/use-bom).

PR #397, opened 2020-07-08
[JENKINS-62988] Do not use 3rd-party dependencies in ModelInterpreter.groovy

dwnusbaum:
* JENKINS issue(s):
    * [JENKINS-62988](https://issues.jenkins-ci.org/browse/JENKINS-62988)
* Description:
    * The classes `ModelInterpreter.groovy` has available to it at runtime depend on the versions of Jenkins and `workflow-cps` (among other plugins) being used, so we should minimize usage of any transitive dependencies that may be removed or modified, particularly annotations.
    * TODO:
        * [x] Check that tests pass
        * [ ] Check the behavior of this PR in a real Jenkins instance (since tests have a different class loading setup)
        * [ ] Verify that switching from `ExceptionUtils.getFullStackTrace` to `Functions.printThrowable` doesn't break things
* Documentation changes:
    * N/A
* Users/aliases to notify:
    * N/A

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.pipeline.modeldefinition
 import com.cloudbees.groovy.cps.NonCPS
 import com.cloudbees.groovy.cps.impl.CpsClosure
 import hudson.FilePath
+import hudson.Functions
 import hudson.Launcher
 import hudson.model.Result
 import org.jenkinsci.plugins.pipeline.StageStatus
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -39,14 +40,13 @@ import org.jenkinsci.plugins.workflow.job.WorkflowRun
...
-import javax.annotation.CheckForNull
-import javax.annotation.Nonnull
-
-import static org.apache.commons.lang.exception.ExceptionUtils.getFullStackTrace
-
 /**
  * CPS-transformed code for actually performing the build.
  *
+ * WARNING: Avoid using 3rd-party non-Jenkins dependencies (including annotations)
+ * in this code, because even they work in tests, they may not be available at
```
dwnusbaum, line 47:
```suggestion
 * in this code, because even if they work in tests, they may not be available at
```

bitwiseman:
I can run this up in a local test.

bitwiseman:
@dwnusbaum  

I'm not able to reproduce the original problem on 2.244 with a simple declarative pipeline.  Still, I have no problem with accepting this change.

dwnusbaum:
@bitwiseman Bizarre, that is when running Jenkins in a "realistic" way, not `mvn hpi:run` or in a test and with `workflow-cps` 2.81 installed? #382 fixed a similar issue with annotations in `ModelInterpreter.groovy` that completely broke Declarative whenever it manifested, but the Declarative code had used that annotation for months and [JENKINS-61007](https://issues.jenkins-ci.org/browse/JENKINS-61007) only had 7 watchers, so it seems like the issue might only occur under specific circumstances or something...

PR #399, opened 2020-07-09
Use bom for more dependencies including Jackson

chriskilding:
Increment the Jackson version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -227,7 +227,7 @@
     <java.level>8</java.level>
     <groovy.version>2.4.12</groovy.version>
     <argLine>-Djava.awt.headless=true -Xmx4g -Xms512m -XX:MaxPermSize=512m</argLine>
-    <jackson.version>2.9.8</jackson.version>
+    <jackson.version>2.11.0</jackson.version>
     <workflow-multibranch.version>2.16</workflow-multibranch.version>
     <workflow-cps.version>2.81</workflow-cps.version>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -80,17 +80,6 @@
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
-        <artifactId>workflow-cps</artifactId>
-        <version>${workflow-cps.version}</version>
-      </dependency>
-      <dependency>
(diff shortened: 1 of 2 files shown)


bitwiseman:
@chriskilding  We don't require 2.11 to work.  Users may have an older version without ill effect.

bitwiseman:
@chriskilding  Is this change needed for the linked PR?

chriskilding:
Yep it's for the linked PR. Not needed outside of that. It's a test to see how the plugin would handle being fed a newer Jackson version from the BOM; I assume that for conformity you'd want to inherit the Jackson version from the BOM once it's added, rather than override it to 2.9.8.

pom.xml
```diff
@@ -227,7 +227,7 @@
     <java.level>8</java.level>
     <groovy.version>2.4.12</groovy.version>
     <argLine>-Djava.awt.headless=true -Xmx4g -Xms512m -XX:MaxPermSize=512m</argLine>
-    <jackson.version>2.9.8</jackson.version>
+    <jackson.version>2.11.0</jackson.version>
```
bitwiseman, line 230:
```suggestion
    <jackson.version>2.11.2</jackson.version>
```

bitwiseman:
@chriskilding 
Yes, this project would certainly inherit the Jackson version from the bom. 
I'm getting a very strange failure on Windows with this PR. Not sure why yet.  Trying a newer version.  Very likely just brittle tests.

chriskilding:
I've tried upgrading it to BOM v13 and removing the Jackson2-api artifact version locally, which should allow it to be inherited from the BOM. However when I ran `mvn validate` I got disagreements about what it should be from the various Jackson modules that are in play.

There is some discussion about adding more of the Jackson modules to either the BOM or the jackson2-api artifact in https://github.com/jenkinsci/bom/pull/294. This will make our bundled Jackson offering more 'batteries included' or plug and play for downstream authors. I suspect we'd need something like that here to make pipeline-model-definition happy.

bitwiseman:
https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fpipeline-model-definition-plugin/detail/PR-399/22/pipeline this passed.

PR #404, opened 2020-08-27
[JENKINS-63499] Make password parameters work with the parameters directive in Jenkins 2.236+

dwnusbaum:
### JENKINS issue(s):

See [JENKINS-63499](https://issues.jenkins-ci.org/browse/JENKINS-63499).

### Description:

https://github.com/jenkinsci/jenkins/pull/4630 modified the `@DataBoundConstructor` for `PasswordParameterDefinition` in a way that broke backwards compatibility for structs-based data binding. `PasswordParameterDefinition` is used by Declarative for the parameters directive, so this PR adds a special case so that old Pipelines that configure `password` parameters as described in [the documentation](https://www.jenkins.io/doc/book/pipeline/syntax/#parameters) continue to work.

The directive generator is also affected by the original core change, here is the output when you try to generate a `parameters` directive with a password parameter:

```
parameters {
  password defaultValueAsSecret: <object of type hudson.util.Secret>, description: 'test', name: 'test'
}
```

TODO: Decide if the directive generator should also be fixed as part of this PR.

### Documentation changes:

Usage is the same as before.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,5 @@
-buildPlugin(timeout: 120)
+buildPlugin(timeout: 120, useAci: true, platforms: [
+  [ platform: 'linux', jdk: '8'],
+  [ platform: 'windows', jdk: '8'],
+  [ platform: 'linux', jdk: '11', jenkins: '2.236']
+])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,4 @@
-buildPlugin(timeout: 120, useAci: true, platforms: [
+buildPlugin(timeout: 120, useAci: true, configurations: [
   [ platform: 'linux', jdk: '8'],
   [ platform: 'windows', jdk: '8'],
   [ platform: 'linux', jdk: '11', jenkins: '2.236']
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -34,6 +34,7 @@ import hudson.ExtensionList
 import hudson.model.*
 import hudson.util.Secret
 import hudson.triggers.Trigger
+import java.util.function.Function
 import jenkins.model.Jenkins
 import org.apache.commons.codec.digest.DigestUtils
 import org.codehaus.groovy.ast.ASTNode
@@ -457,24 +458,24 @@ class Utils {
         DescribableModel<T> model = new DescribableModel<>(c)
         // Special case for JENKINS-63499.
         if (model != null && model.type == PasswordParameterDefinition.class && model.getParameter("defaultValueAsSecret") != null) {
-            args = transformPasswordParameterDefinitionArgs(args)
+            args = copyMapReplacingEntry(args, "defaultValue", "defaultValueAsSecret", String.class, { s -> Secret.fromString(s) })
         }
         return model?.instantiate(args)
     }
 
     /**
-     * Convert structure of password parameter definition arguments to fix JENKINS-63499.
+     * Copy a map, replacing the entry with the specified key if it matches the specified type.
      */
-    private static <T> Map<String, Object> transformPasswordParameterDefinitionArgs(Map<String, ?> args) {
-        Map<String, Object> newArgs = new HashMap<>()
-        for (Map.Entry<String, ?> entry : args.entrySet()) {
-            if (entry.getKey().equals("defaultValue") && entry.getValue() instanceof String) {
-                newArgs.put("defaultValueAsSecret", Secret.fromString((String)entry.getValue()))
+    public static <T> Map<String, Object> copyMapReplacingEntry(Map<String, ?> map, String oldKey, String newKey, Class<T> requiredValueType, Function<T, Object> replacer) {
+        Map<String, Object> newMap = new TreeMap<>()
+        for (Map.Entry<String, ?> entry : map.entrySet()) {
+            if (entry.getKey().equals(oldKey) && requiredValueType.isInstance(entry.getValue())) {
+                newMap.put(newKey, replacer.apply(requiredValueType.cast(entry.getValue())))
             } else {
-                newArgs.put(entry.getKey(), entry.getValue())
+                newMap.put(entry.getKey(), entry.getValue())
             }
         }
-        newArgs
+        newMap
     }
 
     /**
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -32,6 +32,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
 import hudson.BulkChange
 import hudson.ExtensionList
 import hudson.model.*
+import hudson.util.Secret
 import hudson.triggers.Trigger
 import jenkins.model.Jenkins
 import org.apache.commons.codec.digest.DigestUtils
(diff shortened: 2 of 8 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -454,9 +455,28 @@ class Utils {
     @Restricted(NoExternalUse.class)
     static <T> T instantiateDescribable(Class<T> c, Map<String, ?> args) {
         DescribableModel<T> model = new DescribableModel<>(c)
+        // Special case for JENKINS-63499.
+        if (model != null && model.type == PasswordParameterDefinition.class && model.getParameter("defaultValueAsSecret") != null) {
```
dwnusbaum, line 459: `model.getParameter("defaultValueAsSecret") != null` makes sure that we only run this code when running Jenkins 2.236 or newer.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -454,9 +455,28 @@ class Utils {
...
     }
 
+    /**
+     * Convert structure of password parameter definition arguments to fix JENKINS-63499.
+     */
+    private static <T> Map<String, Object> transformPasswordParameterDefinitionArgs(Map<String, ?> args) {
+        Map<String, Object> newArgs = new HashMap<>()
+        for (Map.Entry<String, ?> entry : args.entrySet()) {
+            if (entry.getKey().equals("defaultValue") && entry.getValue() instanceof String) {
+                newArgs.put("defaultValueAsSecret", Secret.fromString((String)entry.getValue()))
```
dwnusbaum, line 472: Copy all arguments, but replace `defaultValue: 'foo'` with `defaultValueAsSecret: Secret.fromString('foo')` so that data binding succeeds.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/validator/ModelValidatorImpl.groovy
```diff
@@ -455,6 +456,14 @@ class ModelValidatorImpl implements ModelValidator {
                         }
                         ModelASTKeyValueOrMethodCallPair kvm = (ModelASTKeyValueOrMethodCallPair) a
 
+                        // Special case, see JENKINS-63499.
+                        if (model.type == PasswordParameterDefinition.class && "defaultValue".equals(kvm.key.key)) {
+                            if (!validateParameterType((ModelASTValue) kvm.value, String.class, kvm.key)) {
```
dwnusbaum, line 461: The validator is smart and checks to make sure that all parameters exist and have the expected types. This is a workaround because `model.getParameter('defaultValue')` will return `null` on Jenkins 2.236 and newer causing validation to fail. On earlier versions of Jenkins, this is just a shortcut for the check that would have happened anyway.

Jenkinsfile
```diff
@@ -1 +1,5 @@
-buildPlugin(timeout: 120)
+buildPlugin(timeout: 120, useAci: true, platforms: [
```
dwnusbaum, line 1: Oops, forgot to fix locally:

```suggestion
buildPlugin(timeout: 120, useAci: true, configurations: [
```

dwnusbaum:
@bitwiseman From a quick look I think that the directive generator should be fixable with some copy/paste from https://github.com/jenkinsci/pipeline-input-step-plugin/pull/46, so I'll try to fix it here.

CI failure looks infra-related: 

> Cannot contact aci-maven-windows-jggx1: hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@953601:JNLP4-connect connection from 104.46.124.118/104.46.124.118:6273": Remote call on JNLP4-connect connection from 104.46.124.118/104.46.124.118:6273 failed. The channel is closing down or has closed down

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/ParametersDirective.java
```diff
@@ -81,11 +86,23 @@ public String getDisplayName() {
         public String toGroovy(@NonNull ParametersDirective directive) {
             StringBuilder result = new StringBuilder("parameters {\n");
             for (ParameterDefinition param : directive.parameters) {
-                result.append(Snippetizer.object2Groovy(UninstantiatedDescribable.from(param)));
+                result.append(Snippetizer.object2Groovy(customUninstantiate(param)));
```
dwnusbaum, line 89: This is the fix for the directive generator. I wanted to add a test, but the test harness assumes that things round-trip, and as far as I know the directive generator is one-way only. Maybe things could have been tweaked to get the test passing by adding some instantiation-related code here that would only be used only in tests, but I'm not sure.

dwnusbaum:
Infra issues with the Windows branch, rebuilding:

> [WARNING] Corrupted STDOUT by directly writing to native stream in forked JVM 1. See FAQ web page and the dump file C:\Users\jenkins\Work\workspace\e-model-definition-plugin_PR-404\pipeline-model-definition\target\surefire-reports\2020-08-28T14-33-17_907-jvmRun1.dumpstream
> [2020-08-28T14:40:11.092Z] Cannot contact aci-maven-windows-8g6mv: hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@3bf9942d:JNLP4-connect connection from 52.247.2.124/52.247.2.124:17153": Remote call on JNLP4-connect connection from 52.247.2.124/52.247.2.124:17153 failed. The channel is closing down or has closed down

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -454,9 +456,28 @@ class Utils {
...
+    public static <T> Map<String, Object> copyMapReplacingEntry(Map<String, ?> map, String oldKey, String newKey, Class<T> requiredValueType, Function<T, Object> replacer) {
+        Map<String, Object> newMap = new TreeMap<>()
+        for (Map.Entry<String, ?> entry : map.entrySet()) {
+            if (entry.getKey().equals(oldKey) && requiredValueType.isInstance(entry.getValue())) {
+                newMap.put(newKey, replacer.apply(requiredValueType.cast(entry.getValue())))
+            } else {
+                newMap.put(entry.getKey(), entry.getValue())
+            }
+        }
+        newMap
```
car-roll, line 478: How come there is no `return` keyword here?

dwnusbaum: Because this is Groovy 😢 . If I had some spare time on my hands I would like to convert everything in `src/main/groovy` to Java for better IDE support and to avoid confusion (see also a few lines up in the other method where I have to pass a closure to a method expecting a `Function`, since Groovy doesn't support Java 8 lambdas directly, instead it coerces the closure to the interface type)

car-roll: My bad, I totally missed that!

dwnusbaum:
Ok @bitwiseman the `CredentialWrapperStepTest.sshCredentialsInEnv` failure here when testing against Jenkins 2.236 is awkward. It is related to the trilead-api split in Jenkins 2.190 and the version of the `git-client` plugin being used. 

The fix would be to update the baseline here to 2.204.1 and then pick up a version of `git-client` with https://github.com/jenkinsci/git-client-plugin/pull/532. With the 2.249.x LTS coming out, that would be the fourth-oldest LTS line, so it's not totally unreasonable, but I'm not sure what you want to do.

If we don't want to do that just yet, we could either ignore that test temporarily (removing test coverage for JENKINS-52850), or just not test against Jenkins 2.236 (meaning we wouldn't have coverage for the changes in this PR).

bitwiseman:
@dwnusbaum 
I say we ignore the test and immediately create a PR to reenable it and move to 2.204.1, which we'll merge in the next month or so. We're not seeing a lot of churn in this plugin (nor in that area of this plugin), so I'm comfortable with creating a short term test hole to get this released.

PR #405, opened 2020-09-19
[JENKINS-34987] - Allow better script splitting in more scenarios

bitwiseman:
## JENKINS issue(s):

* [JENKINS-34987](https://issues.jenkins-ci.org/browse/JENKINS-37984)

## Description:

Automatically splitting the pipeline script into multiple smaller parts makes it less likely
for the resulting compiled class throw "Method code too large!" or "Class too large"
exceptions during the class file generation phase of compilation.

Script splitting has two different strategies: method-grouped and closure-only. Closure-only is less effective but method-grouped splitting is not compatible with some groovy structures.  In particular locally declared variables, such as such as `def a = 1`, are incompatible with the more effective method-grouped script splitting.

Before this change, any statements outside the pipeline directive would cause the plugin to switch to using the less effective closure-only script splitting.

This change does three things:
1. It uses the method-grouped script splitting unless it detects locally declared variables such as `def a = 1` outside the `pipeline {}` directive. 
2. If it detects locally declared variables such as `def a = 1`:
a. it defaults to failing the run telling the user to add the `@Field` annotation to the declaration. 
b. If `SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true`, it does not report an error, instead falling back to closure-only splitting.  

Script splitting ignores other non-problematic statements such as method/function declarations or method calls.  Note, this is experimental. There may be other structures aside from local variable declaration that are incompatible with script splitting.  More testing is needed.

WARNING: this is a global change. If you install this version an have `SCRIPT_SPLITTING_TRANSFORMATION=true`, any declarative pipeline Jenkinsfile with a local variable declaration outside the pipeline block will fail.  You may switch back to allowing locally declared variables by setting `SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true`.

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
@@ -1262,7 +1262,7 @@ class RuntimeASTTransformer {
                     }
                 }
 
-                if (declarations.size() != 0) {
+                if (!declarations.isEmpty()) {
                     if (SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES) {
                         result.addAll(pipelineElementHandles)
                         pipelineElementHandles.clear()
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
@@ -76,6 +76,16 @@ class RuntimeASTTransformer {
             false
     )
 
+    /**
+     * Enables or disables allowing local variable declarations while script splitting.
+     * This severely reduces the effectiveness of script splitting.
+     */
+    @SuppressFBWarnings(value="MS_SHOULD_BE_FINAL", justification="For access from script console")
+    public static boolean SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES = SystemProperties.getBoolean(
+            RuntimeASTTransformer.class.getName() + ". SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES",
+            false
+    )
+
     Wrapper wrapper = null
 
     RuntimeASTTransformer() {
@@ -1156,7 +1166,7 @@ class RuntimeASTTransformer {
          * nor from closures defined in other functions or classes.  Thus, when the wrapper puts the closure declarations
          * into functions they can no longer access script-local "def" variables.
          *
-         * To maintain some support for local variables, this method detects the presnce of script-local "def" variables
+         * To maintain some support for local variables, this method detects the presence of script-local "def" variables
          * and adds handles to closures instead of methods.
          *
          * Currently, it only checks if "pipeline {}" is not the only top level element in script, and in that case it
@@ -1165,20 +1175,21 @@ class RuntimeASTTransformer {
          * This solution is sufficient for now, but a better solution would be to:
          *
          * * Specifically detect that local variables are used
-         * * Log a warning that this is no advised
+         * * Log a warning that this is not advised
          * * detect which handles reference local variables and then only declare those handles in closures.
          *
          * @param pipelineBlock the block statement to add declarations to
          */
         @NonNull
         private void declareClosureScopedHandles(@NonNull BlockStatement pipelineBlock) {
-            if (moduleNode.statementBlock.statements.size() == 1 || pipelineElementHandles.size() == 0) {
+            def closureScopedHandles = prepareClosureScopedHandles()
+            if (closureScopedHandles.size() == 0) {
                 return
             }
 
             BlockStatement currentBlock = block()
             int count = 0
-            pipelineElementHandles.each { item ->
+            closureScopedHandles.each { item ->
                 if (count++ >= declarationGroupSize) {
                     count = 1
                     pipelineBlock.addStatement(stmt(callX(closureX(currentBlock), 'call')))
@@ -1190,10 +1201,37 @@ class RuntimeASTTransformer {
             // These variable handles are declared in functions, but will still be bound to the script context
             // Doing this here ensures the variables make the trip across to run time - if they don't make it, neither did this pipeline
             pipelineBlock.addStatement(stmt(callX(closureX(currentBlock), 'call')))
-            pipelineElementHandles.clear()
         }
 
-        /**
+        @NonNull
+        private List<Statement> prepareClosureScopedHandles(@NonNull BlockStatement pipelineBlock) {
+            ArrayList<Statement> result = new ArrayList<Statement>()
+            if (SCRIPT_SPLITTING_TRANSFORMATION) {
+                ArrayList<DeclarationExpression> declarations = new ArrayList<DeclarationExpression>()
+                moduleNode.statementBlock.statements.each { item ->
+                    if (((ExpressionStatement) item).expression instanceof DeclarationExpression) {
+                        declarations.add((DeclarationExpression) ((ExpressionStatement) item).expression)
+                    }
+                }
+
+                if (declarations.size() != 0) {
+                    if (SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES) {
+                        result.addAll(pipelineElementHandles)
+                        pipelineElementHandles.clear()
+                    } else {
+                        throw new IllegalStateException("SCRIPT_SPLITTING_TRANSFORMATION is incompatible with local variable declarations. " +
+                                "Add the the '@Field' annotation to local variable declarations: " +
+                                declarations.join(", ") + ".")
+                    }
+                }
+            }
+
+            // In a future version, it may be possible to detect closures that reference script-local variables
+            // and then only use closure scoped handles for those closures.
+            return result
+        }
+
+            /**
          * Adds groups of handle declarations to functions and adds calls to those functions to the pipeline block.
          * Avoid "method code too large" errors and other compiler breaks related to Groovy, JVM, and CPS limitations.
          * @param pipelineBlock
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
@@ -1268,7 +1268,7 @@ class RuntimeASTTransformer {
                         pipelineElementHandles.clear()
                     } else {
                         throw new IllegalStateException("[JENKINS-34987] SCRIPT_SPLITTING_TRANSFORMATION is an experimental feature of Declarative Pipeline and is incompatible with local variable declarations inside a Jenkinsfile. " +
-                                "Local variable declarations found: " + declarations.join(", ") + ". "
+                                "Local variable declarations found: " + declarations.join(", ") + ". " +
                                  "As a temporary workaround, you can add the '@Field' annotation to these local variable declarations. " +
                                  "However, use of Groovy variables in Declarative pipeline, with or without the '@Field' annotation, is not recommended or supported. " + 
                                  "To use less effective script splitting which allows local variable declarations without changing your pipeline code, set SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true .")
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
@@ -1268,8 +1268,10 @@ class RuntimeASTTransformer {
                         pipelineElementHandles.clear()
                     } else {
                         throw new IllegalStateException("[JENKINS-34987] SCRIPT_SPLITTING_TRANSFORMATION is an experimental feature of Declarative Pipeline and is incompatible with local variable declarations inside a Jenkinsfile. " +
-                                "Add the '@Field' annotation to local variable declarations: " +
-                                declarations.join(", ") + ". To use less effective script splitting which allows local variable declarations without changing your pipeline code, set SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true .")
+                                "Local variable declarations found: " + declarations.join(", ") + ". "
+                                 "As a temporary workaround, you can add the '@Field' annotation to these local variable declarations. " +
+                                 "However, use of Groovy variables in Declarative pipeline, with or without the '@Field' annotation, is not recommended or supported. " + 
+                                 "To use less effective script splitting which allows local variable declarations without changing your pipeline code, set SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true .")
                     }
                 }
             }
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
@@ -1267,9 +1267,9 @@ class RuntimeASTTransformer {
                         result.addAll(pipelineElementHandles)
                         pipelineElementHandles.clear()
                     } else {
-                        throw new IllegalStateException("SCRIPT_SPLITTING_TRANSFORMATION is incompatible with local variable declarations. " +
-                                "Add the the '@Field' annotation to local variable declarations: " +
-                                declarations.join(", ") + ".")
+                        throw new IllegalStateException("[JENKINS-34987] SCRIPT_SPLITTING_TRANSFORMATION is an experimental feature of Declarative Pipeline and is incompatible with local variable declarations inside a Jenkinsfile. " +
+                                "Add the '@Field' annotation to local variable declarations: " +
+                                declarations.join(", ") + ". To use less effective script splitting which allows local variable declarations without changing your pipeline code, set SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true .")
                     }
                 }
             }
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
@@ -71,6 +71,7 @@ import static org.junit.Assert.assertThat;
 public abstract class AbstractModelDefTest extends AbstractDeclarativeTest {
 
     private boolean defaultScriptSplitting = RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION;
+    private boolean defaultScriptSplittingAllowLocalVariables = RuntimeASTTransformer.SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES;
 
     @ClassRule
     public static BuildWatcher buildWatcher = new BuildWatcher();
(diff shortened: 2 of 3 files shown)


pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -76,6 +76,16 @@ class RuntimeASTTransformer {
             false
     )
 
+    /**
+     * Enables or disables allowing local variable declarations while script splitting.
+     * This severely reduces the effectiveness of script splitting.
+     */
+    @SuppressFBWarnings(value="MS_SHOULD_BE_FINAL", justification="For access from script console")
+    public static boolean SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES = SystemProperties.getBoolean(
+            RuntimeASTTransformer.class.getName() + ".SCRIPT_SPLITTING_TRANSFORMATION",
```
bitwiseman, line 85:
```suggestion
            RuntimeASTTransformer.class.getName() + ". SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES",
```

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
```diff
@@ -111,14 +112,18 @@ private static String symbolFromDescriptor(Descriptor d) {
...
         // For testing we want to default to exercising splitting
+        // and not allowing local variables
         RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION = true;
+        RuntimeASTTransformer.SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES = true;
     }
 
     @After
     public void cleanupFeatureFlags() {
         RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION = defaultScriptSplitting;
+        RuntimeASTTransformer.SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES = defaultScriptSplittingAllowLocalVariables;
```
jglick, line 126: Try https://github.com/jenkinsci/jenkins-test-harness/pull/242

bitwiseman:
To test this change locally: 

Install the following incrementals:
https://repo.jenkins-ci.org/incrementals/org/jenkinsci/plugins/pipeline-model-api/1.7.3-rc1872.9504c794d213/
https://repo.jenkins-ci.org/incrementals/org/jenkinsci/plugins/pipeline-model-definition/1.7.3-rc1872.9504c794d213/
https://repo.jenkins-ci.org/incrementals/org/jenkinsci/plugins/pipeline-model-extensions/1.7.3-rc1872.9504c794d213/
https://repo.jenkins-ci.org/incrementals/org/jenkinsci/plugins/pipeline-stage-tags-metadata/1.7.3-rc1872.9504c794d213/

(Sorry, yes, you need to do all four of them.)

Then via command-line ops or in Jenkins script console set the flag: `org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION=true`

WARNING: this is a global change. If you install this version and have `SCRIPT_SPLITTING_TRANSFORMATION=true`, any **declarative** pipeline `Jenkinsfile` with a local variable declaration (`def varName="value"`) outside the pipeline block will fail. This is intentional, keeping users from accidentally defeating the improved script splitting.  The fix is to add an `@Field` annotation - `@Field def varName="value"`. You may switch back to allowing locally declared variables by setting `org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true`.

jglick:
@bitwiseman note that if you want Incrementals deployment, you need to merge in `master`.

henryborchers:
I've testing this out in a dockerfile container that I extended from jenkinsfile-runner. The jenkinsfile-runner was able to load the pipeline that was originally failing from being too large.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -1238,10 +1249,37 @@ class RuntimeASTTransformer {
...
+                }
+
+                if (declarations.size() != 0) {
+                    if (SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES) {
+                        result.addAll(pipelineElementHandles)
+                        pipelineElementHandles.clear()
+                    } else {
+                        throw new IllegalStateException("SCRIPT_SPLITTING_TRANSFORMATION is incompatible with local variable declarations. " +
+                                "Add the the '@Field' annotation to local variable declarations: " +
+                                declarations.join(", ") + ".")
```
bitwiseman, line 1272:
```suggestion
                        throw new IllegalStateException("[JENKINS-34987] SCRIPT_SPLITTING_TRANSFORMATION is an experimental feature of Declarative Pipeline and is incompatible with local variable declarations inside a Jenkinsfile. " +
                                "Add the the '@Field' annotation to local variable declarations: " +
                                declarations.join(", ") + ".  To partially disable script splitting to allow declarations without changing your pipeline code, set SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true .")
```

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -104,6 +104,16 @@ public void stages100WithOutsideVarAndFunc() throws Exception {
             .go();
     }
 
+    @Issue("JENKINS-37984")
+    @Test
+    public void stages100WithOutsideVarAndFuncNotAllowed() throws Exception {
+        RuntimeASTTransformer.SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES = false;
+        expect(Result.FAILURE,"basic/stages100WithOutsideVarAndFunc")
+            .logContains("Add the the '@Field' annotation to local variable declarations")
```
bitwiseman, line 112:
```suggestion
            .logContains("Add the '@Field' annotation to local variable declarations")
```

bitwiseman:
@dwnusbaum @timja 
I'd like to include this in the release.  It is experimental behavior gated behind a feature flag. Failures are known flaky tests.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -1238,10 +1249,37 @@ class RuntimeASTTransformer {
...
+            ArrayList<Statement> result = new ArrayList<Statement>()
+            if (SCRIPT_SPLITTING_TRANSFORMATION) {
+                ArrayList<DeclarationExpression> declarations = new ArrayList<DeclarationExpression>()
+                moduleNode.statementBlock.statements.each { item ->
+                    if (((ExpressionStatement) item).expression instanceof DeclarationExpression) {
+                        declarations.add((DeclarationExpression) ((ExpressionStatement) item).expression)
+                    }
+                }
+
+                if (declarations.size() != 0) {
```
timja, line 1265:
```suggestion
                if (!declarations.isEmpty()) {
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -1238,10 +1249,37 @@ class RuntimeASTTransformer {
...
+                    if (SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES) {
+                        result.addAll(pipelineElementHandles)
+                        pipelineElementHandles.clear()
+                    } else {
+                        throw new IllegalStateException("[JENKINS-34987] SCRIPT_SPLITTING_TRANSFORMATION is an experimental feature of Declarative Pipeline and is incompatible with local variable declarations inside a Jenkinsfile. " +
+                                "Add the '@Field' annotation to local variable declarations: " +
```
dwnusbaum, line 1271: This recommendation seems weird to me. Even if turning local variables into fields using `@Field` makes it possible to do the script splitting transformation, I don't think it's something we should recommend, since this message makes it sound like using `@Field` is something we endorse as a good idea when really we do not recommend using Groovy variables with Declarative at all.

timja (approved):

dwnusbaum (approved):
In general the change seems fine since it is all gated behind `SCRIPT_SPLITTING_TRANSFORMATION` anyway, but I feel like the message about using `@Field` is misleading because it implies that we endorse using `@Field` with Declarative. My view is that any use of Groovy with Declarative Pipeline apart from `script` steps is unsupported and should not make any related guarantees or recommendations about things that happen to work right now.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
bitwiseman: However, we do not provide any alternative facility, so this is what people do.  Given the number of people that have continued to complain about this issue, it is clear that there are plenty of people using Declarative with variables (otherwise the existing feature would probably have worked for them).  I'll update the message to be clear that this is not supported.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -1238,10 +1249,37 @@ class RuntimeASTTransformer {
...
+                                declarations.join(", ") + ". To use less effective script splitting which allows local variable declarations without changing your pipeline code, set SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true .")
```
bitwiseman, line 1272:
```suggestion
                        throw new IllegalStateException("[JENKINS-34987] SCRIPT_SPLITTING_TRANSFORMATION is an experimental feature of Declarative Pipeline and is incompatible with local variable declarations inside a Jenkinsfile. " +
                                "Local variable declarations found: " +
                                declarations.join(", ") + ". Use of Groovy var 
                                 "As a temporary workaround, you can add the '@Field' annotation to these local variable declarations. " +
                                 "However, use of Groovy variables in Declarative pipeline, with or without '@Field' annotation, is not recommended or supported. " + 
                                 "To use less effective script splitting which allows local variable declarations without changing your pipeline code, set SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true .")
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -1238,10 +1249,39 @@ class RuntimeASTTransformer {
...
+                    }
+                }
+
+                if (!declarations.isEmpty()) {
+                    if (SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES) {
+                        result.addAll(pipelineElementHandles)
+                        pipelineElementHandles.clear()
+                    } else {
+                        throw new IllegalStateException("[JENKINS-34987] SCRIPT_SPLITTING_TRANSFORMATION is an experimental feature of Declarative Pipeline and is incompatible with local variable declarations inside a Jenkinsfile. " +
+                                "Local variable declarations found: " + declarations.join(", ") + ". "
```
dwnusbaum, line 1271: I think the compilation error is because you are missing a `+` here:

```suggestion
                                "Local variable declarations found: " + declarations.join(", ") + ". " +
```

bitwiseman: Dang it, I knew I was going to screw that up.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -104,6 +104,16 @@ public void stages100WithOutsideVarAndFunc() throws Exception {
             .go();
     }
 
+    @Issue("JENKINS-37984")
+    @Test
+    public void stages100WithOutsideVarAndFuncNotAllowed() throws Exception {
+        RuntimeASTTransformer.SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES = false;
+        expect(Result.FAILURE,"basic/stages100WithOutsideVarAndFunc")
+            .logContains("Add the '@Field' annotation to local variable declarations")
```
bitwiseman, line 112:
```suggestion
            .logContains("add the '@Field' annotation to these local variable declarations")
```

bitwiseman:
OMG, green build finally.

eplodn:
@bitwiseman was it released recently and could it be that pipelines working previously now fail with cryptic error messages because of that? I moved to it an hour ago and rolled it back (should be mentioned: with other plugins too.)

eplodn:
Another pipeline is failing with a similarly cryptic message under 1.8.0. After a rollback to 1.7.2 it works OK as previously:
```

GitHub has been notified of this commit’s build result

org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
General error during semantic analysis: Cannot cast object 'org.codehaus.groovy.ast.stmt.IfStatement@41ca087b' with class 'org.codehaus.groovy.ast.stmt.IfStatement' to class 'org.codehaus.groovy.ast.stmt.ExpressionStatement'

org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object 'org.codehaus.groovy.ast.stmt.IfStatement@41ca087b' with class 'org.codehaus.groovy.ast.stmt.IfStatement' to class 'org.codehaus.groovy.ast.stmt.ExpressionStatement'
	at org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation.continueCastOnSAM(DefaultTypeTransformation.java:405)
	at org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation.continueCastOnNumber(DefaultTypeTransformation.java:319)
	at org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation.castToType(DefaultTypeTransformation.java:232)
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.castToType(ScriptBytecodeAdapter.java:603)
	at org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer$Wrapper$_prepareClosureScopedHandles_closure2.doCall(RuntimeASTTransformer.groovy:1260)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
```

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -1238,10 +1249,39 @@ class RuntimeASTTransformer {
...
         }
 
-        /**
+        @NonNull
+        private List<Statement> prepareClosureScopedHandles(@NonNull BlockStatement pipelineBlock) {
+            ArrayList<Statement> result = new ArrayList<Statement>()
+            if (SCRIPT_SPLITTING_TRANSFORMATION) {
+                ArrayList<DeclarationExpression> declarations = new ArrayList<DeclarationExpression>()
+                moduleNode.statementBlock.statements.each { item ->
+                    if (((ExpressionStatement) item).expression instanceof DeclarationExpression) {
```
dwnusbaum, line 1260: The error mentioned by @eplodn in https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/405#issuecomment-771677208 is because this is an unchecked cast, so anyone using `SCRIPT_SPLITTING_TRANSFORMATION` is probably hitting similar issues in the latest release. I will file a PR in a minute.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -1238,10 +1249,39 @@ class RuntimeASTTransformer {
...
+                    }
+                }
+
+                if (!declarations.isEmpty()) {
+                    if (SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES) {
+                        result.addAll(pipelineElementHandles)
+                        pipelineElementHandles.clear()
+                    } else {
+                        throw new IllegalStateException("[JENKINS-34987] SCRIPT_SPLITTING_TRANSFORMATION is an experimental feature of Declarative Pipeline and is incompatible with local variable declarations inside a Jenkinsfile. " +
+                                "Local variable declarations found: " + declarations.join(", ") + ". " +
```
dwnusbaum, line 1271: https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/405#issuecomment-771657059 shows that the output here is confusing. I think that since Pipeline doesn't support Groovy's multiple assignment feature (`def (x, y) = [1, 2]`), you could get the variable name with code like this to make the message easier to read:

`declarations.stream().map(DeclarationExpression::getVariableExpression).map(VariableExpression::getName).collect(Collectors.joining(","))`

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -77,6 +77,16 @@ class RuntimeASTTransformer {
             false
     )
 
+    /**
+     * Enables or disables allowing local variable declarations while script splitting.
+     * This severely reduces the effectiveness of script splitting.
+     */
+    @SuppressFBWarnings(value="MS_SHOULD_BE_FINAL", justification="For access from script console")
+    public static boolean SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES = SystemProperties.getBoolean(
+            RuntimeASTTransformer.class.getName() + ". SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES",
```
Benjaminvdv, line 86: Line 86 contains a space before `SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES`.

bitwiseman: @Benjamindv
Thank you for finding this. Perhaps next time open a PR.  Commenting on closed PRs and issues is a great way for your comments to not be seen.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
Moskovych: @bitwiseman 
Is this definition `SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES=true` will have permanent affect or it is also implemented as temporary workaround?

bitwiseman: @Moskovych
I'm not sure I understand your question.  The setting will have effect for as long as you keep it.  If you turn it off, pipelines that were broken before will be broken again.   

The question of whether this setting will be there in future versions is not decided yet.  Script splitting as a whole is still experimental and may change.

Moskovych: Let's assume I didn't enabled anything. So means by default it's disabled.
Just updated plugins to 1.8.2. So why the pipelines broken then?

PR #406, opened 2020-09-24
Allow ModelParser to be run outside of a Jenkins instance

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * I'm working on tooling for converting Jenkinsfiles to JSON outside of a Jenkins instance, and there are a number of things that break as things stand now. This fixes those, by allowing for providing a `ModelValidator` implementation and a `DescriptorLookupCache` instance to `ModelParser`, and uses a fallback for the zero-arg agent model map if run outside of Jenkins.
* Documentation changes:
    * n/a - this isn't user-facing.
* Users/aliases to notify:
    * @bitwiseman

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -89,6 +89,8 @@ class ModelParser implements Parser {
 
     private final Run<?,?> build
 
+    private final Map<String,DescribableModel> zeroArgModels = new HashMap<>()
+
     @Deprecated
     ModelParser(SourceUnit sourceUnit) {
         this(sourceUnit, [], null)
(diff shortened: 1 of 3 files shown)


bitwiseman (approved):

bitwiseman:
I assume you'll add a test before merge.

abayer:
@bitwiseman Yup, just added, along with `AlwaysTrueValidator` since it's needed for the test and would be used for at least the simple case of parsing outside of Jenkins.

abayer:
Hrm, tests are still crashing on Windows. Not sure why - gonna try opening a PR without any code changes to see if it's specific to this change.

abayer:
Looks like it is something specific to this change - it seems to be passing on Windows without this change. I'll dig into it.

abayer:
The failure is in setting up a local agent on the Windows boxes, which...yeah, I don't get it at all. Nothing changed related to that here - nothing here could actually change anything relating to that, really, since it's `JenkinsRule` and core that are relevant. So...I guess I'll just keep trying to rebuild.

abayer:
...I am so confused.

bitwiseman:
@abayer I'm seeing the same failure on #405.

abayer:
The current run may actually be working! Finally!

bitwiseman:
@abayer Do you plan to write a test?

abayer:
@bitwiseman - https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/406/files#diff-9e57fed65a85ab8deb8e7034fbe286c1

bitwiseman:
🤦

PR #410, opened 2021-01-20
JENKINS-62808 - Update for tables to divs

bitwiseman:
* JENKINS issue(s):
    * JENKINS-62808 was fixed as it only required 
* Description:
    * JENKINS-62808 was fixed as it only required test changes.  This PR replaces all tables with divs just for completeness. 
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
@@ -24,7 +24,7 @@
   -->
 
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core" xmlns:l="/lib/layout" xmlns:f="/lib/form" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:local="local">
+<j:jelly xmlns:j="jelly:core" xmlns:l="/lib/layout" xmlns:f="/lib/form" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:local="local" xmlns:ttd="/org/jenkinsci/plugins/pipeline/modeldefinition/generator">
     <l:layout title="${%Pipeline Syntax: Directive Generator}" norefresh="true">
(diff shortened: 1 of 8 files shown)


bitwiseman:
Manually tested and verified it looks okay in the UI. Now to get a green build.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
```diff
@@ -56,10 +57,10 @@
                                     <j:set var="help" value="${descriptor.helpFile}"/>
                                     <j:set var="inStage" value="false"/>
                                     <j:if test="${help != null}">
-                                        <tr>
-                                            <td colspan="3"/>
+                                        <ttd:row>
+                                            <ttd:td colspan="3"/>
                                             <f:helpLink url="${help}"/>
-                                        </tr>
+                                        </ttd:row>
```
dwnusbaum, line 63: From a quick test against 2.276 that @timja noticed in https://github.com/jenkinsci/workflow-cps-plugin/pull/411#pullrequestreview-578079700: With this change, clicking the help icon for the first directive (`agent`) works correctly, but clicking the icon for any other directive does nothing. I don't see any messages in my browser console when I click the help icons.

Interestingly, if I run this plugin using `master` against Jenkins 2.276, everything works correctly (and the appearance seems to be the same with or without this PR), so maybe we should just leave things as-is? Other than that, perhaps [`helpButtonOnClick` in `hudson-behavior.js`](https://github.com/jenkinsci/jenkins/blob/022a09cec0919e9bed7ce6535e13a19055b4190b/war/src/main/webapp/scripts/hudson-behavior.js#L765-L794) needs to be modified in some way?

dwnusbaum: Oh, also I think the `colspan` attribute in `<ttd:td colspan="3"/>` is ignored since it is not defined as an attribute in the Jelly tag, right?

bitwiseman:
Leaving this unchanged.

PR #411, opened 2021-01-25
In validation and elsewhere, prefer metasteps to other non-step descriptors

abayer:
* JENKINS issue(s):
    * n/a
* Description:
    * We always prefer `StepDescriptor`s to other `Descriptor`s, but can still hit symbol collisions and end up with the wrong `Descriptor` for a metastep symbol. It's not consistent across all cases - I think it has something to do with which `Symbol`-annotated class gets checked first in the symbol lookup, but I really can't be sure. So when we're looking up functions, we should prefer first `StepDescriptor`s, then `Descriptor`s for a metastep, and finally any ol' arbitrary `Descriptor`.
    * Note that I had trouble getting a full end-to-end test to fail correctly - something about the `Symbol`-annotated class check ordering or, worse, some caching of lookups that actually do what they're supposed to do before we even get there. I really can't tell. So there is a test in `pipeline-model-defintion`'s `StepsTest`, but it tends to always pass even when it shouldn't. I've added a new test to `pipeline-model-api`'s `DescriptorCacheLookupTest`, which tests the `lookupFunction` method directly and does fail correctly when expected. 
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-api/pom.xml
+++ b/pipeline-model-api/pom.xml
@@ -69,6 +69,22 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>structs</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-cps</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-job</artifactId>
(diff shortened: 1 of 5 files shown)


jglick (approved):
Looks good from a brief glance.

oleg-nenashev (approved):
Looks good in principle. At the same time it may cause regressions on existing instances which work based on the previous behavior (undocumented one ofc). needs a proper compat notice in the changelog IMHO.

bitwiseman (approved):

PR #412, opened 2021-02-02
Check statement type before casting to ExpressionStatement

dwnusbaum:
* JENKINS issue(s):
    * See https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/405#issuecomment-771677208.
* Description:
    * #405 introduced an unchecked cast to `ExpressionStatement` when `SCRIPT_SPLITTING_TRANSFORMATION` was active which caused failures in Pipelines that had other types of top-level statements. This PR moves that cast behind an `instanceof` check.
    * Side note: Do we not have SpotBugs here? Or does it not work with the code in `src/main/groovy` or something? I think that normally it would catch this kind of issue.
* Documentation changes:
    * N/A
* Users/aliases to notify:
    * @eplodn

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
@@ -1257,8 +1257,11 @@ class RuntimeASTTransformer {
             if (SCRIPT_SPLITTING_TRANSFORMATION) {
                 ArrayList<DeclarationExpression> declarations = new ArrayList<DeclarationExpression>()
                 moduleNode.statementBlock.statements.each { item ->
-                    if (((ExpressionStatement) item).expression instanceof DeclarationExpression) {
-                        declarations.add((DeclarationExpression) ((ExpressionStatement) item).expression)
+                    if (item instanceof ExpressionStatement) {
+                        ExpressionStatement es = (ExpressionStatement) item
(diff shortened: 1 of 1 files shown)


eplodn:
Thanks @dwnusbaum for your quick response. 

Hope you'll excuse my asking: are there any plans in place to somehow address 1.8.0 release in its current state? I am blessed to work for an  organization where we check updates on a dummy Jenkins instance before rolling them into production, but others might not be so lucky.

dwnusbaum:
> Hope you'll excuse my asking: are there any plans in place to somehow address 1.8.0 release in its current state?

We could suspend distribution of that release, but as far as I know it the issue only affects users who are using `SCRIPT_SPLITTING_TRANSFORMATION`, which is essentially an experimental option, so I am not sure whether that is warranted as long we get this PR released today.

Also note that the warning in https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/405#issuecomment-771657059 is different: as of the latest release you must change your Pipeline or configure a new `SCRIPT_SPLITTING_ALLOW_LOCAL_VARIABLES` property in Jenkins to address this error. This should be added as a highlighted warning to the 1.8.0 release notes.

abayer (approved):
Looks good to me.

bitwiseman (approved):

bitwiseman:
@dwnusbaum 
Thank you so much for your fast response. I'm publishing the new release now. 

@eplodn 
As Devin noted, the `SCRIPT_SPLITTING_TRANSFORMATION` is experimental at this time.  This change was intentional.  The only mistake (aside from this bug) was that I did not properly highlight the change in the release notes, which I will go do now.

PR #413, opened 2021-02-02
[JENKINS-34987] Improved output for local variable declarations with script splitting

bitwiseman:
## JENKINS issue(s)
JENKINS-34987 , #405, #412
    
## Description

Per suggestion from @dwnusbaum, I've improved the output for variable declarations to only include the name.  The space was a bit more complex that just a stream statement since `getVariableExpression()` will throw for some types of declarations.
    
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/test/resources/basic/stages100WithOutsideVarAndFunc.groovy
+++ b/pipeline-model-definition/src/test/resources/basic/stages100WithOutsideVarAndFunc.groovy
@@ -23,6 +23,7 @@
  */
 
 def someVar = "Hi there"
+def (firstVar, secondVar) = [1, 2]
 def someFunc() {
     return "This comes from a function"
 }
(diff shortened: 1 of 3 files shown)


pipeline-model-definition/src/test/resources/basic/stages100WithOutsideVarAndFunc.groovy
```diff
@@ -23,6 +23,7 @@
  */
 
 def someVar = "Hi there"
+def (firstVar, secondVar) = [1, 2]
```
dwnusbaum, line 26: Looks like this does work in Pipeline as of [JENKINS-45575](https://issues.jenkins.io/browse/JENKINS-45575). Maybe I was thinking of [JENKINS-55001](https://issues.jenkins.io/browse/JENKINS-55001), which apparently only matters for sandboxed scripts using `script-security` but not `workflow-cps`.

dwnusbaum (approved):

PR #415, opened 2021-02-11
[JENKINS-64846] Fix use of local declared variables inside matrix

bitwiseman:
* JENKINS issue(s):
    * [JENKINS-64846](https://issues.jenkins.io/browse/JENKINS-64846)
* Description:
    * We have script splitting turned on by default in most tests.  The assumption as was that script splitting being turned on would break things, not make them work unexpectedly. Turns out that is not always the case.  With the change to script splitting from 1.8.0, if script splitting is turned off local declared variables do not get retained appropriately in matrix.

* Documentation changes:
* Users/aliases to notify:
    * @abayer

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -192,6 +192,64 @@
     <workflow-multibranch.version>2.16</workflow-multibranch.version>
   </properties>
 
+  <profiles>
+    <profile>
+      <id>no-script-splitting</id>
+      <activation>
+        <jdk>[,1.9)</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <systemPropertyVariables>
+                <org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION>false</org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION>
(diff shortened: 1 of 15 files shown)


bitwiseman:
I'm reviewing existing tests to make sure this scenario is covered.

abayer (approved):
Looks reasonable. I do wonder if we might want to make a profile in the POM for turning script splitting on and run the full suite of tests with and without script splitting on in CI?

bitwiseman:
@abayer 
Sounds like a good idea. I plan to make script splitting the default behavior at some point, but until then we should cover both behaviors.

car-roll (approved):
LGTM, though admittedly I'm just looking more for syntax errors as I'm not completely familiar with the code

abayer (approved):

pom.xml
```diff
@@ -192,6 +192,64 @@
...
+        <jdk>[,1.9)</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <systemPropertyVariables>
+                <org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION>false</org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION>
```
jglick, line 208: I do not get it. Why not just set these flags directly in the code, at runtime, depending on the Java version? Or if you want to test in both modes, use `@Parameterized` etc. See https://github.com/jenkinsci/bom/pull/451#issuecomment-796090220

PR #417, opened 2021-03-11
 Test needs to assume script splitting is enabled

bitwiseman:
This test failure made it past CI because the project [uses the Java version](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pom.xml#L195-L252) to decide whether to enable script splitting or not.  When run outside of CI, that choice is apparently not preserved.


jglick:
See https://github.com/jenkinsci/bom/pull/451#issuecomment-796090220 and https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/415#discussion_r591866854

jglick (approved):
Fine as a hotfix if it works. I suspect the basic problem is the abuse of POM profiles to control test execution. (To start with, recall that `<activation>` is ignored if you explicitly pass a list of profiles on the CLI!) Basically anything “creative” in a POM is a warning sign.

If this script splitting, whatever it is, behaves differently on Java 8 vs. 11, then better to have the test code itself decide whether to run or not. Or if you have some sort of feature flag, then use `FlagRule` to verify behavior with nondefault settings (which ensures that the flag is reset when the test finishes), and as needed use `@Parameterized` or other POJO idioms to run analogous tests with the flag both on and off.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
```diff
@@ -67,7 +67,7 @@ import static org.objectweb.asm.Opcodes.ACC_PUBLIC
  * {@link ExecutionModelAction} to the run.
  */
 @SuppressFBWarnings(value="SE_NO_SERIALVERSIONID")
-class RuntimeASTTransformer {
+public class RuntimeASTTransformer {
```
jtnord, line 70: guessing you do not actually want this to be public (except for test code?) so you can change it at will?
```suggestion
@Restricted(NoExternalUse.class) // for test use only
public class RuntimeASTTransformer {
```

jtnord (approved):
if it works :shrug:

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/RuntimeASTTransformer.groovy
bitwiseman: @jtnord 
Actually, no.  I considered that, but Groovy that is loaded and run in pipeline has difficulty with attributes.  Making this public is not great, but not horrible.

bitwiseman:
This is a temporary solution until script splitting is turned on by default.

PR #155, opened 2017-04-26
[FIXED JENKINS-43872, JENKINS-43910] Yet more env tweaks

abayer:
* JENKINS issue(s):
    * [JENKINS-43872](https://issues.jenkins-ci.org/browse/JENKINS-43872)
    * [JENKINS-43910](https://issues.jenkins-ci.org/browse/JENKINS-43910)
* Description:
    * Hey look, another dumb `environment`/`script.evaluate` bug! I still should probably rewrite all this yet again, but for now...this "escapes" dollar signs in literal values and credentials values before `evaluate`ing non-literal env vars that may contain references to these literal/credential vars, and then "unescapes" the dollar signs back into place after all the pesky `evaluate`ing is done.
    * Yet more circuitous processing to make sure we have credentials (including file credentials) processed before we try to process the environment, and definitely making agent get processed before either credentials or environment, along with an ignored test for [JENKINS-43911](https://issues.jenkins-ci.org/browse/JENKINS-43911), which existed prior to this but is more obvious now.
* Documentation changes:
    * n/a
* Users/aliases to notify:
    * @rsandell 
    * @reviewbybees 
    * @stephendonner since it's another one of your reports. =)

PR #418, opened 2021-05-03
Deleting `CredentialsBindingHandler.toBindings`

jglick:
The call site was deleted in #155 so there seems to be no reason to retain this method. The active https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/4d290aa8d26ac2f3e110fe3cb600ac5fe18fa84c/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy#L463-L475 calls a different method.

Note that https://github.com/jenkinsci/aws-credentials-plugin/blob/2ae45df5330d9e64a574979736def83fa8dece0f/src/main/java/com/cloudbees/jenkins/plugins/awscredentials/AWSDeclarativeCredentialsHandler.java#L41-L47 and https://github.com/jenkinsci/docker-commons-plugin/blob/7078a70448719715ae2c91b6a7fd718de76cf5a2/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsHandler.java#L40-L44 and https://github.com/jenkinsci/xcode-plugin/blob/e57051344470d0336bc077eb5ec10371544c5009/src/main/java/au/com/rayh/XcodeDeclarativeCredentialsHandler.java#L41-L49 still implement it, so this would be source-incompatible for them. (I.e., whenever next updating this plugin dep, the maintainer would need to delete the `@Override`n method.)


bitwiseman (changes requested):
@jglick 
So, the downstream plugins will will still work, but they won't build.  Could you file downstream PRs?

bitwiseman (approved):

jglick:
@bitwiseman ping

PR #440, opened 2021-07-27
Update schema validator to 2.2.14

bitwiseman:
This includes some namespace changes.

* JENKINS issue(s):
    * Issue link or n/a
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...


bitwiseman:
Tried to do an quick update to a more recent version.  It looks like this should wait until the Guava work is done (either updating or supporting isolated loading).

PR #441, opened 2021-07-28
Call `Jenkins.setNodes` to adapt to core change in 2.302

jglick:
See https://github.com/jenkinsci/jenkins/pull/5450#issuecomment-888512618 and following discussion. Whether the new core behavior is a genuine regression is hard to pin down because I am not exactly sure how this test worked before—`Slave.setNumExecutors` updates a field but has no other effect itself.


basil (approved):
I confirmed this gets the test to pass again in 2.294+ at least.

timja (approved):

bitwiseman (approved):

PR #443, opened 2021-07-29
[JENKINS-66247] Properly report an unrecognized `agent` type

jglick:
Formerly this would just throw a NPE, observed in https://github.com/jenkinsci/bom/pull/587.

This patch does _not_ solve the actual bug, which is that https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/f06ef76fb39148bf1ee166f39e2ae191fee81955/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/None.java#L43 is not found, due to https://github.com/jenkinsci/lib-annotation-indexer/pull/10 via https://github.com/jenkinsci/plugin-pom/pull/424.

#442 did not reproduce the issue because then we are compiling `pipeline-model-definition` with the same version of core as we are running tests with. You need to run in PCT to see the problem.


bitwiseman (approved):
Need a test, but otherwise great.

jglick:
The test is running `bom` with and without this patch. :grin:

bitwiseman (approved):

jglick:
see #444

PR #444, opened 2021-07-29
[JENKINS-66247] Use an annotation index compatible with the core baseline

jglick:
Picks up https://github.com/jenkinsci/plugin-pom/pull/426.

PR #445, opened 2021-07-29
[JENKINS-66247] Add a RealJenkinsRule test

bitwiseman:
* JENKINS issue(s):
    * JENKINS-66247 made it into the release because we didn't have even one RealJenkinsRule smoke test.  This PR adds one. 

* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativePipelineTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativePipelineTest.java
@@ -1,0 +1,63 @@
+package org.jenkinsci.plugins.pipeline.modeldefinition;
+
+import groovy.lang.DelegatingMetaClass;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.junit.ClassRule;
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/DeclarativePipelineTest.java
```diff
@@ -0,0 +1,63 @@
...
+public class DeclarativePipelineTest {
+    @Rule
+    public RealJenkinsRule r = new RealJenkinsRule();
+
+    @Test
+    public void smokes() throws Throwable {
+        r.then(DeclarativePipelineTest::smokes_run);
+    }
+
+    public static void smokes_run(JenkinsRule j) throws Throwable {
```
jglick, line 23:
```suggestion
    private static void smokes_run(JenkinsRule j) throws Throwable {
```

jglick (approved):
IIUC this reproduces the bug because it is using the old `Index.class` from `jenkins.war` rather than the new version from the Maven test classpath.

jglick:
see #444

jglick:
I was initially embarrassed that as the author of `RealJenkinsRule` I had not thought of using it here, until I realized that was a stupid attitude and I am glad you found a novel use for it! BTW it currently hard-codes the version of `jenkins.war` tested against to that defined in the test classpath (~ `${jenkins.version}`), which could be changed with a little tricky usage of `dependency:copy`, but I guess that is not necessary here.

PR #462, opened 2021-11-15
Assume that configured agents are online

jglick:
Subclasses of `AbstractModelDefTest` frequently have blocks like https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/3f496434de4cf2fb618bf85d4fa306978a485630/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/ValidatorTest.java#L64-L70 which invariably use `createOnlineSlave`. Yet according to test failures described in https://github.com/jenkinsci/bom/issues/613#issuecomment-969300344, occasionally these agents go offline for reasons not currently understood but probably environmental. Since this is very likely not a bug in tested plugin code, it ought not cause a test failure.


timja (approved):

jetersen (approved):

jglick:
Note that there are a _lot_ of skipped tests in this plugin, but no new skips in this PR that I can see.

car-roll (approved):

PR #459, opened 2021-10-29
Bump `json-schema-validator` from 2.0.4 to 2.2.14

basil:
Downstream of jenkinsci/jenkins#5707.

PR #466, opened 2021-11-17
Use container agent in `Jenkinsfile`

basil:


basil:
@car-roll I think merging this PR would make the build in #459 green.

car-roll (approved):

PR #477, opened 2022-01-07
Skip resource-intensive tests on CI

basil:
This is a total hack, but if it stabilizes CI runs such that we can process the PR queue and then go back later to fix these tests, I suppose that is better than nothing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -224,6 +224,32 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>resource-constrained</id>
+      <activation>
+        <property>
+          <name>env.CI</name>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <!-- TODO flakes on CI for inscrutable reasons -->
+              <excludes>
+                <exclude>org.jenkinsci.plugins.pipeline.modeldefinition.BasicModelDefTest</exclude>
+                <exclude>org.jenkinsci.plugins.pipeline.modeldefinition.MatrixTest</exclude>
+                <exclude>org.jenkinsci.plugins.pipeline.modeldefinition.WhenStageTest</exclude>
+                <exclude>org.jenkinsci.plugins.pipeline.modeldefinition.WhenStageMultibranchTest</exclude>
+                <exclude>org.jenkinsci.plugins.pipeline.modeldefinition.PostStageTest</exclude>
+                <exclude>org.jenkinsci.plugins.pipeline.modeldefinition.ToolsTest</exclude>
+              </excludes>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
   </profiles>
   <repositories>
     <repository>


pom.xml
```diff
@@ -224,6 +224,32 @@
...
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <!-- TODO flakes on CI for inscrutable reasons -->
+              <excludes>
+                <exclude>org.jenkinsci.plugins.pipeline.modeldefinition.BasicModelDefTest</exclude>
```
jglick, line 241: I am afraid this is a big chunk of the test coverage in this plugin. Do we have any more specific list of flakes?

basil: > Do we have any more specific list of flakes?

We do not. I know Jesse, it sucks. I didn't cause this regression in test coverage. I am just trying to help us get back to the point where we can start processing the PR queue again. As people find the time and motivation to look into this, these tests can be re-enabled later.

jglick:
So for example https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/PR-472/2/testReport/ and https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/master/545/testReport/ seem to agree on (certain test cases in) `MatrixTest` on `Windows` being the flakes. I wonder if we could just ignore `MatrixTest`, or skip building on Windows, or something like this.

basil:
Just two runs isn't enough of a sample set. The current list was curated over time over many, many BOM runs. It might not have great coverage but I can assure you that it is stable. :-) I think once we have a "TODO list" of excluded tests and stable builds on the main branch, it becomes a lot easier to move the needle forward. Someone can just try removing an exclusion, open a PR, rebuild it a few times, make sure no new flakiness is introduced, and merge it. Slow and steady wins the race. Of course all that depends on someone having the time and motivation to whack away at the problem, which I currently do not (beyond this PR).

basil:
Build is green

basil:
@car-roll (see #478)

jglick:
Possibly relevant, from @jtnord in https://github.com/jenkinsci/git-plugin/pull/1207

> when consuming this in pipeline-model-definition's ValidatorTest test locally go from failing due to timeout exception (10 minutes) to completing in 2 minutes

jtnord:
> Possibly relevant, from @jtnord in [jenkinsci/git-plugin#1207](https://github.com/jenkinsci/git-plugin/pull/1207)
> 
> > when consuming this in pipeline-model-definition's ValidatorTest test locally go from failing due to timeout exception (10 minutes) to completing in 2 minutes

so locally I have `git-secrets` which is a PITA on windows.  I would hope that the `git` installation on CI is vanilla and not messed with so hopefully this would not matter - but I am trying to get a clean build to get a release for #459 and the CI is not making me confident at the moment.

jglick:
OK so probably not relevant here.

car-roll (approved):

jtnord:
@car-roll I donlt think you should have merged this pending the memory things that are in flight here?

car-roll:
@jtnord yeah, you are right...too premature...
I'll put up another PR to undo

jtnord:
reverted in #480

basil:
> reverted in #480

Right, that was the explicit intention all along. When the test suite can be stabilized, this temporary fix should be reverted.

PR #478, opened 2022-01-07
Suppress `MatrixTest` on Windows CI

jglick:
Does not solve https://github.com/jenkinsci/bom/issues/613 but may at least restore CI in this repo: https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/477#issuecomment-1007614950

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/MatrixTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/MatrixTest.java
@@ -68,7 +68,7 @@ public class MatrixTest extends AbstractModelDefTest {
     private static String password;
     @BeforeClass
     public static void setUpAgent() throws Exception {
-        assumeFalse("TODO frequent failures in CI on Windows", Functions.isWindows());
+        assumeFalse("TODO frequent failures in CI on Windows", Functions.isWindows() && System.getenv("CI") != null);
         s = j.createOnlineSlave();
         s.setLabelString("agent-one some-label");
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/MatrixTest.java
```diff
@@ -66,6 +68,7 @@
     private static String password;
     @BeforeClass
     public static void setUpAgent() throws Exception {
+        assumeFalse("TODO frequent failures in CI on Windows", Functions.isWindows());
```
basil, line 71:
```suggestion
        assumeFalse("TODO frequent failures in CI on Windows", Functions.isWindows() && System.getenv("CI") != null);
```

basil (approved):
:man_shrugging: If it passes

basil:
> If it passes

But it doesn't, while #477 does. :man_shrugging:

jglick:
Hmm, that is a different error than other builds have though.

basil:
So if we lack the time/motivation to take this PR to the finish line for the time being, why not take #477 and deal with this later when it is practical?

jglick:
Probably going to leave that for @car-roll.

jtnord:
> C:\Jenkins\workspace\e-model-definition-plugin_master\pipeline-model-definition\target\tmp\j h3483184721116739069\workspace\test12\subdir>mvn --version 
Error occurred during initialization of VM
Could not reserve enough space for object heap


and we are using [4GB heap for surefire the tests](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/2f4089bc15fb391769eb4dd439b573372301df98/pom.xml#L167)?!?

jtnord:
see also #479 and jenkinsci/jenkins-test-harness#380

PR #479, opened 2022-01-19
The JVM used by maven is unconfigured and uses default ergonomics

jtnord:
this is wastefull of memory and at least on my machine the memory of the
JVM grows above 3GB.

Whilst this impacts `hpi:run` when run from the root of this repo, if run from the directory of a submodule
it will have no effect. Tt can be overridden on the command line
with `MAVEN_OPTS=-Xmx3G mvn hpi:run` (verified on windows with `set MAVEN_OPTS=-Xmx1 & mvn install` and observing a very rapid OOME) 

Heap output of the maven JVM when running `mvn clean install -Denforcer.skip`

image

GC stats for the same command

image

So the memory could be lowered further, but this seems good enough.
There where no adverse effects on my machine with this - maven took about the same time to run (`31:39` without this change `31:58` with - which considering the majority of the time is tests is within the bounds or error noise for me)

* JENKINS issue(s):
    * n/a
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...

@car-roll


jglick:
From the log it looks like there is at least one test failure here, but likely some unrelated flake.

.mvn/jvm.config
```diff
@@ -0,0 +1 @@
+-Xmx512m -XX:+HeapDumpOnOutOfMemoryError -XX:+TieredCompilation -XX:TieredStopAtLevel=1
```
jtnord, line 1:
```suggestion
-Xmx512m -XX:+HeapDumpOnOutOfMemoryError -XX:+TieredCompilation -XX:TieredStopAtLevel=1

```

raul-arabaolaza (approved):

car-roll (approved):

PR #393, opened 2020-05-28
[JENKINS-58732] Remove docker-workflow dependency (refiled)

jglick:
Just the first commit from #391; whatever the purpose of the subsequent commits, they may have been made obsolete by changes in `master` since then.

Follows up #373 & #383.

Admins using `agent docker` / `agent dockerfile` should have updated to at least https://github.com/jenkinsci/docker-workflow-plugin/releases/tag/docker-workflow-1.22 first.

PR #484, opened 2022-01-26
enable cd

car-roll:
Enable cd
see also: https://github.com/jenkins-infra/repository-permissions-updater/pull/2324


pom.xml
```diff
@@ -34,7 +34,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.jenkinsci.plugins</groupId>
   <artifactId>pipeline-model-parent</artifactId>
-  <version>${revision}${changelist}</version>
+  <version>${revision}.${changelist}</version>
```
car-roll, line 37: per @jtnord 's suggestion, going with a 2.{revision}.{changelist}

jglick: Which suggestion? Why?

jtnord: whilst you do not care about semver having a prefix instead of just starting with a huge number means we can more easily to `compatable-since` (can just bump the first number and say since `3.0` without having to constantly update it if something else gets merged first) and well these numbers are meaningless anyway so have a `2.whatever` is no worse than otherwise, and allows us to have the  compatable-since should it ever be needed (so is a good general practice in my opinion).

ie there is a pro with no noticeable drawback as far as I can see, as this gives us more flexibility to do things in the future.

.mvn/maven.config
```diff
@@ -1,2 +1,3 @@
 -Pconsume-incrementals
 -Pmight-produce-incrementals
+-Dchangelist.format=%d.v%s
```
jtnord, line 3: jenkinsci/incrementals-tools#29

jtnord (approved):

pom.xml
jglick: No major drawback, just extra stuff in the version string before what you need to pay attention to, and inconsistency with other plugins.

In the rare case that a `compatibleSince` is added or updated, this is easily supported: leave that version `TODO` in the PR, and when merging, do so by hand and make the merge commit use `$(git rev-list --count master) + 1`.

jtnord: > this is easily supported

but requires a special flow - which I think is good to avoid in general?

jtnord:
build failed as other properties need to be updated.  or we just change the version to be `changelist` and change the changelist format to `Dchangelist.format=2.%d.v%s` which is probably simple for the multi module should we wish to keep the `2.` prefix

pom.xml
jglick: If this is were regular occurrence perhaps. AFAICT here it is not frequent, especially recently (5bb4e3751245c0e741357ef4a635cda115c0ca43, #174, 10f7c8ee77eb0834bd8657d4592f3b593a5fd7de, #393) so I would pick the cosmetically simpler version and deal with the extra few minutes’ work every couple of years. FWIW

.mvn/maven.config
```diff
@@ -1,3 +1,3 @@
 -Pconsume-incrementals
 -Pmight-produce-incrementals
--Dchangelist.format=%d.v%s
+-Dchangelist.format=2.%d.v%s
```
jglick, line 3: Standard

```suggestion
-Dchangelist.format=%d.v%s
```

suffices IMO.

PR #498, opened 2022-03-17
Update plugin parent POM and plugin BOM

basil:
Updates the plugin parent POM from a weekly to an LTS, along with updating the plugin BOM accordingly. This is necessary in order to be able to test this plugin on Java 17, as I am trying to do in jenkinsci/bom#935. The CodeNarc Maven plugin doesn't seem to work at all on Java 17, but it appears to be adding little value so I just removed it. With these changes I can run `mvn clean verify -Dtest=org.jenkinsci.plugins.pipeline.modeldefinition.DeclarativePipelineTest -DfailIfNoTests=false -Djenkins-test-harness.version=1721.v385389722736 '-DargLine=--add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.util.concurrent=ALL-UNNAMED'` successfully on Java 17. CC @car-roll

### DIFF
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -53,25 +53,6 @@
         <artifactId>gmavenplus-plugin</artifactId>
       </plugin>
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>codenarc-maven-plugin</artifactId>
-        <version>0.22-1</version>
-        <configuration>
-          <groovyVersion>2.4.7</groovyVersion>
-          <sourceDirectory>${project.basedir}/src/main/</sourceDirectory>
-          <maxPriority1Violations>0</maxPriority1Violations>
-        </configuration>
-        <executions>
-          <execution>
-            <id>codenarc</id>
-            <goals>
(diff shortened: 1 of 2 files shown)


olamy (approved):

olamy:
windows failure are not related to the change (we should fix but later)
```
remote: Total 6 (delta 1), reused 0 (delta 0), pack-reused 0        
error: unable to write file .git/objects/61/392ad7ea76f42f304663ab3d819624b404bdf2: Filename too long
```

car-roll:
@olamy it does appear the failures are related to the change. See #501 
I reverted this PR (#501 ) and filed #502 so that the issue can be further investigated

olamy:
> @olamy it does appear the failures are related to the change. See #501 I reverted this PR (#501 ) and filed #502 so that the issue can be further investigated

sounds good.
ok as I don't have any windows env for testing, I thought there were some issues on windows only due to checkout of too long paths in Jenkins but maybe something different. 
Thanks to investigate

PR #499, opened 2022-03-17
Stabilize Windows build

basil:
The Windows build appears to have become destabilized recently. This PR attempts to stabilize the build by skipping the failing tests on Windows.

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTest.java
@@ -24,6 +24,7 @@
 package org.jenkinsci.plugins.pipeline.modeldefinition;
 
 import com.cloudbees.hudson.plugins.folder.Folder;
+import hudson.Functions;
 import hudson.model.Result;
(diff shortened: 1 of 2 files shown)


basil:
CC @car-roll

basil:
The incremental build from this PR has been successfully tested with Java 17 in the draft PR jenkinsci/bom#935. In order to help facilitate the transition of that draft PR into non-draft status, it would be helpful for this PR to be merged and released.

car-roll:
Looks like #498 was merged a bit prematurely. But what kind of dependency update would have caused these git `filename` too long errors?

basil:
Not sure offhand.

basil:
> But what kind of dependency update would have caused these git filename too long errors?

@car-roll So rather than answering this question, you reverted the whole PR, setting back my Java 17 efforts?

car-roll:
@basil No, the problem was that #498 should  never have been merged without those issues being resolved in the first place. I went ahead and refiled #502 so we can continue to diagnose the problem. My  concern was that then any additional PRs that would have been merged to declarative that were considered "significant" would then trigger a release via automated CD and then would have caused downstream failures in PCT.

basil:
Closing in favor of #503.

PR #501, opened 2022-03-22
Revert "Update plugin parent POM and plugin BOM"

car-roll:
This reverts commit 573bae01baf906b2f5aacad3da4e33387da3656f.
Identifying #498 as the cause of the recent windows CI failures

* JENKINS issue(s):
    * Issue link or n/a
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/pom.xml
+++ b/pipeline-model-definition/pom.xml
@@ -53,6 +53,25 @@
         <artifactId>gmavenplus-plugin</artifactId>
       </plugin>
       <plugin>
+        <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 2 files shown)


car-roll:
Confirmed that reverting this commit fixes the windows CI failures

jtnord (approved):

basil:
@car-roll Why was the whole PR reverted? Rather I think only the part that caused the Windows CI failures should have been reverted.

basil:
@car-roll Should have instead only reverted the problematic portions of #498 as was done in #503.

PR #502, opened 2022-03-22
Update plugin BOM and jenkins version

car-roll:
This PR refiles the BOM portion of #498 as it was reverted in #501 due to windows CI failures.

Failures generated are `filename too long`. See: https://github.com/jenkinsci/pipeline-model-definition-plugin/runs/5580211853

See also: https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/499

cc @basil

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTest.java
@@ -161,6 +161,7 @@ public class LibrariesTest extends AbstractModelDefTest {
     @Test
     public void folderLibraryParsing() throws Exception {
         otherRepo.init();
+        otherRepo.git("config", "--global", "core.longpaths", "true");
         otherRepo.git("checkout", "-b", "test");
         otherRepo.write("src/org/foo/Zot.groovy", "package org.foo;\n" +
                 "\n" +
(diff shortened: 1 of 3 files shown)


basil:
Let's see if just updating the parent POM as in #503 (ignoring the BOM update and version bump) is sufficient to keep the tests stable and for my Java 17 work.

car-roll:
I am currently testing on my local windows env to see if I can reproduce

basil (changes requested):
Build failed. Suggest closing this PR in favor of #503.

car-roll:
Looks like enable `longpaths` worked. Only did it for one test and it passed all other tests (most likely because of `--global`. Will need to clean this up properly. Either call it once for only windows, or just call it always.

car-roll:
going to keep this PR open to see if we can resolve the BOM bump issues

jglick (approved):

basil (approved):

jglick:
Also see #476, #504, #487 (override should be removed), and probably others.

jglick:
Hmm, seems like #503 should have caused some of those to be automatically closed? :eyes:

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/LibrariesTest.java
```diff
@@ -161,6 +161,8 @@ public void librariesDirectiveWithOutsideVarAndFunc() throws Exception {
     @Test
     public void folderLibraryParsing() throws Exception {
         otherRepo.init();
+        // longpaths to support longer filenames for windows. Only needs to be set once.
+        otherRepo.git("config", "--global", "core.longpaths", "true");
```
jglick, line 165: Can be reverted now that https://github.com/jenkins-infra/pipeline-library/pull/334 is merged?

car-roll: Filed #506

PR #514, opened 2022-04-21
Update plugin parent POM and BOM

basil:


basil:
@jenkinsci/pipeline-model-definition-plugin-developers A release of this would facilitate Java 17 PCT testing.

car-roll (approved):

PR #529, opened 2022-05-17
Use GroovySourceFileAllowlist to adapt to SECURITY-359 changes

dwnusbaum:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/538. The fix for SECURITY-359 added a new API that should be used by plugins that define Groovy DSLs. Declarative defines two main kinds of Groovy DSLs:

* `ModelInterpreter.groovy` via `ModelStepLoader`: The main declarative syntax
* Various Groovy files via subclasses of `WithScriptDescribable`: Extensions related to `agent` (`DeclarativeAgent`, used by other plugins, specifically `docker-workflow`, `kubernetes`, and `amazon-ecs`) and `when` (`DeclarativeStageConditional`).`

This PR implements `GroovySourceFileAllowlist` for various individual files, but also for `WithScriptDescribable` in a generic way that automatically allows the scripts used by subclasses. This allows us to remove entries from `default-allowlist` upstream in `workflow-cps`.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java
@@ -32,6 +32,8 @@ import org.jenkinsci.plugins.workflow.cps.CpsThread;
 import org.jenkinsci.plugins.workflow.cps.GlobalVariable;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Objects;
+import org.jenkinsci.plugins.workflow.cps.GroovySourceFileAllowlist;
 
 /**
  * Loads the main "pipeline" step as well as the additional CPS-transformed code it depends on.
@@ -70,6 +72,16 @@ public class ModelStepLoader extends GlobalVariable {
     @Initializer(after = InitMilestone.EXTENSIONS_AUGMENTED)
     public static void invalidateOptionTypeCaches() {
         Options.invalidateCaches();
+    }
+
+    @Extension
+    public static class ModelInterpreterAllowlist extends GroovySourceFileAllowlist {
+        private final String scriptUrl = Objects.requireNonNull(getClass().getClassLoader().getResource("org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy")).toString();
+
+        @Override
+        public boolean isAllowed(String groovyResourceUrl) {
+            return groovyResourceUrl.equals(scriptUrl);
+        }
     }
 
 }
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java
@@ -30,10 +30,9 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.model.Options;
 import org.jenkinsci.plugins.workflow.cps.CpsScript;
 import org.jenkinsci.plugins.workflow.cps.CpsThread;
 import org.jenkinsci.plugins.workflow.cps.GlobalVariable;
+import org.jenkinsci.plugins.workflow.cps.GroovySourceFileAllowlist;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
-import java.util.Objects;
-import org.jenkinsci.plugins.workflow.cps.GroovySourceFileAllowlist;
 
 /**
  * Loads the main "pipeline" step as well as the additional CPS-transformed code it depends on.
@@ -76,7 +75,7 @@ public class ModelStepLoader extends GlobalVariable {
 
     @Extension
     public static class ModelInterpreterAllowlist extends GroovySourceFileAllowlist {
-        private final String scriptUrl = Objects.requireNonNull(getClass().getClassLoader().getResource("org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy")).toString();
+        private final String scriptUrl = ModelStepLoader.class.getResource("/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy").toString();
 
         @Override
         public boolean isAllowed(String groovyResourceUrl) {
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java
@@ -75,7 +75,7 @@ public class ModelStepLoader extends GlobalVariable {
 
     @Extension
     public static class ModelInterpreterAllowlist extends GroovySourceFileAllowlist {
-        private final String scriptUrl = ModelStepLoader.class.getResource("/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy").toString();
+        private final String scriptUrl = ModelStepLoader.class.getResource("ModelInterpreter.groovy").toString();
 
         @Override
         public boolean isAllowed(String groovyResourceUrl) {
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeLogConditional.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeLogConditional.java
@@ -38,7 +38,9 @@ import org.kohsuke.stapler.DataBoundConstructor;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Objects;
 import java.util.regex.Pattern;
+import org.jenkinsci.plugins.workflow.cps.GroovySourceFileAllowlist;
 
 /**
  * Conditional that checks the messages in the changelog.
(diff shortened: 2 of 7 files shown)


pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescribable.java
```diff
@@ -55,20 +52,7 @@ public WithScriptScript getScript(CpsScript cpsScript) throws Exception {
...
-        } catch (ClassNotFoundException e) {
-            // This is special casing to deal with PluginFirstClassLoaders, which don't have a functional findResource method.
-            // That results in GroovyClassLoader.loadClass failing to find resources to parse and load.
-            // TODO delete JENKINS-44898 workaround as of Jenkins 2.66
-            URL res = getDescriptor().getClass().getClassLoader().getResource(getDescriptor().getScriptClass().replace('.', '/') + ".groovy");
-            if (res != null) {
-                clz = ((GroovyClassLoader) cpsScript.getClass().getClassLoader()).parseClass(new GroovyCodeSource(res));
-            } else {
-                throw e;
-            }
```
dwnusbaum, line 70: Strictly speaking, we don't have to remove this fallback. However, it is obsolete as far as I can tell, and the fallback interacts weirdly with SECURITY-359, since it causes any files that are rejected by the allowlist to be loaded using a regular Java class loader and then parsed it using the _untrusted_ `GroovyClassLoader` for the Pipeline, causing the Groovy source file to be sandbox-transformed, which leads to unexpected `RejectedAccessException`s and confusing error messages.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescriptor.java
```diff
@@ -83,4 +101,18 @@ public T newInstance(Map<String,Object> arguments) throws Exception {
     public T newInstance() throws Exception {
         return clazz.newInstance();
     }
+
+    @Extension
+    public static class WithScriptAllowlist extends GroovySourceFileAllowlist {
```
dwnusbaum, line 106: This implementation of `GroovySourceFileAllowlist` is used for implementations of `DeclarativeAgent` and `DeclarativeStageConditional` (aka `when` conditions). `registerScript` is called by the `WithScriptDescriptor` constructor, so it automatically handles most implementations (barring a few special cases like abstract Groovy scripts), even those in other plugins like `docker-workflow` and `kubernetes`.

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgent.java
```diff
@@ -104,4 +107,14 @@ public DeclarativeAgentDescriptor getDescriptor() {
         return (DeclarativeAgentDescriptor) super.getDescriptor();
     }
 
+    @Extension
+    public static class CheckoutScriptAllowlist extends GroovySourceFileAllowlist {
```
dwnusbaum, line 111: We need this because `DeclarativeAgent` itself is abstract. Another option could be to hard-code this file in `WithScriptDescriptor`.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java
```diff
@@ -72,4 +74,14 @@ public static void invalidateOptionTypeCaches() {
         Options.invalidateCaches();
     }
 
+    @Extension
+    public static class ModelInterpreterAllowlist extends GroovySourceFileAllowlist {
+        private final String scriptUrl = Objects.requireNonNull(getClass().getClassLoader().getResource("org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy")).toString();
```
jglick, line 79:
```suggestion
        private final String scriptUrl = ModelInterpreterAllowlist.class.getResource("/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy").toString();
```

and fix imports and similarly for others:

* do not risk using the wrong `ClassLoader` if subclassed
* no need for `getClassLoader` call if you give an absolute path
* `requireNonNull` redundant since `toString` would fail anyway if the resource were missing

dwnusbaum:
Rebuilding

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescriptor.java
```diff
@@ -39,6 +46,10 @@
  */
 public abstract class WithScriptDescriptor<T extends WithScriptDescribable<T>> extends Descriptor<T> {
 
+    protected WithScriptDescriptor() {
+        ExtensionList.lookupSingleton(WithScriptAllowlist.class).registerScript(getScriptResource());
```
jglick, line 50: SpotBugs might complain that you are calling an overridable method from a constructor?

pom.xml
```diff
@@ -94,6 +94,23 @@
...
+      <dependency>
+        <!-- TODO: Remove once this version is included in BOM -->
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>script-security</artifactId>
+        <version>1172.v35f6a_0b_8207e</version>
+      </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins.workflow</groupId>
+        <artifactId>workflow-cps</artifactId>
+        <version>2699.v7231c3b_4a_231</version> <!-- TODO: https://github.com/jenkinsci/workflow-cps-plugin/pull/538 -->
```
jglick, line 106: This override is just here temporarily right?

jglick (approved):

pom.xml
dwnusbaum: I have it set to this version right now to test that everything works once the relevant entries are removed from `default-allowlist` upstream. Once this PR has been reviewed, I will change it to `2692.v76b_089ccd026` (first version with the new API) so it can be merged and released, unless that version of `workflow-cps` has made it into a BOM release in the meantime, in which case I will update the BOM and delete this.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ModelStepLoader.java
dwnusbaum: All good points, I made the corresponding changes in [e1feeac](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/529/commits/e1feeac84e37eaa879c04da5e21363f509ba707d).

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescriptor.java
dwnusbaum: Yeah, I was surprised that it does not complain about this. Maybe it is because `getScriptResource` is `protected`, so SpotBugs assumes the behavior is intentional.

dwnusbaum:
I am requesting reviews but leaving this in draft state so that it is not accidentally merged with the incremental dependencies.

car-roll (approved):

jglick:
Is there something holding up release of this? I was just working on something in #533 that would have added another file next to `CheckoutScript.groovy` and found that it is not yet possible.

pom.xml
dwnusbaum: No BOM release yet, so I downgraded `workflow-cps` in [36347d3](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/529/commits/36347d379ba56dc60bacb17e7c3f9f6ab4eadfbc).

jglick:
No rush, I can work around it for now.

yaroslavafenkin (approved):

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/ChangeLogConditional.java
```diff
@@ -90,4 +91,18 @@ public Expression transformToRuntimeAST(@CheckForNull ModelASTWhenContent origin
...
         return "(?m)(?s)^[^\\r\\n]*?" + pattern + "[^\\r\\n]*?$";
     }
+
+    /**
+     * AbstractChangelogConditionalScriptAllowlist.groovy is a superclass of the Groovy scripts for some subclasses of
+     * {@link DeclarativeStageConditional}, but does not have any direct equivalent Java class, so we just allow it here.
+     */
+    @Extension
+    public static class ChangelogConditionalScriptAllowlist extends GroovySourceFileAllowlist {
+        private final String scriptUrl = ChangeLogConditional.class.getResource("/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/AbstractChangelogConditionalScript.groovy").toString();
```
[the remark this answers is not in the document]
dwnusbaum: Ah, yeah this did not work with `ClassLoader.getResource` but now it is ok. Done in [778707d](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/529/commits/778707d02aaac703b74cca47751fd5976dae44f3).

jglick (approved):

jglick:
Or just enable automerge, if you have admin rights to the repo.

PR #532, opened 2022-05-25
Simplify `AnyScript`

jglick:
Simplifying a rather roundabout system to just be like https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/12b420f036e5e4d114791602b896c5790567f7a6/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/LabelScript.groovy#L41-L45 except without the `label` and `customWorkspace` fields. The weird logic seems to have been introduced in #94, at which time perhaps it made sense but it seems `CheckoutScript` when introduced in #123 was not also used here.

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/AnyScript.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/AnyScript.groovy
@@ -25,6 +25,7 @@
 
 package org.jenkinsci.plugins.pipeline.modeldefinition.agent.impl
 
+import org.jenkinsci.plugins.pipeline.modeldefinition.agent.CheckoutScript
 import org.jenkinsci.plugins.pipeline.modeldefinition.agent.DeclarativeAgentScript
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/agent/impl/AnyScript.groovy
```diff
@@ -36,11 +37,10 @@ class AnyScript extends DeclarativeAgentScript<Any> {
 
     @Override
     Closure run(Closure body) {
-        Label l = (Label) Label.DescriptorImpl.instanceForName("label", [label: null])
-        l.copyFlags(describable)
```
jglick, line 40: Note two methods which are no longer used in non-test code in this plugin but still used in https://github.com/jenkinsci/docker-workflow-plugin/blob/c7dedefcf68f6ce1320c8aaaeb831628ea21a086/src/main/java/org/jenkinsci/plugins/docker/workflow/declarative/DeclarativeDockerUtils.java#L91-L92

jtnord (approved):
as far as I understand this code, I see nothing obviously wrong

PR #544, opened 2022-06-15
Update BOM and remove unnecessary version specifiers

dwnusbaum:
Takes care of #539, #538, #537, and #536 by updating the BOM and then removing version specifiers and `dependencyManagement` entries that are no longer needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
         <artifactId>bom-2.332.x</artifactId>
-        <version>1289.v5c4b_1c43511b_</version>
+        <version>1409.v7659b_c072f18</version>
         <scope>import</scope>
         <type>pom</type>
       </dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -146,8 +129,6 @@
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/pipeline-model-definition-plugin</gitHubRepo>
     <jenkins.version>2.332.1</jenkins.version>
-    <groovy.version>2.4.12</groovy.version>
```
dwnusbaum, line 149: None of the modules here depend on `groovy` directly, so I'm not entirely sure why this was here to start with. Either way we want to pick up the version defined in the core BOM, which is what will be used at runtime.

basil (approved):

car-roll (approved):

dwnusbaum:
(I'm marking this as `enhancement` to get an automated release of #529 without having to watch the `master` build, but I'll fix up the release notes afterwards.)

PR #596, opened 2023-03-16
[JENKINS-70808] Add a specific isRestartable method that accept a sta…

Dohbedoh:
* JENKINS issue(s):
    * [JENKINS-70808](https://issues.jenkins.io/browse/JENKINS-70808)
* Description:
    * Add a method to specifically check if a stage is restartable that will return as soon as one is found and will not check the status of other irrelevant stages. To be consumed downstream by blueocean-pipeline-api-impl at https://github.com/jenkinsci/blueocean-plugin/blob/49440310221d3e68040c0a46842b8a91d0a577be/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java#L189-L193.
* Users/aliases to notify:
    * @olamy

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
@@ -38,6 +38,7 @@ import org.acegisecurity.AccessDeniedException;
 import org.jenkinsci.plugins.pipeline.StageStatus;
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils;
 import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStage;
+import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStages;
 import org.jenkinsci.plugins.pipeline.modeldefinition.causes.RestartDeclarativePipelineCause;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -170,6 +171,30 @@ public List<String> getRestartableStages() {
...
+        FlowExecution execution = getExecution();
+        if (execution != null) {
+            ExecutionModelAction execAction = run.getAction(ExecutionModelAction.class);
+            if (execAction != null) {
+                ModelASTStages stages = execAction.getStages();
+                if (stages != null) {
+                    for (ModelASTStage s : stages.getStages()) {
+                        if(s.getName().equals(stageName)) {
+                            return !Utils.stageHasStatusOf(s.getName(), execution,
+                                StageStatus.getSkippedForFailure(), StageStatus.getSkippedForUnstable());
```
jglick, line 189: So a failed/unstable stage is _not_ restartable? The logic here is unclear. Is this just a copy / adaptation of code elsewhere?

olamy:
might need to take into account https://github.com/jenkinsci/pipeline-model-definition-plugin/commit/821aabd0b6069f1d25b6b171a61f9ee40a9577aa

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
Dohbedoh: @jglick it is the same logic as [isRestartable()](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/a3f80da24c5bc4ad366c60b86165b845df4c2327/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java#L154-L172) but specific to a single stage.

PR #598, opened 2023-03-24
Filter nodes with propagate: false when evaluating WarningActions

stuartrowe:
* JENKINS issue(s):
    * https://issues.jenkins.io/browse/JENKINS-70844
* Description:
    * https://github.com/jenkinsci/pipeline-build-step-plugin/pull/107 added a `WarningAction` to `build` nodes to allow pipeline visualizers (blueocean, pipeline graph view, etc.) to show the result of downstream builds even if ignored by `propagate: false`. See https://issues.jenkins.io/browse/JENKINS-70623.
    * As a knock-on this broke `post` condition sections for declarative pipelines which are using the presence of `WarningAction`s in order to determine which post conditions to evaluate.
    * This fix adds a filter to check if a node has an entry for `propagete: false` in its [`ArgumentsAction`](https://javadoc.jenkins.io/plugin/workflow-api/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.html).
    * This fix is an alternative to https://github.com/jenkinsci/pipeline-build-step-plugin/pull/110 which skips adding `WanringAction`s when `propagate: false`.
* Documentation changes:
* Users/aliases to notify:
    * @car-roll @olamy

### DIFF
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/BuildCondition.java
@@ -30,6 +30,7 @@ import hudson.Util;
 import hudson.model.Result;
 import org.jenkinsci.plugins.pipeline.modeldefinition.CommonUtils;
 import org.jenkinsci.plugins.structs.SymbolLookup;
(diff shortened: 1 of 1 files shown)


stuartrowe:
PR job timed out / aborted on the `linux-17` branch. Closing and re-opening to trigger another build.

rsandell (commented):
Thanks!
A test or two would be nice.

PR #660, opened 2023-08-24
Allow valid agent types in PCT context

basil:
While trying to add the Kubernetes plugins to `jenkinsci/bom` in https://github.com/jenkinsci/bom/pull/2417 I got some test failures of the form:

> "error": "Invalid agent type \"foo\" specified. Must be one of [any, kubernetes, label, none, otherField]"
}]) didn't contain expected error 'Invalid agent type "foo" specified. Must be one of [any, label, none, otherField]'

The existing test logic had tried to deal with this case by defining a `legalAgentTypes` field, but it was flawed in a few ways:

- It was not used consistently in `AbstractModelDefTest#configsWithErrors`: `unknownBareAgentType` used it but not `unknownAgentType`
- Any usage of `legalAgentTypes` from `AbstractModelDefTest#configsWithErrors` was flawed and yielding the empty string (thus making the test pass erroneously) because Jenkins had not been started when populating the list during JUnit parameterized test initialization
- It did not assert that the types expected to be contributed by this plugin itself were present

This PR fixes all three mistakes by:

- Consistently _not_ using `legalAgentTypes` in `AbstractModelDefTest#configsWithErrors`, since that is too early for it to be used
- Making the field non-static so that it cannot be used too early by accident in the future
- Ensuring that any tests that need `legalAgentTypes` use it in the test itself, after Jenkins startup (currently this list includes `ErrorsEndpointOpsTest`, `ErrorsJSONParserTest`, and `ValidatorTest`)
- Asserting that the types expected to be contributed by this plugin are present

### Testing done

To test this PR I reproduced the original problem by adding the Kubernetes plugins to the test classpath and verified the BOM error occurred. Next I applied the changes from this PR and verified the problem was fixed. Then I removed the Kubernetes plugins from the test classpath and verified the tests still passed in the non-PCT scenario.

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
@@ -82,19 +82,22 @@ public abstract class AbstractModelDefTest extends AbstractDeclarativeTest {
     @Rule public GitSampleRepoRule otherRepo = new GitSampleRepoRule();
     @Rule public GitSampleRepoRule thirdRepo = new GitSampleRepoRule();
 
-    protected static String legalAgentTypes = "";
+    protected String legalAgentTypes = "";
 
-    @BeforeClass
-    public static void setUpPreClass() throws Exception {
-        List<String> agentTypes = new ArrayList<>();
+    @Before
+    public void setUpPreClass() {
+        // Ensure that the agent types expected to be contributed by this plugin are present
+        NavigableSet<String> agentTypes = new TreeSet<>(List.of("any", "label", "none", "otherField"));
 
+        // Allow agent types to be contributed by other plugins; for example, the Kubernetes plugin PCT context
         for (DeclarativeAgentDescriptor d : j.jenkins.getExtensionList(DeclarativeAgentDescriptor.class)) {
             String symbol = symbolFromDescriptor(d);
             if (symbol != null) {
                 agentTypes.add(symbol);
             }
         }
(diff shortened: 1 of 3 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java
```diff
@@ -223,7 +226,7 @@ public static Iterable<Object[]> configsWithErrors() {
         result.add(new Object[]{"invalidWrapperType", "Invalid option type \"echo\". Valid option types:"});
         result.add(new Object[]{"invalidStageWrapperType", "Invalid option type \"echo\". Valid option types:"});
 
-        result.add(new Object[]{"unknownBareAgentType", Messages.ModelValidatorImpl_InvalidAgentType("foo", legalAgentTypes)});
+        result.add(new Object[]{"unknownBareAgentType", Messages.ModelValidatorImpl_InvalidAgentType("foo", "[any, label, none, otherField]")});
```
basil, line 229: Newly consistent with `unknownAgentType` in the sense that both now hardcode a list of agent types rather than using `legalAgentTypes`, which is not yet initialized when this code is running.

basil:
Currently blocking https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/660

basil:
Currently blocking https://github.com/jenkinsci/bom/pull/2417

basil:
Currently blocking https://github.com/jenkinsci/bom/issues/2453

basil:
Currently blocking https://github.com/jenkinsci/bom/issues/2453

jglick (approved):

PR #675, opened 2023-09-30
Test with Java 21

MarkEWaite:
## Test with Java 21

Java 21 released Sep 19, 2023. We'd like to announce full support for Java 21 in early October and would like the most used plugins to be compiling and testing with Java 21.  The acceptance test harness and plugin bill of materials tests are already passing with Java 21.  This is a further step to improve plugin readiness for use with Java 21 and for development with Java 21.

This pull request includes and supersedes:

* #668

This pull request and #668 supersede the following pull requests:

* #674
* #670
* #669
* #664
* #636

* JENKINS issue(s):
    * n/a
* Description:
    * See the text at the top of the pull request
* Documentation changes:
    * No documentation changes required because all changes are internal to the plugin
* Users/aliases to notify:
    * @jglick
    * @rsandell

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,12 @@
-buildPlugin(timeout: 120, useContainerAgent: true, configurations: [
-    [platform: 'linux', jdk: 17],
-    [platform: 'windows', jdk: 11],
+/*
+ See the documentation for more options:
+ https://github.com/jenkins-infra/pipeline-library/
+*/
+buildPlugin(
+  timeout: 120, // reduce test failures due to timeout
+  useContainerAgent: true, // Set to `false` if you need to use Docker for containerized tests
+  configurations: [
+    [platform: 'linux', jdk: 21],
+    [platform: 'windows', jdk: 17],
+    [platform: 'linux', jdk: 11],
 ])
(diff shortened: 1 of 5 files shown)


MarkEWaite:
I used my ci.jenkins.io administrator permissions to run the build with the new Jenkinsfile.

Jenkinsfile
```diff
@@ -1,4 +1,12 @@
...
+/*
+ See the documentation for more options:
+ https://github.com/jenkins-infra/pipeline-library/
+*/
+buildPlugin(
+  timeout: 120, // reduce test failures due to timeout
+  useContainerAgent: true, // Set to `false` if you need to use Docker for containerized tests
+  configurations: [
+    [platform: 'linux', jdk: 21],
+    [platform: 'windows', jdk: 17],
```
rsandell, line 10: I see this is what it looked like before, bu it doesn't really seem good to me to run with a different jdk version on windows then any of the Linux ones. If a test fails on the windows branch it won't be obvious if it is a Win or jdk problem.

MarkEWaite: Thanks for the review comment!

This is a consistent pattern that is used in many Jenkins plugins and has been used across those plugins for an extended period.  I've not heard of a case where we failed to detect an issue by this intentional choice to reduce cost and still retain coverage of two Java versions and two platforms.

The prior configuration was a calculated risk that intentionally reduced the cost of execution (especially on Windows where the tests require over 60 minutes to complete) and accepted there would be greater difficulty in diagnosis of platform specific failure.  Since declarative Pipeline is not strongly platform dependent, that seems like a reasonable compromise that was chosen before and continues to be chosen in this pull request.

jglick (approved):

PR #683, opened 2023-12-08
Remove dead test code

MarkEWaite:
## Remove dead test code

* Description:

The VersionNumber() of the running Jenkins controller used for the tests will never be less than the Jenkins minimum version defined in the plugin pom.  Remove the conditional so that it is not a distraction in the test suite source code.


basil (commented):
Duplicate of #672?

PR #684, opened 2023-12-09
Use joda-time api plugin

jonesbusy:
Use joda-time-api plugin

- [x] Remove version when bom released https://github.com/jenkinsci/bom/pull/2732

### DIFF
--- a/pipeline-model-api/pom.xml
+++ b/pipeline-model-api/pom.xml
@@ -53,7 +53,16 @@
           <groupId>com.google.guava</groupId>
           <artifactId>guava</artifactId>
         </exclusion>
+        <exclusion>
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

jonesbusy:
Fixed conflicts

NotMyFault (approved):
Thanks!

olamy (approved):

balakine:
I understand why it's excluded, but why do we need this dependency? I don't see a single import of org.joda.

basil:
> I understand why it's excluded, but why do we need this dependency? I don't see a single import of org.joda.

We consume it transitively via `json-schema-validator`.

PR #686, opened 2024-01-09
Make WithScriptAllowlist look up instances of WithScriptDescriptor dynamically instead of registering allowed scripts directly

dwnusbaum:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/538#issuecomment-1883042255 and https://github.com/jenkinsci/workflow-cps-plugin/pull/538#issuecomment-1883348978.

I have not ben able to reproduce the issue yet, so this is just a speculative change.

### DIFF
--- a/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescriptor.java
+++ b/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescriptor.java
@@ -29,15 +29,16 @@ import org.jenkinsci.plugins.structs.SymbolLookup;
 import org.jenkinsci.plugins.structs.describable.DescribableModel;
 import org.jenkinsci.plugins.workflow.cps.GroovySourceFileAllowlist;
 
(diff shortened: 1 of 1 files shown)


pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/withscript/WithScriptDescriptor.java
```diff
@@ -104,15 +101,17 @@ public T newInstance() throws Exception {
...
-
-        private void registerScript(URL scriptUrl) {
-            scriptUrls.add(scriptUrl.toString());
-        }
-
         @Override
         public boolean isAllowed(String groovyResourceUrl) {
-            return scriptUrls.contains(groovyResourceUrl);
+            for (WithScriptDescriptor descriptor : ExtensionList.lookup(WithScriptDescriptor.class)) {
+                URL scriptResource = descriptor.getScriptResource();
```
dwnusbaum, line 107: This will be slower than before. I suspect it would be ok as-is, but https://github.com/jenkinsci/workflow-cps-plugin/pull/829 will cache all of the allowlist responses, which should make performance a non-issue.

jglick (approved):

PR #700, opened 2024-03-11
[JENKINS-72841] - Pass flow node ID when skipping stage

timja:
* JENKINS issue(s):
    * [JENKINS-72841](https://issues.jenkins.io/browse/JENKINS-72841)
    * Fixes https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/220
    * Fixes https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/698
* Description:
    * Previously skipping stages would only be tagged for the first occurrence of the stage. This now passes the flow node id rather than guessing by stage name
* Documentation changes:
* Users/aliases to notify:
    * ...

Before:
image

After:

image

see https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/220

### DIFF
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -314,11 +314,29 @@ class Utils {
 
         def stepContextFlowNode = execution.getNode(stepContextFlowNodeId)
 
-        def stageFlowNode = stepContextFlowNode.iterateEnclosingBlocks().find { blockStartNode ->
+        List<FlowNode> nodes = []
+
+        FlowNode stage = (FlowNode) stepContextFlowNode.iterateEnclosingBlocks().find { blockStartNode ->
             CommonUtils.isStageWithOptionalName(stageName).test(blockStartNode)
         }
 
-        addTagToFlowNode(stageFlowNode, tagName, tagValue)
+        if (stage != null) {
+            nodes.add(stage)
+
+            // Additional check needed to get the possible enclosing parallel branch for a nested stage.
+            Filterator<FlowNode> filtered = FlowScanningUtils.fetchEnclosingBlocks(stage)
+                    .filter(isParallelBranchFlowNode(stageName))
+
+            filtered.each { f ->
+                if (f != null) {
+                    nodes.add(f)
+                }
+            }
+        }
+
+        nodes.each { currentNode ->
+            addTagToFlowNode(currentNode, tagName, tagValue)
+        }
 
     }
 
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -314,30 +314,16 @@ class Utils {
 
         def stepContextFlowNode = execution.getNode(stepContextFlowNodeId)
 
-        List<FlowNode> nodes = []
-
-        FlowNode stage = (FlowNode) stepContextFlowNode.iterateEnclosingBlocks().find { blockStartNode ->
-            CommonUtils.isStageWithOptionalName(stageName).test(blockStartNode)
-        }
-
-        if (stage != null) {
-            nodes.add(stage)
-
-            // Additional check needed to get the possible enclosing parallel branch for a nested stage.
-            Filterator<FlowNode> filtered = FlowScanningUtils.fetchEnclosingBlocks(stage)
-                    .filter(isParallelBranchFlowNode(stageName))
-
-            filtered.each { f ->
-                if (f != null) {
-                    nodes.add(f)
-                }
+        int count = 0
+        for (FlowNode node : stepContextFlowNode.iterateEnclosingBlocks()) {
+            if (CommonUtils.isStageWithOptionalName(stageName).test(node) || isParallelBranchFlowNode(stageName).apply(node)) {
+                addTagToFlowNode(node, tagName, tagValue)
+                count++
+            }
+            if (count == 2) {
+                break
             }
         }
-
-        nodes.each { currentNode ->
-            addTagToFlowNode(currentNode, tagName, tagValue)
-        }
-
     }
 
     private static void addTagToFlowNode(FlowNode currentNode, String tagName, String tagValue) {
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -312,17 +312,24 @@ class Utils {
         CpsThread thread = CpsThread.current()
         FlowExecution execution = thread.execution
 
-        def stepContextFlowNode = execution.getNode(stepContextFlowNodeId)
+        try {
+            def stepContextFlowNode = execution.getNode(stepContextFlowNodeId)
+            if (stepContextFlowNode == null) {
+                return;
+            }
 
-        int count = 0
-        for (FlowNode node : stepContextFlowNode.iterateEnclosingBlocks()) {
-            if (CommonUtils.isStageWithOptionalName(stageName).test(node) || isParallelBranchFlowNode(stageName).apply(node)) {
-                addTagToFlowNode(node, tagName, tagValue)
-                count++
+            int count = 0
+            for (FlowNode node : stepContextFlowNode.iterateEnclosingBlocks()) {
+                if (CommonUtils.isStageWithOptionalName(stageName).test(node) || isParallelBranchFlowNode(stageName).apply(node)) {
+                    addTagToFlowNode(node, tagName, tagValue)
+                    count++
+                }
+                if (count == 2) {
+                    break
+                }
             }
-            if (count == 2) {
-                break
-            }
+        } catch (IOException e) {
+            // Ignore.
         }
     }
 
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
@@ -308,21 +308,39 @@ class Utils {
         }
     }
 
+    static void markStageWithTag(String stageName, String stepContextFlowNodeId, String tagName, String tagValue) {
+        CpsThread thread = CpsThread.current()
+        FlowExecution execution = thread.execution
+
+        def stepContextFlowNode = execution.getNode(stepContextFlowNodeId)
+
+        def stageFlowNode = stepContextFlowNode.iterateEnclosingBlocks().find { blockStartNode ->
+            CommonUtils.isStageWithOptionalName(stageName).test(blockStartNode)
+        }
+
+        addTagToFlowNode(stageFlowNode, tagName, tagValue)
+
+    }
+
+    private static void addTagToFlowNode(FlowNode currentNode, String tagName, String tagValue) {
+        if (currentNode != null) {
+            TagsAction tagsAction = currentNode.getAction(TagsAction.class)
+            if (tagsAction == null) {
+                tagsAction = new TagsAction()
+                tagsAction.addTag(tagName, tagValue)
+                currentNode.addAction(tagsAction)
+            } else if (tagsAction.getTagValue(tagName) == null) {
+                tagsAction.addTag(tagName, tagValue)
+                currentNode.save()
+            }
+        }
+    }
+
     static void markStageWithTag(String stageName, String tagName, String tagValue) {
         List<FlowNode> matched = findStageFlowNodes(stageName)
 
         matched.each { currentNode ->
-            if (currentNode != null) {
-                TagsAction tagsAction = currentNode.getAction(TagsAction.class)
-                if (tagsAction == null) {
-                    tagsAction = new TagsAction()
-                    tagsAction.addTag(tagName, tagValue)
-                    currentNode.addAction(tagsAction)
-                } else if (tagsAction.getTagValue(tagName) == null) {
-                    tagsAction.addTag(tagName, tagValue)
-                    currentNode.save()
-                }
-            }
+            addTagToFlowNode(currentNode, tagName, tagValue)
         }
     }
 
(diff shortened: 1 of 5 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -222,6 +223,11 @@ class ModelInterpreter implements Serializable {
 
     }
 
+    @NonCPS
+    private String getFlowNodeId() {
+        return script.getContext(FlowNode.class).getId()
```
timja, line 228: @dwnusbaum this returns the flownode for the org.jenkinsci.plugins.workflow.steps.GetContextStep not the current flow node. 

The unit test passes which tripped me up when opening this as it just counts tags not if they are the exact right ones.

timja: ah I see you meant scanning is still needed:
https://issues.jenkins.io/browse/JENKINS-72841?focusedId=445970&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-445970

dwnusbaum: Yes sorry I should have mentioned that the node will not be exactly the one you want. I think though the immediate parent of this node will be the one you want, or maybe the immediate grandparent? I would some Pipeline with matrix to see. You might not need to scan.

timja: I iterated enclosing blocks and it worked fine.

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -308,21 +308,39 @@ class Utils {
...
+                tagsAction.addTag(tagName, tagValue)
+                currentNode.addAction(tagsAction)
+            } else if (tagsAction.getTagValue(tagName) == null) {
+                tagsAction.addTag(tagName, tagValue)
+                currentNode.save()
+            }
+        }
+    }
+
     static void markStageWithTag(String stageName, String tagName, String tagValue) {
```
timja, line 339: I've left this in place as it can be called from scripted pipelines (and people definitely do it)

pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy
```diff
@@ -308,21 +308,57 @@ class Utils {
...
+            filtered.each { f ->
+                if (f != null) {
+                    nodes.add(f)
+                }
+            }
+        }
+
+        nodes.each { currentNode ->
+            addTagToFlowNode(currentNode, tagName, tagValue)
+        }
```
dwnusbaum, line 339: `FlowScanningUtils.fetchEnclosingBlocks(stage)` should just return a subset of `stepContextFlowNode.iterateEnclosingBlocks()` (and IDK why the method even exists in the first place). I think you should be able to simplify this:

```suggestion
        for (FlowNode node : stepContextFlowNode.iterateEnclosingBlocks()) {
            if (CommonUtils.isStageWithOptionalName(stageName).test(node) || isParallelBranchFlowNode(stageName).apply(node)) {
                addTagToFlowNode(node, tagName, tagValue)
            } 
        }
```

Really though IIUC, the problem has to do with [`CommonUtils.isStageWithOptionalName`](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/b364814683740de3e3af94e521df8d5f9ca8c544/pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/CommonUtils.java#L52) returning true both for the stage step itself as the node for the parallel branch with the same name, so maybe `isParallelBranchFlowNode` is redundant once you switch to iterating over all of the enclosing blocks. Also it might be best to stop iterating once you hit two matches in case the Pipeline is really complicated.

Visualizations are probably only targeting one or the other of those nodes, so I guess you were targeting the parallel branch node rather than the stage, or something like that? I recall having to do something weird in this case too for my visualization, maybe checking both the parallel branch and the stage step for the skipped tag.

timja: yes the parallel branch node is the one that is visualised from what I can tell.

The tests in this plugin expect both the branch and the stage to have the skipped tag on it.

timja: (edge case I hit while testing this which includes some JSON to show what is used if you are interested https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/358)

timja:
@dwnusbaum any chance of another review please?

timja:
@jenkinsci/pipeline-model-definition-plugin-developers could someone take a look please?

dwnusbaum:
@timja Yes sorry I'm just running the OSS BOM and another proprietary PCT suite to see if it's worth trying to adjust `getContext` to only be called when a stage is going to be skipped to avoid having to update a bunch of tests.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -238,15 +244,16 @@ class ModelInterpreter implements Serializable {
                       SkippedStageReason skippedReason) {
         return {
             script.stage(thisStage.name) {
+                def flowNodeId = getFlowNodeId()
```
dwnusbaum, line 247: Still waiting on the full results, but so far this breaks some tests in `blueocean-pipeline-api-impl` and another ~10 spread across some proprietary plugins I have access to:

* `blueocean-pipeline-api-impl / io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.sequentialParallelStagesWithPost`
* `blueocean-pipeline-api-impl / io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.submitInputPostBlock`
* `blueocean-pipeline-api-impl / io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.declarativeSyntheticSteps`
* `blueocean-pipeline-api-impl / io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.encodedStepDescription`
 * `blueocean-pipeline-api-impl / io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.submitInputPostBlockWithParallelStages`
* `blueocean-pipeline-api-impl / io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.declarativeSyntheticSkippedStage`

Maybe we could move the call to `getFlowNodeId` inside of `skipStage` so that the changes only affects tests that are directly related to skipped stages?

dwnusbaum:
I pushed https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/700/commits/bd144259a8a92027109360538429b4f9e7813e51 to try to reduce the number of impacted tests and will rerun the PCT once a new incremental build is available. If nothing else it at least prevents us from having to update `ParallelTest` here.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
dwnusbaum: [bd14425](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/700/commits/bd144259a8a92027109360538429b4f9e7813e51) will hopefully help with this, but tests like `PipelineNodeTest.declarativeSyntheticSkippedStag` will probably still need to be adjusted.

dwnusbaum:
I triggered another PCT run (Just a CloudBees internal one this time) and will check the results tomorrow.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
timja: created a test run in https://github.com/jenkinsci/blueocean-plugin/pull/2555

I tried to run locally but I can't get the plugin to build 😢

dwnusbaum: > I tried to run locally but I can't get the plugin to build due to a like 6 year old npm / node version and unmaintained dependencies 😢

Yep, I ran into the same issue in https://github.com/jenkinsci/blueocean-plugin/pull/2540.


With the latest changes, the total affected tests went from 15 to 6, so that's nice. The only remaining affected test in Blue Ocean appears to be `blueocean-pipeline-api-impl / io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.submitInputPostBlockWithParallelStages`. Here's the output: 

```
org.junit.ComparisonFailure: expected:<[47]> but was:<[50]>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.submitInputPostBlockWithParallelStages(PipelineNodeTest.java:2314)
```

Let's try to update your Blue Ocean PR to fix the test. There are other tests that will need to be updated internally in CloudBees, but I will take care of that. I'll work on all of this later today.

timja: Thanks, I've pushed a guess in https://github.com/timja/blueocean-plugin/commit/5184941cb10b8ae8b6ed0abd2d95a728c5f81bf7

dwnusbaum (approved):
It's a bit awkward to use `getContext` like this, and it's annoying that it affects some tests, but even so, this seems preferable to the kind of approach that was used before and in #698. I've prepped some CloudBees-internal PRs to fix affected tests and will keep an eye on the build of https://github.com/jenkinsci/blueocean-plugin/pull/2555. Once it looks like we've got things working there I'll go ahead and merge this to get a release out.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
jglick, line 228: Can we not use the `withContext` step? It is intended for use in (really) power user trusted libraries, and it is just going to add overhead to the flow node graph for every Declarative build (as well as breaking tons of tests).

Would be better to add some helper method to `Utils` that is run silently, without recording a flow node.

dwnusbaum: > Can we not use the withContext step?

`getContext`

> It is intended for use in (really) power user trusted libraries

What is `ModelInterpreter.groovy` if not an incredibly complicated trusted library written by power users?

> and it is just going to add overhead to the flow node graph for every Declarative build

One extra `FlowNode` per skipped stage. I do not think this is a big problem.

> as well as breaking tons of tests

It breaks 6 known tests across OSS plugins and CloudBees-internal plugins, and PRs have already been filed to fix all of them.

> Would be better to add some helper method to Utils that is run silently, without recording a flow node.

Yes, but here is the problem: Declarative has only a stage name, and wants to add/update tags on the corresponding `FlowNode` objects. Without an initial `FlowNode` to guide the scan, we have to scan the whole `FlowGraph`, and I am suspicious that there are some unfixable bugs with that approach given duplicate stage names in complex Pipelines. #698 shows that kind of approach. I think it is much clearer to just get a contextual ID to guide the search and the tradeoffs of having to use `getContext` do not seem that severe.

Perhaps the code could be adjusted to check the current heads and guess which one is the right one, but I think that is still problematic. For example if you use `matrix` and have two identically named stages/parallel branches, but then only one of them gets skipped, it does not seem trivial to figure out which is the "right" one.

dwnusbaum: > Perhaps the code could be adjusted to check the current heads and guess which one is the right one, but I think that is still problematic. For example if you use matrix and have two identically named stages/parallel branches, but then only one of them gets skipped, it does not seem trivial to figure out which is the "right" one.

Maybe `CpsThread.head` could be used to avoid this problem, but I am not sure, particularly given its Javadoc.

jglick: > Without an initial `FlowNode` to guide the scan

I see, this is the crux of the problem: there actually is no obvious Java/Groovy to call that would return usable information at this point. `CpsThread.current().etc()` would be tricky.

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
basil: > > I tried to run locally but I can't get the plugin to build due to a like 6 year old npm / node version and unmaintained dependencies 😢
> 
> Yep, I ran into the same issue in [jenkinsci/blueocean-plugin#2540](https://github.com/jenkinsci/blueocean-plugin/pull/2540).

Based on a suggestion by @batmat, I tried building this plugin in the environment created by https://github.com/jenkinsci/blueocean-plugin/blob/1ce0c18265882733f3251dcad0b4af616da5a7c9/Dockerfile.build and it built just fine on my AMD64 system. So while definitely annoying, building this plugin locally is not insurmountable on AMD64. Presumably there is some way to run AMD64 code on ARM64 via emulation/virtualization, where the same `Dockerfile.build` could be used.

timja: It doesn’t even work on amd64 macOS for me. Sure you can make it work on docker.

I’ve got a branch locally that has made a fair bit of progress but now need to update the JS builder and figure out who has access to release there

timja: I got it working in https://github.com/jenkinsci/blueocean-plugin/pull/2558

PR #723, opened 2024-06-21
do not rely on long deprecated FileOnMasterKeyStoreSource

jtnord:
`FileOnMasterKeyStoreSource` was deprecated 5 years ago, this replaces it with a non deprecated way of defining a credential.

Adapt to https://github.com/jenkinsci/credentials-plugin/pull/540 by not using code about to be removed

testing done: relying on the test to pass which is testing the test itself.


pipeline-model-definition/src/test/resources/org/jenkinsci/plugins/pipeline/modeldefinition/steps/dummy.p12
jtnord, line 1: cut and paste of https://github.com/jenkinsci/credentials-plugin/blob/5e08c0a77a2f2105824fab15a7067346bd5c47a7/src/test/resources/com/cloudbees/plugins/credentials/impl/test.p12

dwnusbaum (approved):

PR #730, opened 2024-07-08
use valid password for test code

jtnord:
amends #723 that was working as the KeyStore was not validated.

However https://github.com/jenkinsci/credentials-plugin/pull/543 now validates the keystore and hence this test started failing as the password did not match

* JENKINS issue(s):
    * Issue link or n/a
* Description:
    * ...
* Documentation changes:
* Users/aliases to notify:
    * ...

### DIFF
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/CredentialWrapperStepTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/CredentialWrapperStepTest.java
@@ -120,7 +120,7 @@ public class CredentialWrapperStepTest extends AbstractModelDefTest {
         folderStore.addCredentials(Domain.global(), c);
 
         File crt = new File(CredentialWrapperStepTest.class.getResource("dummy.p12").toURI());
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/CredentialWrapperStepTest.java
```diff
@@ -120,7 +120,7 @@ public static void setup() throws Exception {
         folderStore.addCredentials(Domain.global(), c);
 
         File crt = new File(CredentialWrapperStepTest.class.getResource("dummy.p12").toURI());
-        CertificateCredentialsImpl certCred = new CertificateCredentialsImpl(CredentialsScope.GLOBAL, "certCred1", "sample", "test", 
+        CertificateCredentialsImpl certCred = new CertificateCredentialsImpl(CredentialsScope.GLOBAL, "certCred1", "sample", "password",
```
jtnord, line 123: double checked this was the correct password with `openssl pkcs12 -legacy -info -in ./src/test/resources/org/jenkinsci/plugins/pipeline/modeldefinition/steps/dummy.p12`

PR #738, opened 2024-09-05
[JENKINS-73726] Reproduced error in test

jglick:
https://issues.jenkins.io/browse/JENKINS-73726 presumably caused by https://github.com/jenkinsci/workflow-cps-plugin/pull/925


jglick:
https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/6ae91663b44eca3a883608774815da223e2d39bd/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy#L219 is somehow critical. Have not yet figured out how to reproduce without Declarative.

dwnusbaum (approved):

PR #752, opened 2025-01-10
Stop testing Java 11

strangelookingnerd:
## Stop testing Java 11

Jenkins stopped supporting Java 11 with the release of Jenkins 2.463 (weekly) and Jenkins 2.479.1 (LTS). Most plugins stopped spending `ci.jenkins.io` time to run tests that are specific to Java 11.

Let's save money and time by removing the Java 11 test configuration. It has not found any issues that are not also found with Java 17 and Java 21.

Jenkins plugin BOM still tests with Java 11 on older lines (currently 2.452.x and 2.462.x) and the plugin build will continue to generate Java 11 byte code until the parent pom is upgraded to 5.x and the minimum Jenkins version is upgraded to 2.479.1.

Switches to test Java 17 / Java 21 since to match the test configuration used in the plugin archetype.

### Testing done

None. Rely on `ci.jenkins.io` to test it.

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,6 +4,7 @@
 */
 buildPlugin(
   timeout: 120, // reduce test failures due to timeout
+  forkCount: '1C', // run this number of tests in parallel for faster feedback.  If the number terminates with a 'C', the value will be multiplied by the number of available CPU cores
   useContainerAgent: true, // Set to `false` if you need to use Docker for containerized tests
   configurations: [
     [platform: 'linux', jdk: 21],


Jenkinsfile
```diff
@@ -4,9 +4,9 @@
 */
 buildPlugin(
   timeout: 120, // reduce test failures due to timeout
+  forkCount: '1C', // run this number of tests in parallel for faster feedback.  If the number terminates with a 'C', the value will be multiplied by the number of available CPU cores
```
basil, line 7: Not sure if this is safe offhand. This plugin has highly memory-intensive tests and I am not sure if our agents can handle them (even more so given that our Windows agents are in a temporary situation right now).

strangelookingnerd: Would there be a way to give it a try though? If it helps reduce the runtime of tests and the agents can handle it, everyone wins.

basil: > Would there be a way to give it a try though?

I think running the tests a few times (a dozen at most) against the final CI infrastructure would be enough to give me confidence that this isn't introducing new flakiness. Right now the Windows CI infrastructure is temporary, not final, so I would want to wait until it is finalized.

strangelookingnerd: Sounds reasonable. Let’s wait for the Windows CI to recover and come back to this 👍🏼

strangelookingnerd: https://github.com/jenkins-infra/helpdesk/issues/4490 has been resolved and windows builds should be back to normal. Should we give this a try?

basil: @strangelookingnerd I ran two builds of this PR today:

https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/PR-752/3/
https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/PR-752/3/

These took about 1 hour and 45 minutes, while the same builds today on the default branch took only an hour:

https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/master/625/
https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/master/626/

I'm not sure why enabling parallelism slowed down the total test runtime, but looks like it did...

strangelookingnerd: It seems to only affect the Windows build. Linux build runtime is unchanged, give or take a couple minutes. But overall it does not seems worth it. Closing this PR.

PR #762, opened 2025-02-21
Reduce depth of `Closure`s being `call`ed

jglick:
Output from https://github.com/jenkinsci/workflow-support-plugin/pull/305 indicate that a lot of stack frames are `Closure.call`. In some cases this is actually helping to structure code. In others, it is not. Two half-cheese, half-sausage pizzas are just one cheese and one sausage pizza.

[CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-54716)

### DIFF
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
@@ -148,7 +148,7 @@ class ModelInterpreter implements Serializable {
      * @param restartedStageName the name of the stage we're restarting at. Null if this is not a restarted build or this is
      *     called from a nested stage.
      * @param skippedReason Possibly null reason the container for the stages was skipped.
-     * @return A closure to execute
+     * @return A closure to execute (TODO just run directly)
      */
     def evaluateSequentialStages(Root root, Stages stages, Throwable firstError, Stage parent, String restartedStageName,
                                  SkippedStageReason skippedReason) {
@@ -238,7 +238,7 @@ class ModelInterpreter implements Serializable {
      * @param firstError An error that's already occurred earlier in the build. Can be null.
      * @param parent The possible parent stage, defaults to null.
      * @param skippedReason Possibly null reason this stage's parent, and therefore itself, is skipped.
-     * @return
+     * @return TODO just run directly
      */
     def evaluateStage(Root root, Agent parentAgent, Stage thisStage, Throwable firstError, Stage parent,
                       SkippedStageReason skippedReason) {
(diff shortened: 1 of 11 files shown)


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/ModelInterpreter.groovy
```diff
@@ -334,36 +334,31 @@ class ModelInterpreter implements Serializable {
...
-                        // Single parameter, it's the submitter.
-                        inputEnv = ["${input.submitterParameter}=${submitted}"]
-                    } else if (input.parameters.size() == 1) {
-                        // One defined parameter, so we know its name.
-                        inputEnv = ["${input.parameters.first().name}=${submitted}"]
-                    }
-                    script.withEnv(inputEnv) {
-                        body.call()
-                    }
+            def submitted = script.input(message: input.message, id: input.id, ok: input.ok, submitter: input.submitter,
```
jglick, line 337: (ignore WS)

pipeline-model-extensions/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/agent/DeclarativeAgentScript.java
```diff
@@ -27,7 +27,7 @@
 import org.jenkinsci.plugins.pipeline.modeldefinition.withscript.WithScriptScript;
 import org.jenkinsci.plugins.workflow.cps.CpsScript;
 
-
+@Deprecated
```
jglick, line 30: Implemented in a few other plugins:
* `kubernetes`
* `docker`
* `docker-workflow`
* `amazon-ecs`

jglick:
I presume a (Linux-only) timeout in `BasicModelDefTest.stages300` is a flake, retrying…

Vlatombe (approved):

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/BasicModelDefTest.java
```diff
@@ -78,7 +78,7 @@ public static void setUpAgent() throws Exception {
 
     @Issue("JENKINS-47363")
     // Give this a longer timeout
-    @Test(timeout=5 * 60 * 1000)
+    @Test(timeout=10 * 60 * 1000)
```
jglick, line 81: And now it is [passing in 98s](https://ci.jenkins.io/job/Plugins/job/pipeline-model-definition-plugin/job/PR-762/9/testReport/org.jenkinsci.plugins.pipeline.modeldefinition/BasicModelDefTest/stages300/). :shrug:

felipecrs:
This is listed as a breaking change in the release notes, but the details are shady.

Can you please clarify whether it's safe to upgrade this plugin for users using docker-workflow or not? And what's the game plan otherwise.

jglick:
@felipecrs if all goes well, https://github.com/jenkinsci/docker-workflow-plugin/pull/337 should release shortly. Sorry for any disruption.

felipecrs:
Got it, that's alright, no need to apologize.

jglick:
Good call; the plugin includes no tests of serial form that I could see. (Other Pipeline plugins do test that scenario, for Scripted syntax.) I will think about whether this could be made compatible (for anyone who has not yet upgraded).

Technically, the difficulty here is `WithScriptDescribable` does not offer a clear way to select an alternate script to use for a new API interface; and `DeclarativeAgentScript` could not use the usual idiom used in Jenkins plugins of `Util.isOverridden` to delegate to a new method variant because it is written in Java yet the `Closure` is CPS-transformed and so it would not work to either `.call()` it or in the reverse direction to create a `Closure` wrapping the new method in Java code. The same difficulty forced https://github.com/jenkinsci/docker-workflow-plugin/pull/337.

felipecrs:
@jglick thank you very much for the detailed response. I must admit I understand little of what you say, but I got the gist: it would be very hard to keep this kind of backwards compatibility, or to avoid this disruption.

PR #765, opened 2025-02-25
Converting some Groovy sources to Java

jglick:
Experimenting with whether Copilot is able to figure out how to do this efficiently. Only considering `src/main/groovy/`; `src/main/resources/**/*.groovy` are CPS-transformed.

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
@@ -268,7 +268,7 @@ public class Utils {
         if (causeClass instanceof Cause.UserIdCause && Cause.UserIdCause.class.getSimpleName().equals(cause)) {
             return detail == null || ((Cause.UserIdCause) causeClass).getUserId().equals(detail);
         } else {
-            return causeClass.getClass().getSimpleName().matches("(?i).*" + cause + ".*");
+            return causeClass.getClass().getSimpleName().matches("(?i)^" + cause + ".*");
         }
     }
 


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
jglick, line 1: This file is right at the similarity limit for Git, so any further edits would risk making the diff unreviewable!

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
```diff
@@ -22,257 +22,238 @@
...
-     * This exists for pre-1.1.2 -> later upgrades of running builds only.
-     *
-     * @param environment The environment to pull credentials from
-     * @return A non-null but possibly empty map of strings to {@link CredentialWrapper}s
-     */
-    @NonNull
-    @Restricted(NoExternalUse.class)
-    static Map<String, CredentialWrapper> getLegacyEnvCredentials(@NonNull Environment environment) {
-        Map<String, CredentialWrapper> m = [:]
-        environment.each {k, v ->
```
jglick, line 139: No longer compilable. Introduced in #147 but (AFAICT) without test coverage, and not clear it ever worked.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
```diff
@@ -283,152 +264,151 @@ static boolean isRunCausedBy(CpsScript script, String cause, String detail = nul
...
-    static boolean shouldRunBeAllowed(Cause causeClass, String cause, String detail){
-        if( causeClass instanceof Cause.UserIdCause && Cause.UserIdCause.simpleName == cause){
-            return detail == null || causeClass.userId == detail
-        }else {
-            return causeClass.class.simpleName.matches("(?i)\\.*${cause}.*")
+    public static boolean shouldRunBeAllowed(Cause causeClass, String cause, String detail) {
+        if (causeClass instanceof Cause.UserIdCause && Cause.UserIdCause.class.getSimpleName().equals(cause)) {
+            return detail == null || ((Cause.UserIdCause) causeClass).getUserId().equals(detail);
+        } else {
+            return causeClass.getClass().getSimpleName().matches("(?i).*" + cause + ".*");
```
jglick, line 271: Unclear what was meant by `\.*cause.*`. This would seem to mean `mycauselala`, `.mycause`, `..`mycausexxx`, etc.? Does not seem to make any sense. I think `.*` (any characters) was meant. Maybe `.*\.`?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
```diff
@@ -881,13 +886,15 @@ private static List<JobProperty> getPropertiesToApply(@CheckForNull List<JobProp
...
+        Set<JobProperty> toApply = new TreeSet<>(Comparator.comparing(p -> p.getDescriptor().getId()));
+        toApply.addAll(existingProperties);
         // Remove all properties defined in previous Jenkinsfile and may be no longer exist
-        toApply.removeAll {it.descriptor.id in prevDefined}
+        toApply.removeIf(p -> prevDefined.contains(p.getDescriptor().getId()));
         // Replace all properties defined in current Jenkinsfile
-        toApply.addAll(newProperties)
-        return toApply.asList()
+        if (newProperties != null) {
+            toApply.addAll(newProperties);
```
jglick, line 895: Copilot quietly inserted a null check here. The parameter is declared `@CheckForNull` so this might have been an NPE, unless it was in fact always non-null?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
```diff
@@ -22,257 +22,238 @@
...
-                if (input != null) {
-                    if (input.getAction(LabelAction.class) != null &&
-                        input.getAction(ThreadNameAction.class) != null &&
-                        (stageName == null || input.getAction(ThreadNameAction)?.threadName == stageName)) {
-                        // This is actually a parallel block
-                        return true
-                    }
+    public static Predicate<FlowNode> isParallelBranchFlowNode(final String stageName) {
+        return input -> {
+            if (input != null) {
```
jglick, line 199: (ignore WS)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
```diff
@@ -796,17 +798,18 @@ private static boolean isObjectsEqualsXStream(@NonNull Object first, @NonNull Ob
...
-        //Replace all triggers defined in current Jenkinsfile
-        toApply.addAll(newTriggers)
-        return toApply.asList()
+        // Store triggers with unique descriptor id (Class)
+        Set<Trigger> toApply = new TreeSet<>(Comparator.comparing(t -> t.getDescriptor().getId()));
+        toApply.addAll(existingTriggers);
+        // Remove all triggers defined in previous Jenkinsfile and may be deprecated
+        toApply.removeIf(t -> prevDefined.contains(t.getDescriptor().getId()));
+        // Replace all triggers defined in current Jenkinsfile
+        if (newTriggers != null) {
```
jglick, line 807: (ditto)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
jglick: …which is causing a problem since resolving SpotBugs violations puts it over the limit.

jglick: :thinking: maybe I could merge a change to trunk first which just pads out the `*.groovy` file with lots of long multiline comments that say nothing, and then leave them there, so that Git considers them unmodified.

jglick: (Mercurial IMO is better in this regard since `hg mv` explicitly records the rename, rather than leaving subsequent `diff` commands to guess at whether it might have happened.)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
```diff
@@ -926,8 +933,12 @@ private static List<JobProperty> getPropertiesToUpdate(@CheckForNull List<JobPro
     @NonNull
     private static List<JobProperty> getPropertiesToRemove(@CheckForNull List<JobProperty> currentProperties,
                                                            @NonNull List<JobProperty> existingProperties) {
-        Set<String> currentPropertiesDescriptors = currentProperties.collect{ it.descriptor.id }.toSet()
-        return existingProperties.findAll{ !(it.descriptor.id in currentPropertiesDescriptors)}
+        Set<String> currentPropertiesDescriptors = currentProperties != null ? currentProperties.stream()
```
jglick, line 936: Was able to resolve one SpotBugs violation here (`currentProperties` null check), but if I do the same in a nearby method, the diff goes over 50% modified: https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/765#discussion_r1970688710

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
```diff
@@ -902,17 +907,19 @@ private static List<JobProperty> getPropertiesToApply(@CheckForNull List<JobProp
...
+        Map<String, Integer> countMap = new HashMap<>();
+        for (JobProperty property : existingProperties) {
+            countMap.compute(property.getDescriptor().getId(), (key, count) -> count == null ? 1 : count + 1);
         }
 
-        return currentProperties.findAll{ descriptorsToExistingProperties[it.descriptor.id] == null ||
-                countMap[it.descriptor.id] > 1 ||
-                !(Objects.equals(it, descriptorsToExistingProperties[it.descriptor.id]) ||
-                        isObjectsEqualsXStream(it, descriptorsToExistingProperties[it.descriptor.id]))}
+        return currentProperties.stream()
```
jglick, line 917: (`currentProperties` null check)

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.java
daniel-beck: Called via `#isRunCausedBy` by `TriggeredByConditionalScript`. So seems to support `when { triggeredBy 'SCMTrigger' }` etc.

What the pattern generally seems to do is it allows shorthand, e.g. `triggeredBy 'SCM'` (or `'scm'` thanks to case-insensitivity). I don't see how a `#getSimpleName` would start with `.` though, so all this currently does is effectively `^` (no prefix).

The "fix" appears to change behavior (from prefix to anywhere) so should be done only carefully considered.

jglick: > all this currently does is effectively `^` (no prefix)

Agreed. Maybe better to just write it that way for now.

dwnusbaum (approved):
Seems fine. The only thing that I noticed was that `[:]` has been converted to `HashMap` everywhere rather than `LinkedHashMap`. Whether this makes a meaningful difference anywhere was not immediately obvious to me.

joseblas (approved):

jglick:
> `[:]` has been converted to `HashMap` everywhere rather than `LinkedHashMap`

Hmm, good point, probably needs to be checked.

PR #766, opened 2025-02-26
`DeclarativeAgentScriptTest.noneCompat`

jglick:
https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/762#issuecomment-2683427746


jglick:
Updated to trunk, this fails with

```
java.lang.NullPointerException: Cannot invoke method call() on null object
	at org.codehaus.groovy.runtime.NullObject.invokeMethod(NullObject.java:91)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:47)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
	at org.codehaus.groovy.runtime.callsite.NullCallSite.call(NullCallSite.java:34)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:116)
	at com.cloudbees.groovy.cps.sandbox.DefaultInvoker.methodCall(DefaultInvoker.java:20)
	at org.jenkinsci.plugins.workflow.cps.LoggingInvoker.methodCall(LoggingInvoker.java:117)
	at org.jenkinsci.plugins.pipeline.modeldefinition.ModelInterpreter.evaluateSequentialStages(ModelInterpreter.groovy:173)
```

Apparently it is running the old version of `evaluateSequentialStages` but calling the new version of `evaluateStage`. And the error does _not_ reproduce the `InvalidClassException` claimed from https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/762#issuecomment-2683241331 which might be specific to `LabelScript` even though `NoneScript` also changed superclass.

Neither `ModelStepLoader.getValue` nor `None.getScript` seem to be called during the test. I can insert a `static` initializer into `ModelInterpreter` but that seems to be about it. There is no obvious route to a fix.

jglick:
Might be easier to test this with RJR.

PR #767, opened 2025-02-26
`DeclarativeAgentScriptTest.noneCompat` (II)

jglick:
Resolves the failure mentioned in https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/766#issuecomment-2685912239. Probably does not help with the `InvalidClassException` and there are probably other errors involving `parallel` etc. which would only appear with more complex tests. And conversely breaks compatibility for builds running in 2.2234, as can be seen by merging 230ad2bd912ad9850f0e1bb94a39c1e60e080468:

```
java.lang.AssertionError: 

Expected: log containing "OK ran stage three"
     but: was 
Started
[Pipeline] Start of Pipeline
[Pipeline] stage
[Pipeline] { (one)
[Pipeline] echo
OK ran stage one
[Pipeline] }
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (two)
[Pipeline] input
continue stage two
Proceed or Abort
Resuming build at Wed Feb 26 14:41:00 EST 2025 after Jenkins restart
Ready to run at Wed Feb 26 14:41:00 EST 2025
Approved by SYSTEM
[Pipeline] }
[Pipeline] // stage
[Pipeline] End of Pipeline
Finished: SUCCESS

	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.jenkinsci.plugins.pipeline.modeldefinition.agent.impl.DeclarativeAgentScriptTest.noneCompat(DeclarativeAgentScriptTest.java:89)
```

Filing for completeness though I do not think this approach is practical.


jglick:
Another idea I had was to introduce an extension point to https://github.com/jenkinsci/workflow-support-plugin/blob/2a79d8a4be14c15e1ab7502407c42b5dcff0a14e/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverReader.java#L154 allowing plugins to inject alternate class loading locations. Then, if this plugin could detect that `build.xml` referred to a specific old version in `ModelASTPipelineDef` (note that the current test data setup via `mvn test` would _not_ work in that case without manual fixup, since it always uses `999999-SNAPSHOT`), load `LabelScript.old.groovy` and `ModelInterpreter.old.groovy` etc. with the original contents. This would be relatively nonintrusive since the compatibility code would clearly be unused for newly started builds.

PR #778, opened 2025-03-22
[JENKINS-75468] Fix Generator rendering/generation of "options" directive

rrjjvv:
* JENKINS issue(s):
    * https://issues.jenkins.io/browse/JENKINS-75468
* Description:
    * Fixes NPEs which stop snippet generation from fully working
* Documentation changes:
    * Bug fix; no new functionality
* Users/aliases to notify:
    * @basil 

This contains the same fix from https://github.com/jenkinsci/workflow-cps-plugin/pull/1003 and a second fix, both which are needed for the `overrideIndexTriggers` option to render/generate properly.

Before the fix, both a regular pipeline job and multibranch job would result in the same UI
the same stacktrace at the time it was rendered.  They also had the same lack of snippet and resulting stacktrace when actually generating the output.  This representative screenshot is for a multibranch job (which *should* be configurable):
image

With this fix, no stacktraces are generated in any scenario.  A regular pipeline job still looks the same, but generates a valid snippet:
image

A multibranch job now shows the actual config options and all three possible combinations (only one shown here) generate the correct snippet:
image

That property has been [returning `null`](https://github.com/jenkinsci/branch-api-plugin/blame/d4e0cb42cdaf341c6fa7bf778688a4968cc21dce/src/main/java/jenkins/branch/OverrideIndexTriggersJobProperty.java#L77) for many years.  This may have been broken for that long, but if I had to guess, there may have been a UI change that made it relevant.

It seems weird to me that for a multibranch project, you basically have to select it twice to actually use it: once from the menu, and then clicking the checkbox.  Or from another angle, that property isn't relevant for plain pipeline jobs yet shows up in the dropdown; it would make more sense to not be in the list at all.

In other words, maybe at one point that (seemingly redundant) level of indirection was not present and the act of selecting it from the dropdown made it "selected" (thus returning a valid object rather than `null`).

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/OptionsDirective.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/OptionsDirective.java
@@ -46,6 +46,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -58,7 +59,7 @@ public class OptionsDirective extends AbstractDirective<OptionsDirective> {
     @DataBoundConstructor
     public OptionsDirective(List<Describable> options) {
         if (options != null) {
-            this.options.addAll(options);
+            this.options.addAll(options.stream().filter(Objects::nonNull).toList());
         }
     }
 
--- a/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
+++ b/pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
@@ -43,7 +43,7 @@
                 </f:block>
                 <f:section title="${%Directives}"/>
                 <!-- Similar to f:dropdownDescriptorSelector, but adds fallback content to block, and JENKINS-25130 adds per-selection help: -->
-                <j:set var="item" value="${it.getItem(request)}"/>
+                <j:set var="item" value="${it.getItem(request2)}"/>
                 <d:taglib uri="local">
                     <d:tag name="listDirectives">
                         <j:set var="inStage" value="false"/>
--- a/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGeneratorTest.java
+++ b/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGeneratorTest.java
@@ -419,6 +419,17 @@ public class DirectiveGeneratorTest {
                 "}");
     }
 
+    @Issue("JENKINS-75468")
+    @Test
+    public void optionsHandlesNullProperties() throws Exception {
+        List<Describable> o = new ArrayList<>();
+        // This can happen in at least one known instance; the descriptor for OverrideIndexTriggersJobProperty
+        // (from branch-api) will return a null instance if the incoming form data says it hasn't been selected.
+        o.add(null);
+        OptionsDirective options = new OptionsDirective(o);
+        dg.assertGenerateDirective(options, "options {\n}");
+    }
+
     @Test
     public void tools() throws Exception {
         ToolsDirective tools = new ToolsDirective(Collections.singletonList(new ToolsDirective.SymbolAndName("maven::::apache-maven-3.0.1")));


pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
```diff
@@ -43,7 +43,7 @@
                 </f:block>
                 <f:section title="${%Directives}"/>
                 <!-- Similar to f:dropdownDescriptorSelector, but adds fallback content to block, and JENKINS-25130 adds per-selection help: -->
-                <j:set var="item" value="${it.getItem(request)}"/>
+                <j:set var="item" value="${it.getItem(request2)}"/>
```
jglick, line 46: This seems like a fix for a recent regression due to picking up a newer `workflow-cps` dep, right? So this much could be filed separately.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/OptionsDirective.java
```diff
@@ -58,7 +59,7 @@ public class OptionsDirective extends AbstractDirective<OptionsDirective> {
     @DataBoundConstructor
     public OptionsDirective(List<Describable> options) {
         if (options != null) {
-            this.options.addAll(options);
+            this.options.addAll(options.stream().filter(Objects::nonNull).toList());
```
jglick, line 62: This does not look right. A list containing nulls is not a valid argument. What is the stack trace of the code which calls this constructor with a list containing a null?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGeneratorTest.java
```diff
@@ -419,6 +419,17 @@ public void options() throws Exception {
                 "}");
     }
 
+    @Issue("JENKINS-75468")
+    @Test
+    public void optionsHandlesNullProperties() throws Exception {
+        List<Describable> o = new ArrayList<>();
+        // This can happen in at least one known instance; the descriptor for OverrideIndexTriggersJobProperty
+        // (from branch-api) will return a null instance if the incoming form data says it hasn't been selected.
```
jglick, line 427: https://github.com/jenkinsci/branch-api-plugin/blob/d4e0cb42cdaf341c6fa7bf778688a4968cc21dce/src/main/java/jenkins/branch/OverrideIndexTriggersJobProperty.java#L77 yes but that is the same as https://github.com/jenkinsci/jenkins/blob/dddb4598df25527220a1e0f46d1a0d25217dd4c8/core/src/main/java/jenkins/model/OptionalJobProperty.java#L61 (it should probably be retrofitted to extend `OptionalJobProperty` for simplicity), and https://github.com/jenkinsci/jenkins/blob/dddb4598df25527220a1e0f46d1a0d25217dd4c8/core/src/main/java/hudson/model/JobPropertyDescriptor.java#L63 advertises this as legitimate. But the calling code should simply be omitting nulls from the list it passes on.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/OptionsDirective.java
jglick: I am guessing that regular calls to `JobPropertyDescriptor` explicitly filter out nulls somewhere, but this logic is missing from `OptionsDirective.DescriptorImpl` (needs an override a `newInstance` overload).

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGeneratorTest.java
jglick: compare https://github.com/jenkinsci/jenkins/blob/dddb4598df25527220a1e0f46d1a0d25217dd4c8/core/src/main/java/hudson/util/DescribableList.java#L190-L191 called from https://github.com/jenkinsci/jenkins/blob/dddb4598df25527220a1e0f46d1a0d25217dd4c8/core/src/main/java/hudson/model/Job.java#L1406

jglick (changes requested):

pipeline-model-definition/src/main/resources/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGenerator/index.jelly
basil: I merged this separately in #777.

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGeneratorTest.java
rrjjvv: I missed that comment about a null return being legit.  In fact, I originally thought the issue was in those `newInstance` calls since the EE change resulted in making another one of those bridge/delegate methods, I saw mentions of "the request shouldn't be null, but sometimes is", and the intersection of those two.  Obviously that was a red herring.

> But the calling code should simply be omitting nulls from the list it passes on.

IIRC, the caller was jelly evaluation code; it was the one creating the list itself, adding null elements, and passing it to the `OptionsDirective` constructor.  Doing that filtering there would fix all the problems with the other plugins in one shot, but doesn't seem like it should be the one to decide that nulls are worthless...?

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/OptionsDirective.java
rrjjvv: There is no (exception) stack trace involved with calling the constructor (which occurs when the page is being rendered).  The (exception) stack trace occurs when generating the snippet using the nulls that were passed:
```
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "o" is null
at PluginClassLoader for structs//org.jenkinsci.plugins.structs.describable.DescribableModel.uninstantiate2_(DescribableModel.java:771)
at PluginClassLoader for structs//org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable.from(UninstantiatedDescribable.java:213)
at PluginClassLoader for pipeline-model-definition//org.jenkinsci.plugins.pipeline.modeldefinition.generator.OptionsDirective$DescriptorImpl.toGroovy(OptionsDirective.java:144)
at PluginClassLoader for pipeline-model-definition//org.jenkinsci.plugins.pipeline.modeldefinition.generator.OptionsDirective$DescriptorImpl.toGroovy(OptionsDirective.java:70)
at PluginClassLoader for pipeline-model-definition//org.jenkinsci.plugins.pipeline.modeldefinition.generator.DirectiveDescriptor.toIndentedGroovy(DirectiveDescriptor.java:63)
at PluginClassLoader for pipeline-model-definition//org.jenkinsci.plugins.pipeline.modeldefinition.generator.DirectiveGenerator.doGenerateDirective(DirectiveGenerator.java:94)
```
If you're literally asking about the call stack when the constructor is called with null entries in the list (despite not throwing an error), I can dig it up later, but don't have it handy.

So yeah, my guess (and it is just a guess) was that the UI was previously done in a way that resulted in other plugins' `newInstance` not returning null in the first place.  But equally plausible that something was responsible for filtering them before it got to the constructor but is no longer in play.

I chose to filter them in the constructor simply because it seemed most straightforward, safe, and easiest to test.  I couldn't identify any reason why it would be beneficial to continuing storing nulls (and just skipping over them in `toGroovy`).  Is overriding `newInstance` is a 'better' solution?

I knew that my fix was possibly addressing a symptom of some other fundamental problem.  Even if you believe the root cause/solution is elsewhere, I'd think you'd want to deal with this null either way (whether it's silently filtering it like I did, a fail-fast `Objects.requireNonNull`, or something in-between)...?

So I'm clear, are you not a fan of this fix?  Or is it fine, but needs additional fixes (e.g. whatever is sending the nulls in the first place)?  Or do you need that call stack (at point of construction) to decide what is/isn't needed?

rrjjvv: Call stack leading to the constructor:
```
	  at org.jenkinsci.plugins.pipeline.modeldefinition.generator.OptionsDirective.<init>(OptionsDirective.java:61)
	  at java.lang.invoke.DirectMethodHandle$Holder.newInvokeSpecial(DirectMethodHandle$Holder:-1)
	  at java.lang.invoke.LambdaForm$MH/0x00007a50a4001c00.invoke(LambdaForm$MH:-1)
	  at java.lang.invoke.Invokers$Holder.invokeExact_MT(Invokers$Holder:-1)
	  at jdk.internal.reflect.DirectConstructorHandleAccessor.invokeImpl(DirectConstructorHandleAccessor.java:87)
	  at jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
	  at java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
	  at java.lang.reflect.Constructor.newInstance(Constructor.java:486)
	  at org.kohsuke.stapler.RequestImpl.invokeConstructor(RequestImpl.java:686)
	  at org.kohsuke.stapler.RequestImpl.instantiate(RequestImpl.java:1003)
	  at org.kohsuke.stapler.RequestImpl$TypePair.convertJSON(RequestImpl.java:874)
	  at org.kohsuke.stapler.RequestImpl.bindJSON(RequestImpl.java:633)
	  at org.kohsuke.stapler.RequestImpl.bindJSON(RequestImpl.java:628)
	  at hudson.model.Descriptor.bindJSON(Descriptor.java:662)
	  at hudson.model.Descriptor.newInstanceImpl(Descriptor.java:620)
	  at hudson.model.Descriptor.newInstance(Descriptor.java:594)
	  at org.jenkinsci.plugins.pipeline.modeldefinition.generator.DirectiveGenerator.doGenerateDirective(DirectiveGenerator.java:89)
	  at java.lang.invoke.LambdaForm$DMH/0x00007a50a4635400.invokeVirtual(LambdaForm$DMH:-1)
	  at java.lang.invoke.LambdaForm$MH/0x00007a50a4893000.invoke(LambdaForm$MH:-1)
	  at java.lang.invoke.LambdaForm$MH/0x00007a50a4854800.invoke(LambdaForm$MH:-1)
	  at java.lang.invoke.Invokers$Holder.invokeExact_MT(Invokers$Holder:-1)
	  at java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:733)
	  at org.kohsuke.stapler.Function$MethodFunction.invoke(Function.java:484)
	  at org.kohsuke.stapler.Function$InstanceFunction.invoke(Function.java:497)
	  at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:218)
	  at org.kohsuke.stapler.Function.bindAndInvokeAndServeResponse(Function.java:140)
	  at org.kohsuke.stapler.MetaClass$12.doDispatch(MetaClass.java:686)
	  at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:61)
	  at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:800)
	  at org.kohsuke.stapler.Stapler.invoke(Stapler.java:937)
	  at org.kohsuke.stapler.MetaClass$10.dispatch(MetaClass.java:590)
	  at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:800)
	  at org.kohsuke.stapler.Stapler.invoke(Stapler.java:937)
	  at org.kohsuke.stapler.Stapler.invoke(Stapler.java:721)
	  at org.kohsuke.stapler.Stapler.service(Stapler.java:253)
	  at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:587)
<snip>
```

jglick: From a quick look, I suspect Stapler is at fault, missing a null check somewhere like https://github.com/jenkinsci/stapler/blob/e4108480341037ddb1780fddc69cd078a92cee35/core/src/main/java/org/kohsuke/stapler/RequestImpl.java#L929 though I have no idea what might have changed in the GUI recently to trigger this.

> are you not a fan of this fix?

No, it is not fixing the root cause. Fine for the constructor here to check for null entries, but it should then throw `IllegalArgumentException`.

rrjjvv: > though I have no idea what might have changed

If it helps, I can spend time later trying to find a working version to identify whether it was core/stapler or plugin/jelly that changed.  But if you have an automated way of doing that, I won't spend hours doing something that'll take you minutes.

> but it should then throw IllegalArgumentException.

Cool.  If it's okay with you, I'll keep the PR as is for the time being (so anyone needing a quick-fix has one), and change the behavior/test once actual root cause is identified/addressed.

jglick: > an automated way of doing that

Well, `git bisect`, but made more difficult by the fact that Jenkins core lacks CD versioning.

> keep the PR as is for the time being (so anyone needing a quick-fix has one)

Makes sense.

rrjjvv: > Well, `git bisect`

I was more referring to the much more difficult part: actually getting it to build and then plugin compatibility.  Out of the three tags I picked at random, I believe two required required swapping back to an old java version, all had different built instructions (for "quick build"), and two didn't build at all.  I switched tactics and starting using published docker images, but anecdotally, nothing prior to 2.444 works; on the surface, it seems that Jenkins isn't (or wasn't) capable of selecting plugins from the update center applicable to the running version (only the latest).

We maintain all the (derivative) images we've built internally, and naturally, the Dockerfiles (and more importantly, plugin list) used to build them; between the two, I'll figure it out.  But at the very least, I can say that the bug is present in 2.444 (so over a year).

rrjjvv: I snuck in a few minutes to try this, and didn't take long to bear fruit.  It's probably safe to say this has *never* worked.  According to the [changelog](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/CHANGELOG.md#version-128), the generator was introduced in 1.2.8.  The first Jenkins we deployed internally with that functionality was Jenkins 2.121 with plugin version 1.2.9.  Ignoring line numbers, the NPE and stacktrace are identical (when generating the snippet).  Whether the stack at the point of construction is identical I don't know... if you feel we need to prove that it is also identical, I can do so.

In summary, I'm 98% sure this never worked, thus not a regression, so perhaps truly a data validation issue after all.  That, or a 7+ year old bug/design flaw.  I hoped 98% was "good enough"; to get to 100% (using the previous point release of the plugin, stripping away all other irrelevant plugins, and using different core versions) would take a bit more time and effort.

Thoughts?

pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/generator/DirectiveGeneratorTest.java
jglick: When Stapler is building a `List<Describable>` from a `JSONArray` (which I think is the case here), a null element would never be a valid outcome from form databinding. It should probably be documented more clearly that descriptors can elect to return null for an inappropriate/missing input, and that this should be treated as <no value> and thus omitted from any lists, just like `DescribableList` already does. https://github.com/jenkinsci/jenkins/blob/ee42b281fa416452d5a3ade0a6f4e6f3cdb7742e/core/src/main/java/hudson/model/ReconfigurableDescribable.java#L82 already documents this while https://github.com/jenkinsci/jenkins/blob/ee42b281fa416452d5a3ade0a6f4e6f3cdb7742e/core/src/main/java/hudson/model/Descriptor.java#L585-L590 neglects to, even though https://github.com/jenkinsci/jenkins/blob/ee42b281fa416452d5a3ade0a6f4e6f3cdb7742e/core/src/main/java/hudson/model/JobPropertyDescriptor.java#L63 does in an override. https://github.com/jenkinsci/stapler/blob/c9ad5323f423a2face377eef2c39c4583b453cb4/core/src/main/java/org/kohsuke/stapler/BindInterceptor.java#L25 and https://github.com/jenkinsci/stapler/blob/c9ad5323f423a2face377eef2c39c4583b453cb4/core/src/main/java/org/kohsuke/stapler/BindInterceptor.java#L40 do discuss null. https://github.com/jenkinsci/jenkins/blob/ee42b281fa416452d5a3ade0a6f4e6f3cdb7742e/core/src/main/java/hudson/model/Descriptor.java#L1209 fails to check for null, which probably just never mattered because no impls returned null in this context.

rrjjvv: I saw this, but my bad for not saying I couldn't spent time on this at the time.  But I got a chance to revisit it now.  Based on your comment, I have a fix; the existing stapler tests work, the entire Jenkins core tests continue to pass (with the new stapler build), and of course, the original issue is resolved (using the released plugin, without these PR changes).

That project doesn't have contributor guidelines (or for that matter, any recent contributions from non-maintainers).  Do you want:
1. a standalone PR against that repo (possibly linking to the JIRA issue backing this PR)?
2. a PR in that repo alongside a dedicated GH issue?
3. to take it over from here?

I know it'll take you or a maintainer a fraction of the time it takes to walk me through this process.  I'm happy to see this through, but I'm also happy to let you handle it for the sake of efficiency (in which case I'm happy to provide the actual code and tests I currently have, which you can use/modify/ignore as you see fit).

In a nutshell, rather than modifying every place where a list/array was being built up, I replaced https://github.com/jenkinsci/stapler/blob/d77814cf7211b6abe2148dd2df2d7615bb195817/core/src/main/java/org/kohsuke/stapler/RequestImpl.java#L817 with a helper method that wraps it an a new `Lister` that only adds non-null entries.

jglick: A PR to the `stapler` repo would be great. Maintainers can take it from there in terms of making sure it gets tested and integrated properly. Thank you for your patience!

rrjjvv: I took a stab at this in https://github.com/jenkinsci/stapler/pull/660.  Regarding this PR, you had mentioned earlier that it would be better to fail-fast on `null` elements rather than filtering them.  Are you in favor of actually doing that assertion, or simply cancelling/denying this PR?

rrjjvv: Probably better to just cancel it then.  It's of dubious value on its own, and all the other areas of improvement mentioned in these threads are in other plugins.

PR #1495, opened 2026-01-23
feat: Allow list variables and GString expressions

rahulsom:
This supports `values` in a decalarative pipeline's matrix axis to be a list variable.
Also, it allows constructing values out of GStrings.

Refs: [JENKINS-62127](https://issues.jenkins.io/browse/JENKINS-62127)
Refs: https://github.com/jenkinsci/matrix-project-plugin/issues/719

### DIFF
--- a/pipeline-model-definition/src/test/resources/matrix/matrixWithVariable.groovy
+++ b/pipeline-model-definition/src/test/resources/matrix/matrixWithVariable.groovy
@@ -1,0 +1,54 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+def partialName = "win"
+def osList = ["linux", "${partialName}dows", "mac"]
+pipeline {
+  agent none
+  stages {
+    stage("foo") {
+      matrix {
+        axes {
+          axis {
+            name 'OS_VALUE'
+            values osList
+          }
+        }
+        stages {
+          stage("first") {
+            steps {
+              echo "First branch"
+              echo "OS=${OS_VALUE}"
+            }
+          }
+          stage("second") {
+            steps {
+              echo "Second branch"
+            }
+          }
+        }
+      }
+    }
+  }
+}
--- a/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
+++ b/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/parser/ModelParser.groovy
@@ -478,7 +478,11 @@ class ModelParser implements Parser {
                             }
                             break
                         case 'values':
-                            a.values.addAll(method.args);
+                            List<Expression> valArgs = ((TupleExpression) mc.arguments).expressions
+                            valArgs.each { argExpr ->
+                                def resolvedValues = resolveListVariableOrParse(argExpr)
+                                a.values.addAll(resolvedValues)
+                            }
                             break
                         default:
                             // We need to check for unknowns here.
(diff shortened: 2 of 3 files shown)


rahulsom:
@jenkinsci/pipeline-model-definition-plugin-developers

AH-Hammi:
I would really like this feature, is there a reason this is not yet merged, is there some problem?

ssiano:
I would also really like this feature and have had to work around it not being available.

pipeline-model-definition/src/test/resources/matrix/matrixWithVariable.groovy
```diff
@@ -0,0 +1,54 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+def partialName = "win"
+def osList = ["linux", "${partialName}dows", "mac"]
+pipeline {
```
jglick, line 27: If you find yourself doing this, just stop trying to use Declarative Pipeline now and switch to Scripted.

jglick (commented):
Not desirable IMO.

rahulsom:
Hi @jglick! Thanks for taking a look. Could you share some insight into why scripted is better suited to this?

In talking to people maintaining shared libraries, I hear they find the declarative syntax easier to reason about. This could allow them to parameterize things like an OS list or a JDK list without having to repeat stages and write complex conditional logic.

ssiano:
My use case is a Multibranch Pipeline with a large number of active branches.  Older branches support a different set of build platforms than newer branches, and the set of supported platforms can change over time for all of the branches.  For example, we might drop a platform for five older branches.

I have in a library a declarative pipeline that uses `matrix` so that I don't need to update the pipeline in a large number of branches whenever the platforms change.  It would be *very powerful* if I could set the branch-specific axis values (platforms) in a single location and pass that list to the axis' `values` parameter.

And from the [Pipeline Syntax reference section on Matrix](https://www.jenkins.io/doc/book/pipeline/syntax/#declarative-matrix), there is nothing that indicates the values must be hard-coded.  It shows hard-coded values in the examples, but one could easily look at the doc and assume they will be able to set the list dynamically.

So I think this feature both desirable and natural, i.e., it does not violate Jenkins' conventions.

ssiano:
@jglick, I would also like to hear why you suggest that users convert their declarative pipelines to scripted pipelines.

I have been using Jenkins for less than a year, and from the "Jenkins 2: Up & Running" book and the online Jenkins docs, as well as other online help, I understood that declarative is the modern and recommended approach.

Declarative pipelines have worked out very well for me, and I would like to continue to use them.  And this new feature from @rahulsom makes them even more powerful.

pipeline-model-definition/src/test/resources/matrix/matrixWithVariable.groovy
```diff
@@ -0,0 +1,54 @@
...
+  agent none
+  stages {
+    stage("foo") {
+      matrix {
+        axes {
+          axis {
+            name 'OS_VALUE'
+            values osList
```
rsandell, line 35: I would expect this to, in the normal case, be something like a function/step call to a shared library that provides a list of "OS flavours that must be tested according to higher powers"?

jglick:
> declarative is the modern and recommended approach

It is fine if you have a pipeline with trivial structure that requires no dynamic configuration. The moment you do, switch to Scripted.

AH-Hammi:
A possible solution to this problem could be the following section in the Jenkins documentation.
https://www.jenkins.io/doc/book/pipeline/syntax/#matrix-cell-directives
you could use a when clause to check if the current cell should be executed. 
This still does not provide the ability to give arbitrary, dynamically determined values to the cells, but some use cases are possible.

This should not mean, that the pull request can be closed, but this should show that similar behavior is possible but harder to implement. Therefore, I would still very much like to see this request merged.

@jglick, what is the reason against merging?

My use case would be that I read out a pyproject toml and search for all valid python versions to test, against a list of available python versions. The list of available versions changes over time, and therefore I would like to automatically update this over time, without updating the CI Pipeline.

Sevneoy:
您好！
      您发的邮件我已收到，我会尽快回复！

solomax:
@jglick while it is possible to create script and do any amount of nested loops there will be no "bricks" in the UI in this case

I would VOTE for adding this feature to the plugin :)

ssiano:
@rsandell could you please review and approve this PR?

In general the comments are positive and it is viewed as a nice feature.

ssiano:
@jglick why are you advocating that we switch our pipelines to scripted just because there is some dynamic configuration?  At my company all of the pipelines are declarative.  Can you imagine me explaining to my employer that we need to switch our pipelines to scripted even though they are working just fine?

rsandell:
Well, I am not totally against it in principle to have more dynamic axis, mostly because the matrix project type allows it but it allows it in a way that doesn't translate well to this. So Andrew had a point when he intentionally made this not possible.
What makes me hesitate with this implementation, at least the way it is demonstrated to work in the unit test, is that it "forces" scripted content to be intermixed with declarative syntax. And we can't have a solution that relies soley on that.

What I am torn about it if there is a sollution to that problem. Would a shared library function or step work and be enough, or still "too scripted"? Or perhaps a white space separated string like the "vanilla" matrix job configuration? How then to specify to the axis that's the way to interpret the value? Any other ideas that I can't think of now?

ssiano:
@rsandell I am not understanding the concerns regarding scripted vs declarative since declarative pipelines have a [script step](https://www.jenkins.io/doc/book/pipeline/syntax/#script).  Could you please elaborate?

Also I don't understand the concern with having some "dynamic" code above the pipeline.  Many Jenkinsfiles start with dynamic code that is followed by the declarative pipeline.

ssiano:
@jglick and @rsandell, I am starting to understand your concern.

At my company, almost all our pipelines have dynamic code above the pipeline block, and I assumed that was standard practice:
- This [CloudBees video](https://www.youtube.com/watch?v=41uUsWQjKRw) makes it appear normal
- On the other hand, the [Jenkins Pipeline Syntax doc](https://www.jenkins.io/doc/book/pipeline/syntax/) does not show it in the examples

As far as other ideas, it would be nice if a script step could be included within an axis directive and the name and values could be set dynamically.  But I have nowhere near enough Jenkins understanding to know if that is sensible.

And since I have no problem with dynamic code above the pipeline block, I would still love to see this feature added to Jenkins.

ssiano:
For now I can do a workaround by using the `when` directive as described in [Welcome to the Matrix](https://www.jenkins.io/blog/2019/11/22/welcome-to-the-matrix/#controlling-cell-behavior-at-runtime) by @lnewman.

It is not pretty, but here is an attempt:
```groovy
#!/usr/bin/env groovy
// groovylint-disable DuplicateNumberLiteral, DuplicateStringLiteral

// Start release for each platform
Map<String, String> platformStartRelease = [
  'RHEL_1': '0.0.0',
  'RHEL_2': '8.11.0',
  'RHEL_3': '8.14.0',
  'SUSE_1': '0.0.0',
  'SUSE_2': '8.11.0',
  'SUSE_3': '8.14.0',
  'UBUNTU_1': '0.0.0',
  'UBUNTU_2': '8.11.0',
  'UBUNTU_3': '8.14.0'
]

// Final release for each platform
Map<String, String> platformFinalRelease = [
  'RHEL_1': '8.10.4',
  'RHEL_2': '8.13.3',
  'RHEL_3': '100.0.0',
  'SUSE_1': '8.10.4',
  'SUSE_2': '8.13.3',
  'SUSE_3': '100.0.0',
  'UBUNTU_1': '8.10.4',
  'UBUNTU_2': '8.13.3',
  'UBUNTU_3': '100.0.0'
]

// True if the release is at or above (>=) the limit
@NonCPS
Boolean atOrAbove(String release, String limit) {
  List<Integer> rel = release.split(/\./)*.toInteger()
  List<Integer> lim = limit.split(/\./)*.toInteger()
  if (rel[0] != lim[0]) { return rel[0] > lim[0] }
  if (rel[1] != lim[1]) { return rel[1] > lim[1] }
  return (rel[2] >= lim[2])
}

// True if the release is at or below (<=) the limit
@NonCPS
Boolean atOrBelow(String release, String limit) {
  List<Integer> rel = release.split(/\./)*.toInteger()
  List<Integer> lim = limit.split(/\./)*.toInteger()
  if (rel[0] != lim[0]) { return rel[0] < lim[0] }
  if (rel[1] != lim[1]) { return rel[1] < lim[1] }
  return (rel[2] <= lim[2])
}

pipeline {
  agent none
  parameters {
    string(
      name: 'RELEASE',
      description: 'Release'
    )
  }
  stages {
    stage('Platform Builds') {
      matrix {
        when {
          allOf {
            expression { return atOrAbove(params.RELEASE, platformStartRelease[PLATFORM]) }
            expression { return atOrBelow(params.RELEASE, platformFinalRelease[PLATFORM]) }
          }
        }
        axes {
          axis {
            name 'PLATFORM'
            values(
              'RHEL_1',
              'RHEL_2',
              'RHEL_3',
              'SUSE_1',
              'SUSE_2',
              'SUSE_3',
              'UBUNTU_1',
              'UBUNTU_2',
              'UBUNTU_3'
            )
          }
        }
        stages {
          stage('Platform Build') {
            agent none
            steps {
              echo "$PLATFORM Build"
            }
          }
        }
      }
    }
  }
}
```

For Release 8.13.2 this would build the `RHEL_2`, `SUSE_2`, and `UBUNTU_2` platforms and skip the other six.

And of course the code above the pipeline would be moved to a library.

One problem with this workaround is that the UI will show nine cells with six of them empty.  If I could dynamically specify just the three platforms that I want, then the UI would show three cells with none of them empty.

jglick:
> there will be no "bricks" in the UI in this case

Not sure what this means.

> we need to switch our pipelines to scripted even though they are working just fine

I did not say that.

> Many Jenkinsfiles start with dynamic code that is followed by the declarative pipeline.

I am aware of that. Do not do this. It was a terrible mistake that this was ever permitted, but not one we can now compatibly rectify.

PR #1497, opened 2026-03-16
Fix race condition in parallel stage post-condition evaluation

wuillaum:
Stage-level post blocks (success, failure, etc.) were evaluating based on the global build/execution result, which during parallel execution could be tainted by a sibling stage's failure. This caused a successful parallel stage to incorrectly run its 'failure' post block instead of 'success'.

The fix isolates stage-level post-condition evaluation: when a stage has explicit stage-scoped result information (from catchError/warnError via WarningAction, or from a stage error), getCombinedResult() now returns only that stage-scoped result without combining it with the global result.

Additionally, runPostConditions() no longer temporarily mutates the global WorkflowRun.result for stage-level post blocks, eliminating a second race vector where sibling stages could read a dirty global result during concurrent post-block execution.

Includes a new test (parallelCatchErrorSiblingPostCondition) that verifies a successful parallel sibling correctly runs its 'success' post block when another sibling fails with catchError.

* JENKINS issue(s):
    * https://github.com/jenkinsci/workflow-cps-plugin/issues/1750


Sevneoy:
您好！
      您发的邮件我已收到，我会尽快回复！

PR #1500, opened 2026-05-25
`Restart from stage` - should appear after Jenkins restart

gbhat618:
`Restart from stage` button is missing after restart.

Originally noticed in CloudBees High availability environment - where 1 replica executed the build and notified another replica to load it from the disk.

Same problem is able to reproduce in a standalone open source Jenkins instance by a simple restart.
`getOrNull` has a check on a promise which is only present in the JVM where the build ran - https://github.com/jenkinsci/workflow-job-plugin/blob/e4bd0dbfcbbb7f21d942fe84c817811c51100758/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L1025-L1026

|Description|Before restart|After restart|
|---|---|---|
|Before fix|image|image|
|After fix|image|image|

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
@@ -93,8 +93,15 @@ public class RestartDeclarativePipelineAction implements Action {
         if (owner == null) {
             return null;
         }
-        FlowExecution exec = owner.getOrNull();
-        return exec instanceof CpsFlowExecution ? (CpsFlowExecution) exec : null;
+        // Use get() rather than getOrNull(): for a completed build loaded from disk (e.g. after restart),
+        // the FlowExecution is not in memory, but this action is rendered on the build's own page where
+        // loading the execution is appropriate.
+        try {
+            FlowExecution exec = owner.get();
+            return exec instanceof CpsFlowExecution ? (CpsFlowExecution) exec : null;
+        } catch (IOException e) {
+            return null;
+        }
     }
 
     @Exported
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
@@ -104,7 +104,7 @@ public class RestartDeclarativePipelineAction implements Action {
             FlowExecution exec = owner.get();
             return exec instanceof CpsFlowExecution cfe ? cfe : null;
         } catch (IOException e) {
-            LOGGER.log(Level.FINE, "Exception while fetching FlowExecution for " + run + ", returning null", e);
+            LOGGER.log(Level.WARNING, e, () -> "Failed to load metadata of " + run);
             return null;
         }
     }
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
@@ -98,7 +98,7 @@ public class RestartDeclarativePipelineAction implements Action {
         // loading the execution is appropriate.
         try {
             FlowExecution exec = owner.get();
-            return exec instanceof CpsFlowExecution ? (CpsFlowExecution) exec : null;
+            return exec instanceof CpsFlowExecution cfe ? cfe : null;
         } catch (IOException e) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


gbhat618:
manual testing pending

gbhat618:
> manual testing pending

done, works.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -93,8 +93,15 @@ private CpsFlowExecution getExecution() {
...
             return null;
         }
-        FlowExecution exec = owner.getOrNull();
-        return exec instanceof CpsFlowExecution ? (CpsFlowExecution) exec : null;
+        // Use get() rather than getOrNull(): for a completed build loaded from disk (e.g. after restart),
+        // the FlowExecution is not in memory, but this action is rendered on the build's own page where
+        // loading the execution is appropriate.
+        try {
+            FlowExecution exec = owner.get();
+            return exec instanceof CpsFlowExecution ? (CpsFlowExecution) exec : null;
```
jglick:
```suggestion
            return exec instanceof CpsFlowExecution cfe ? cfe : null;
```

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -93,8 +93,15 @@ private CpsFlowExecution getExecution() {
...
+        } catch (IOException e) {
+            return null;
```
jglick, line 108: Should log the stack trace in case it is interesting.

jglick (approved):

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
gbhat618: ty, now added a log.

gbhat618:
Hi @jtnord or @olamy can you help with this PR if all good can help merge this, ty.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java
```diff
@@ -93,8 +97,16 @@ private CpsFlowExecution getExecution() {
...
-        FlowExecution exec = owner.getOrNull();
-        return exec instanceof CpsFlowExecution ? (CpsFlowExecution) exec : null;
+        // Use get() rather than getOrNull(): for a completed build loaded from disk (e.g. after restart),
+        // the FlowExecution is not in memory, but this action is rendered on the build's own page where
+        // loading the execution is appropriate.
+        try {
+            FlowExecution exec = owner.get();
+            return exec instanceof CpsFlowExecution cfe ? cfe : null;
+        } catch (IOException e) {
+            LOGGER.log(Level.FINE, "Exception while fetching FlowExecution for " + run + ", returning null", e);
```
jtnord: do not do unconditional string concatination in a log statement - especially one which will be thrown away in almost every single environment,

either use paramaterized logging, or use a Supplier.

e.g. untested: 
```suggestion
            LOGGER.log(Level.FINE, e, () -> "Exception while fetching FlowExecution for " + run + ", returning null");
```

jglick: or
```suggestion
            LOGGER.log(Level.WARNING, e, () -> "Failed to load metadata of " + run);
```
(The exception indicates something about the build is broken, so it is potentially worth showing by default in the system log.)

gbhat618: ty. verified the log works,
```
         try {
             FlowExecution exec = owner.get();
+            if (1 == 1) {
+                throw new IOException("Simulated IOE");
+            }
             return exec instanceof CpsFlowExecution cfe ? cfe : null;
```

log
```
   0.483 [id=119]       WARNING o.j.p.p.m.a.RestartDeclarativePipelineAction#getExecution: Failed to load metadata of p #1
java.io.IOException: Simulated IOE
        at org.jenkinsci.plugins.pipeline.modeldefinition.actions.RestartDeclarativePipelineAction.getExecution(RestartDeclarativePipelineAction.java:106)
        at org.jenkinsci.plugins.pipeline.modeldefinition.actions.RestartDeclarativePipelineAction.isRestartEnabled(RestartDeclarativePipelineAction.java:123)
        at org.jenkinsci.plugins.pipeline.modeldefinition.actions.RestartDeclarativePipelineActionRestartTest.lambda$restartEnabledAfterJenkinsRestart$1(RestartDeclarativePipelineActionRestartTest.java:30)
        at org.jvnet.hudson.test.fixtures.JenkinsSessionFixture$1.evaluate(JenkinsSessionFixture.java:124)
        at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:659)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:299)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:293)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

jglick (approved):

jglick:
[CloudBees-internal link](https://cloudbees.atlassian.net/browse/BEE-72077)

PR #1502, opened 2026-06-18
Migrate EnvVarsForToolStep off deprecated classes using Guice

HannesWell:
Migrate `EnvVarsForToolStep` off deprecated classes which are using Guice.

When doing this I mainly copied the pattern from the `ToolStep` implementation:
https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/fc7d2b3ca324e5ec44d758a55e0d87e361cae330/src/main/java/org/jenkinsci/plugins/workflow/steps/ToolStep.java#L52

### DIFF
--- a/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/EnvVarsForToolStep.java
+++ b/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/EnvVarsForToolStep.java
@@ -23,7 +23,6 @@
  */
 package org.jenkinsci.plugins.pipeline.modeldefinition.steps;
 
-import com.google.inject.Inject;
 import hudson.AbortException;
 import hudson.EnvVars;
 import hudson.Extension;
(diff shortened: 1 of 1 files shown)


pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/EnvVarsForToolStep.java
```diff
@@ -94,15 +98,20 @@ public boolean isAdvanced() {
...
+        @Override
+        public Set<? extends Class<?>> getRequiredContext() {
+            return Set.of(Node.class, TaskListener.class, EnvVars.class);
+        }
     }
 
-    public static final class EnvVarsForToolStepExecution extends AbstractSynchronousNonBlockingStepExecution<List<String>> {
-        @Inject
-        private transient EnvVarsForToolStep step;
+    public static final class EnvVarsForToolStepExecution extends SynchronousNonBlockingStepExecution<List<String>> {
```
jglick, line 108: Note that https://github.com/jenkinsci/workflow-step-api-plugin/pull/17 did not cover this superclass, because https://github.com/jenkinsci/workflow-step-api-plugin/blob/538c2362bdfb63d61b2838b3de78427557db7172/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java#L71 anyway.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/EnvVarsForToolStep.java
```diff
@@ -94,15 +98,20 @@ public boolean isAdvanced() {
...
+        private final transient EnvVarsForToolStep step;
```
jglick, line 109: Needs `@SuppressFBWarnings`.

Alternately, just copy the `toolId` + `toolVersion` into the execution in its constructor.

pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/EnvVarsForToolStep.java
```diff
@@ -50,7 +53,7 @@
  *
  * @author Andrew Bayer
  */
-public final class EnvVarsForToolStep extends AbstractStepImpl implements Serializable {
+public final class EnvVarsForToolStep extends Step implements Serializable {
 
     private static final long serialVersionUID = 1L;
```
jglick, line 58: No actual reason for this ever to have been `Serializable`—it is never serialized.

